PR #3, opened 2018-01-02
Fix build with mips-mti-elf-gcc 4.9.2

vincent-d:
While integrating littlefs as a package for RIOT (https://github.com/RIOT-OS/RIOT/pull/8239), I had build errors with gcc 4.9.2 on mips (missing-field-initializers).

This PR fix them.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -1992,14 +1992,16 @@ int lfs_format(lfs_t *lfs, const struct lfs_config *cfg) {
     // write superblocks
     lfs_superblock_t superblock = {
         .off = sizeof(superdir.d),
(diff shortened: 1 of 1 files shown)


geky:
Hi @vincent-d, thanks for the patch.

Interesting though, this looks like a compiler bug that's been resolved in gcc version 5:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=64480

If you aren't able to update gcc, is it possible to add the `-Wno-missing-field-initializers` flag in the meantime?

vincent-d:
Hopefully we'll be able to update the toolchain. I'll disable the waning in the meantime.

PR #18, opened 2018-01-29
documentation touch up

aldot:
Some spelling fixes and on top of that some editorial tweaks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ test: test_format test_dirs test_files test_seek test_truncate test_parallel \
 	test_alloc test_paths test_orphan test_move test_corrupt
 test_%: tests/test_%.sh
 ifdef QUIET
-	./$< | sed -n '/^[-=]/p'
+	@./$< | sed -n '/^[-=]/p'
 else
 	./$<
 endif
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -72,12 +72,12 @@ to three strong requirements:
 
 ## Existing designs?
 
-There are of course, many different existing filesystem. Heres a very rough
+There are of course, many different existing filesystem. Here is a very rough
 summary of the general ideas behind some of them.
 
 Most of the existing filesystems fall into the one big category of filesystem
-designed in the early days of spinny magnet disks. While there is a vast amount
-of interesting technology and ideas in this area, the nature of spinny magnet
+designed in the early days of spinning magnetic disks. While there is a vast amount
+of interesting technology and ideas in this area, the nature of spinning magnetic
 disks encourage properties, such as grouping writes near each other, that don't
 make as much sense on recent storage types. For instance, on flash, write
 locality is not important and can actually increase wear destructively.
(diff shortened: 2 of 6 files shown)


aldot:
phrases i either do not understand or that i'm not happy with
DESIGN.md:
(1)
---8<---
mindblown

where:  
ctz(i) = the number of trailing bits that are 0 in i  
popcount(i) = the number of bits that are 1 in i  
---8<---

The formula has popcount(n) but the legend has popcount(i)

(2)
---8<---
We can now solve for the block index + offset while only needed to store the file size in O(1).
---8<---

There is something missing in this sentense. And s/needed/needing/.
We can now solve the equation for the block index + offset in just O(1).

?

(3) Consider how we might add a new directory.
In "directories", what happens in case of power-loss between
- create the new directory block
- power-loss! the next step is not performed:
- update root to point to directory B

How is the pointer reclaimed, resp determined unused?
Ah, deorphan on first block allocation a little bit down in that
paragraph.

thanks for clarification!

geky:
Hi @aldot, thanks for the pr. Are you able to move the test changes out to a separate pr?

aldot:
sure, sorry for that. New pull-request opened.
The questions in https://github.com/geky/littlefs/pull/18#issuecomment-361301374 are still open for your consideration, fwiw.

geky:
Sorry I haven't gotten to it yet. Let's see if I can help

1. > The formula has popcount(n) but the legend has popcount(i)

   Ah, I meant for the _i_ s to be independent for the definitions. We could use a different variable (_x_?) or use `popcount(n)`. I don't have a strong preference.

The solution involves quite a bit of math, but computers are very good at math. We can now solve for the block index + offset while only needed to store the file size in O(1).

2. > We can now solve for the block index + offset while only needed to store the file size in O(1).

   Hmm, how about "Now we can solve for both the block index and offset from the file size in O(1)". Does that make more sense? I'm open to suggestions.

Were you able to figure out 3?

geky:
Whoops, wrong button.

Feel free to open a pr for the test changes as well. I just wanted the pull requests to be organized by the content they are changing.

Also, you can force push to this branch if you want to keep the conversation history. Force pushing on pull requests is pretty innocuous and useful.

geky:
I left a review over in #21, let me know what you think

Makefile
```diff
@@ -24,28 +21,33 @@ endif
...
 .SUFFIXES:
-test: test_format test_dirs test_files test_seek test_truncate test_parallel \
-	test_alloc test_paths test_orphan test_move test_corrupt
-test_%: tests/test_%.sh
+test: $(TESTS)
+$(TESTS): $(OBJ)
+test_%: tests/test_%.sh FORCE
 ifdef QUIET
-	./$< | sed -n '/^[-=]/p'
+	@./$< | sed -n '/^[-=]/p'
```
geky: I'm going to have to at least steal some of these if that's ok with you!

aldot: I have a separate pull-request pending after the #21 which includes this.

aldot:
Re: https://github.com/geky/littlefs/pull/18#issuecomment-361623100
ad (1) i don't have a preference either but the current text is not clear IMHO. I'll leave this to you to fix.
ad (2) That makes sense, yes. Again, i'll defer this to you.
ad (3) yes, i could figure it out.

thanks,

Makefile
geky: How comfortable are you with git branches and rebasing? You could move test changes into a separate branch from doc-tweaks so it could be reveiwed and merged in parallel.

``` bash
# on master
$ git checkout -b parallel-tests
$ git rebase --onto f694b14 25fc42f parallel-tests
# now on branch parallel-tests separate from docs
$ git push origin parallel-tests
```

geky:
I'm going to close this for now since it's tracking your master. But feel free to open more prs for the other features as they're ready.

PR #24, opened 2018-02-02
Silence shadow warnings

aldot:
per $subject.

### DIFF
--- a/emubd/lfs_emubd.c
+++ b/emubd/lfs_emubd.c
@@ -190,7 +190,7 @@ int lfs_emubd_erase(const struct lfs_config *cfg, lfs_block_t block) {
     }
 
     if (!err && S_ISREG(st.st_mode) && (S_IWUSR & st.st_mode)) {
-        int err = unlink(emu->path);
+        err = unlink(emu->path);
         if (err) {
             return -errno;
         }
--- a/tests/template.fmt
+++ b/tests/template.fmt
@@ -37,8 +37,8 @@ static void test_assert(const char *file, unsigned line,
 
 
 // utility functions for traversals
-static int __attribute__((used)) test_count(void *p,
-	lfs_block_t b __attribute__((unused))) {{
+static int __attribute__((used)) test_count(void *p, lfs_block_t b) {{
+    (void)b;
     unsigned *u = (unsigned*)p;
     *u += 1;
     return 0;
--- a/tests/template.fmt
+++ b/tests/template.fmt
@@ -7,11 +7,11 @@
 
 
 // test stuff
-void test_log(const char *s, uintmax_t v) {{
+static void test_log(const char *s, uintmax_t v) {{
     printf("%s: %jd\n", s, v);
 }}
 
-void test_assert(const char *file, unsigned line,
+static void test_assert(const char *file, unsigned line,
         const char *s, uintmax_t v, uintmax_t e) {{
     static const char *last[6] = {{0, 0}};
     if (v != e || !(last[0] == s || last[1] == s ||
@@ -37,7 +37,8 @@ void test_assert(const char *file, unsigned line,
 
 
 // utility functions for traversals
-int test_count(void *p, lfs_block_t b) {{
+static int __attribute__((used)) test_count(void *p,
+	lfs_block_t b __attribute__((unused))) {{
     unsigned *u = (unsigned*)p;
     *u += 1;
     return 0;
@@ -96,7 +97,7 @@ const struct lfs_config cfg = {{
 
 
 // Entry point
-int main() {{
+int main(void) {{
     lfs_emubd_create(&cfg, "blocks");
 
 {tests}
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ test: test_format test_dirs test_files test_seek test_truncate test_parallel \
 	test_alloc test_paths test_orphan test_move test_corrupt
 test_%: tests/test_%.sh
 ifdef QUIET
-	./$< | sed -n '/^[-=]/p'
+	@./$< | sed -n '/^[-=]/p'
 else
 	./$<
 endif
--- a/lfs.c
+++ b/lfs.c
@@ -1016,6 +1016,7 @@ int lfs_dir_seek(lfs_t *lfs, lfs_dir_t *dir, lfs_off_t off) {
 }
 
 lfs_soff_t lfs_dir_tell(lfs_t *lfs, lfs_dir_t *dir) {
+    (void)lfs;
     return dir->pos;
 }
 
@@ -1669,7 +1670,8 @@ int lfs_file_truncate(lfs_t *lfs, lfs_file_t *file, lfs_off_t size) {
         return LFS_ERR_INVAL;
     }
 
-    if (size < lfs_file_size(lfs, file)) {
+    lfs_off_t oldsize = lfs_file_size(lfs, file);
+    if (size < oldsize) {
         // need to flush since directly changing metadata
         int err = lfs_file_flush(lfs, file);
         if (err) {
@@ -1686,11 +1688,11 @@ int lfs_file_truncate(lfs_t *lfs, lfs_file_t *file, lfs_off_t size) {
 
         file->size = size;
         file->flags |= LFS_F_DIRTY;
-    } else if (size > lfs_file_size(lfs, file)) {
+    } else if (size > oldsize) {
         lfs_off_t pos = file->pos;
 
         // flush+seek if not already at end
-        if (file->pos != lfs_file_size(lfs, file)) {
+        if (file->pos != oldsize) {
             int err = lfs_file_seek(lfs, file, 0, SEEK_END);
             if (err) {
                 return err;
@@ -1716,6 +1718,7 @@ int lfs_file_truncate(lfs_t *lfs, lfs_file_t *file, lfs_off_t size) {
 }
 
 lfs_soff_t lfs_file_tell(lfs_t *lfs, lfs_file_t *file) {
+    (void)lfs;
     return file->pos;
 }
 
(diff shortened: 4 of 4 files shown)


geky:
This looks ok to me. Out of curiosity what compiler and flags are you using?

aldot:
-Wshadow

resp. -Wall -Wextra  -Wshadow -Wwrite-strings -Wundef -Wstrict-prototypes -Wunused -Wunused-parameter -Wunused-function -Wunused-value -Wmissing-prototypes -Wmissing-declarations -Wold-style-definition

aldot:
there are some sign-comparison warnings left, among a few others after this, both with gcc as well as with clang. But one step at a time..

geky:
Hmmm, what's the value of these warnings?

Particularly -Wshadow and -Wunused-parameter seem counterproductive as -Wunused-parameter discourages naming parameters and shadowing variables is generally useful for keeping the scope of variables tight.

emubd/lfs_emubd.c
```diff
@@ -190,7 +190,7 @@ int lfs_emubd_erase(const struct lfs_config *cfg, lfs_block_t block) {
     }
 
     if (!err && S_ISREG(st.st_mode) && (S_IWUSR & st.st_mode)) {
-        int err = unlink(emu->path);
+        err = unlink(emu->path);
```
geky: This is a bug as it overrides the value of `err` used on line 199 (6 lines below this). It's passing because of undefined behaviour below when we use the `st` variable after `stat` failed.

aldot:
well unused function parameters should usually be removed since they are bloat, except when they are needed only in certain circumstances like in the test-template where you'd ACK that by flagging it __attribute__((used)).

geky:
What about cases such as lfs_file_tell and lfs_file_size, where the parameter may be used in the future?
https://github.com/geky/littlefs/blob/fd04ed4f254d9e21114d1068b39894248fccd68a/lfs.c#L1718-L1720

emubd/lfs_emubd.c
aldot: I don't follow.

    int err = stat(emu->path, &st);
    if (err && errno != ENOENT) {
        return -errno;
    }

    if (!err && S_ISREG(st.st_mode) && (S_IWUSR & st.st_mode)) {
/* err == 0 here, no harm to set it */
        err = unlink(emu->path);
        if (err) {
            return -errno;
        }
/* ditto */
    }
/* either stat() gave 0 (and left errno at it's value before stat()) or !0 and errno == ENOENT */
/* I.e. the first condition below is superfluous but we still should check if the open for write would succeed, i suppose */
    if (err ||  (S_ISREG(st.st_mode) && (S_IWUSR & st.st_mode))) {

Now this whole thing is heavier than needed (either blindly unlink() or cowardly access(path, W_OK) before unlinking but then again, it's just the emu thing. What am i missing? :)

aldot:
__attribute__((unused)) on a parameter just silences the warning.
The same thing holds true for __attribute__((used)) on e.g. a static function, it only suppresses the warning. That was my reasoning for flagging those test-template static functions accordingly, to make the intention clear.
Especially on public functions no compiler can optimize away unused arguments AFAIK.

emubd/lfs_emubd.c
aldot:
```c
int lfs_emubd_erase(const struct lfs_config *cfg, lfs_block_t block) {
    lfs_emubd_t *emu = cfg->context;

    // Check if erase is valid
    assert(block < cfg->block_count);

    // Erase the block
    int err = snprintf(emu->child, LFS_NAME_MAX, "%x", block);
    assert (err > 0 && err < LFS_NAME_MAX); /* truncated or other error */

    if (access(emu->path, F_OK) == 0) {
        FILE *f = fopen(emu->path, "w");
	if (f && fclose(f) == 0) {
	    emu->stats.erase_count += 1;
	    return 0;
	}
    }
    return -errno;
}
```
```shell
$ ../busybox/scripts/bloat-o-meter gcc7.a gcc7-silence.a 
function                                             old     new   delta
lfs_emubd_erase                                      203     128     -75
------------------------------------------------------------------------------
(add/remove: 0/0 grow/shrink: 0/1 up/down: 0/-75)             Total: -75 bytes
$
```
Just a thought..

geky: Oh, you're right, my bad. Though these are the sort of cases -Wshadow makes me worried about.

Either way I think your restructure of lfs_emubd_erase is a good idea. Although counterintuitively, lfs_emubd_erase should return 0 even if access fails. The tests remove write permission to simulate a block that has frozen data.

Also out of curiousity, is `fopen(path, "w"), fclose(f)` prefered over `truncate(path)` in this situation?

aldot: I've saved that away as https://github.com/aldot/littlefs/commit/1012068bcf13b0de6462f80ca081a8a20388229d now, fwiw.
Note that i didn't mean to spend any time whatsoever on emubd :P

aldot: well open(path,O_CREAT|O_TRUNC|O_NOFOLLOW|O_WRONLY,0644)/close might be an even better choice, sure.

geky:
My concern is more that function parameters are not bloat just because they are unused, since the parameters define the API and the API itself has value. A parameter can bloat an API independently of if it's used or not.

emubd/lfs_emubd.c
geky: Haha that's fair. Welp, it's not an issue with this pr like I thought it was.

geky:
Well, even if we disagree on the value of these warnings, I'm not opposed to merging in the work you've done. Though you may want to add the warnings to the CI or I may forget about them in future modifications. :)

https://github.com/geky/littlefs/blob/master/.travis.yml#L2

emubd/lfs_emubd.c
aldot: but Re: "Although counterintuitively, lfs_emubd_erase should return 0 even if access fails. The tests remove write permission to simulate a block that has frozen data."

but that isn't the case currently, is it?
format fails with read-only block 0 when erase() is just
```C
    int fd = open(emu->path, O_CREAT|O_TRUNC|O_NOFOLLOW|O_WRONLY, 0644);        
    if (fd >= 0 && close(fd) == 0) {                                            
        emu->stats.erase_count += 1;                                            
        return 0;                                                               
    }                                                                           
    return -errno;
```
I thought that an erase of a "bad" block (i.e. RO) have to get us a "good" block (i.e. WR) block.
At least that's the effect of unlink() in your current code.
What am i missing?

aldot: But generally speaking, switching just erase() to fd-based ops doesn't make sense.
If we don't question the usefulness of wasting time on emubd in the first place for a second, all of it should be fd-based and not FILE-based. The latter is very, very complicated -- think about all that buffering and behind the scenes maintenance -- and doesn't make sense for something like emubd i'd say. No?

geky: It's not the prettiest code, but that's what the if statement around unlink is for. If the file is not writeable it would not try to unlink.
``` c
if (!err && S_ISREG(st.st_mode) && (S_IWUSR & st.st_mode)) {
    err = unlink(emu->path);
    ...
```

The purpose of using unlink was to indicate to the `lfs_emubd_prog` function that the block was actually erased, but it might be possible to replace that logic with a check against the file size.

https://github.com/aldot/littlefs/blob/962b0ea5ad941925737be17c0d69404d630b6a1d/emubd/lfs_emubd.c#L140-L146

aldot:
For a library the API itself indeed has value.
I was just silencing the test-template so folks aren't distracted by mere noise and have a chance to actually **see** potentially useful warnings.

I do not dare to suggest to add -Wextra to the default travis cflags. I suppose they'd be there already if you would see value in 'em :)

tests/template.fmt
```diff
@@ -37,7 +37,8 @@ void test_assert(const char *file, unsigned line,
 
 
 // utility functions for traversals
-int test_count(void *p, lfs_block_t b) {{
+static int __attribute__((used)) test_count(void *p,
+	lfs_block_t b __attribute__((unused))) {{
```
geky: Thoughts on using a `(void)b;` statement instead? Would be portable across compilers. (Though I realize this file is already gcc specific). I can update it in a bit if you're not opposed.

aldot:
On 4 February 2018 02:34:45 CET, Christopher Haster <notifications@github.com> wrote:
>geky commented on this pull request.
>
>
>
>> @@ -37,7 +37,8 @@ void test_assert(const char *file, unsigned line,
> 
> 
> // utility functions for traversals
>-int test_count(void *p, lfs_block_t b) {{
>+static int __attribute__((used)) test_count(void *p,
>+	lfs_block_t b __attribute__((unused))) {{
>
>Thoughts on using a `(void)b;` statement instead? Would be portable
>across compilers. (Though I realize this file is already gcc specific).
>I can update it in a bit if you're not opposed.

Sure, please go ahead!
thanks,

geky:
Was able to knock out the rest of the warnings too. Will merge once the tests pass.

Thanks for the pr!

geky:
Hi @aldot, we're looking at changing the license to BSD for better compatibility with GPL v2
https://github.com/ARMmbed/littlefs/issues/59

Since you're a contributor, would you be able to let us know it's ok for us to relicense your contributions? You should have an email about this from your commit email, let me know if you didn't get it.

aldot:
On Tue, 12 Jun 2018 at 04:40, Christopher Haster
<notifications@github.com> wrote:
>
> Hi @aldot, we're looking at changing the license to BSD for better compatibility with GPL v2
> #59
>
> Since you're a contributor, would you be able to let us know it's ok for us to relicense your contributions? You should have an email about this from your commit email, let me know if you didn't get it.


I'm fine with relicensing my previous and future contributions to
littlefs under a BSD license.
Let me state, though, that it's a pity that littlefs is not licensed
under (L)GPL.

kind regards,
Bernhard

aldot:
Oh btw, did you think about supporting XIP (eXecute in place)?
Just curious..

jlovejoy:
@aldot - can I ask you to confirm the re-licensing via the email I sent? That would be really helpful!
thanks!

PR #25, opened 2018-02-05
added a .gitignore

Sim4n6:
Based on the repository *A collection of useful .gitignore templates*, I added a .gitignore that will tell git to  explicitely ignore some files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,55 @@
+# Prerequisites
+*.d
+
+# Object files
+*.o
+*.ko
+*.obj
+*.elf
+
+# Linker output
+*.ilk
+*.map
+*.exp
+
+# Precompiled Headers
+*.gch
+*.pch
+
+# Libraries
+*.lib
+*.a
+*.la
+*.lo
+
+# Shared objects (inc. Windows DLLs)
+*.dll
+*.so
+*.so.*
+*.dylib
+
+# Executables
+*.exe
+*.out
+*.app
+*.i*86
+*.x86_64
+*.hex
+
+# Debug files
+*.dSYM/
+*.su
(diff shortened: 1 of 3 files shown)


geky:
Hi @Sim4n6, thanks for the pr!

Sorry for responding so late. I've been juggling a few things.

Also sorry if my questions are a bit straightforward. 

First question: Did you mean to make a pr against your master branch? This means all the commits you create on your master branch will end up here. Did you meant to make a separate pr for the Makefile improvements changes?

Second question: What's the value of a gitignore? I'd be ok with a gitignore that covers what make generates (.o, .d, blocks), but it seems impossible for a gitignore to cover everything an embedded toolchain could throw at it. And a gitignore seems like it would discourage users from manually controlling what they commit.

Third question: Do you think it's possible to select a compilation target based on the symbol main? What if I added a file "main.c"? If not I think your approach with checking for test.c is the best idea (although we should add main.c as an option as well).

Sim4n6:
Hi @geky well, I'd answer you straighforward too :+1: 

First, thank you for your detailed questions, that will help me to fix it more easily.  

*Answer 1* : oops, I really pull requested only the .gitignore. I will create another dedicated branch for Makefile and test there. Sorry again.

*Answer 2* : Well, the .gitignore suggested is an adapted template. It already covers *.o, .d, and blocks directory*. And I will re-edit it so that it is not so for general purpose use. 

*Answer 3* : IMHO, I think both approaches fit here. I am still working on the right Makefile, so apologize for the duplicate PR #25. I will drop it and made another one with a more suited .gitignore. Makefile can wait :+1:

Sim4n6:
#26

PR #26, opened 2018-02-08
Added a .git ignore file

Sim4n6:
Well , It simply contains the ignore for a created files such as : .o .d blocks and the bin file lfs.

Maintainers can edit it since it is allowed on this PR

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,9 @@
+# Compilation output
+*.o
+*.d
+*.a
+
+# Testing things
+blocks/
+lfs
+test.c


geky:
Thanks for simplifying the .gitignore. 👍 

I added some other things that the current makefile might spit out. I'm thinking that is a good line to draw for what the .gitignore will cover.

I also went ahead and rebased your pr so this history is a bit cleaner. Let me know if this caused any issues for you.

Let me know if you're ok with the pr as it is now. It should be good to go.

Sim4n6:
:+1: it is okey for the pr, please proceed

geky:
Thanks for the pr!

geky:
Hi @Sim4n6, we're looking at changing the license to BSD for better compatibility with GPL v2
https://github.com/ARMmbed/littlefs/issues/59

Since you're a contributor, would you be able to let us know if it's ok for us to relicense your contributions? You should have an email about this from your commit email, let me know if you didn't get it.

Sim4n6:
Just curious what is the difference btw the two licenses??

jlovejoy:
BSD-3-Clause is compatible with pretty much all other open source licenses. thus enabling combining code from a BSD-3-Clause licensed project with projects using other (open source) licenses.

The Free Software Foundation considers Apache-2.0 to be incompatible with GPLv2 - see https://www.gnu.org/licenses/license-list.en.html#apache2 - which means it makes it hard for code under Apache-2.0 to be combined with code under GPLv2. 

Removing this obstacle is the goal of the license change.

PR #31, opened 2018-02-20
implement generic attribute implementation

guillaumerems:
add example with timestamp attribute

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ script:
 
   - make clean test QUIET=1 CFLAGS+="-DLFS_NO_INTRINSICS"
 
-  # compile and find the code size with the smalles configuration
+  # compile and find the code size with the smallest configuration
   - make clean size
         OBJ="$(ls lfs*.o | tr '\n' ' ')"
         CFLAGS+="-DLFS_NO{ASSERT,DEBUG,WARN,ERROR}"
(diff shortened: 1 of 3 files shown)


loboris:
@guillaumerems 

Could yo update this PR for the latest littlefs commits.

luc-github:
@geky I see timestamp is part of this PR - which is very nice feature 
Cannot see why travis failed
Do you think it can be integrated soon or is there any other plan for it ?
Thank you ^_^

geky:
Hi, sorry about late response (almost a year :( ). As I understand this is superseded by custom attributes in v2 (https://github.com/ARMmbed/littlefs/pull/85), which has been merged.

Here's how you might be able to create timestamps with a custom file wrapper:
``` c
#define ATTR_TIMESTAMP 0x74

typedef struct time_file {
    lfs_file_t file;
    uint32_t timestamp;
    struct lfs_attr attrs[1];
    struct lfs_file_config cfg;
} time_file_t;

int time_file_open(lfs_t *lfs, time_file_t *file, char *path, int flags) {
    // set up description of timestamp attribute
    file->attrs[0].type = ATTR_TIMESTAMP;
    file->attrs[0].buffer = &file->timestamp;
    file->attrs[0].size = sizeof(&file->timestamp);

    // set up config to indicate file has custom attributes
    memset(&file->cfg, 0, sizeof(file->cfg));
    file->cfg->attrs = &file->attrs;
    file->cfg->attr_count = 1;

    // attributes will be automatically populated during open call
    return lfs_file_opencfg(lfs, &file->file, path, flags, &file->cfg);
}

int time_file_write(lfs_t *lfs, struct time_file *file, void *buffer, size_t size) {
    // update timestamp
    file->timestamp = now(NULL);
    lfs_file_write(file->lfs, &file->file, buffer, size);

    // note that like file data, custom attributes are not actually written to disk until file sync or file close
}
```

Related https://github.com/ARMmbed/littlefs/issues/23.

I'm going to go ahead and close this. Feel free to create another issue if I've missed anything.

PR #48, opened 2018-04-08
Add custom attributes, inline files, and resizable entries

geky:
Minor version bump to v1.4 - API is backwards compatible
Minor disk version bump to v1.2 - Disk structures are backwards compatible (can be upgraded)

This adds internal resizable entries, which enables inline files and custom attributes. Two features that have been heavily requested.

Note: Currently this pr needs a lot of testing! ~And big-endian support needs to be fixed~.

What's new:
- Internal support for resizing entries, this is a building block for new features
- Inline files

  Now small files (<1024B) can be inlined directly in the directory block instead of getting their own block, which can waste a lot of space on devices with large block sizes.

  Note: Because inline files must be entirely stored in a files cache, a configurable _inline_size_ attribute is written to the superblock at **format** time. If the device has less RAM and a smaller _read_size_, the filesystem won't be able to mount. By default _inline_size_ = _read_size_.

- Custom attributes

  It's now possible to get and set custom attributes on files, directories, and the superblock.

  This isn't quite the same as getxattr/setxattr, instead uses a byte identifier to lookup attributes. Additionally the API is a bit different to avoid the cost of buffering file attributes and enable atomic updates to files.

  More info over here: https://github.com/geky/littlefs/issues/23

- A function to get a count of the used blocks on the filesystem: `lfs_fs_size`

  Note: this is just a wrapper over `lfs_traverse`, but should be easier to use

Related issues:
- https://github.com/geky/littlefs/issues/23 - custom attributes
- https://github.com/geky/littlefs/pull/31 - time as custom attribute (alternative solution)
- https://github.com/geky/littlefs/issues/22 - alen questions
- https://github.com/geky/littlefs/issues/29 - nand devices (inline files can help)
- https://github.com/geky/littlefs/issues/11 - internal flash devices (inline files can help)
- https://github.com/geky/littlefs/issues/45 - available space question

TODO:

- [x] Fix big-endian support
- [ ] Fix littlefs-fuse support
- [x] Prevent reading attributes with bad length
- [x] Add tests for resizable entries
- [x] Add tests for inline files
- [x] Add tests for custom attributes
- [ ] Update DESIGN.md with design changes
- [ ] Update SPEC.md with spec changes
- [ ] Update README.md?
- [ ] Clean up commit history

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ script:
   - make test QUIET=1 CFLAGS+="-DLFS_READ_SIZE=512    -DLFS_PROG_SIZE=512"
   - make test QUIET=1 CFLAGS+="-DLFS_BLOCK_COUNT=1023 -DLFS_LOOKAHEAD=2048"
 
+  - make clean test QUIET=1 CFLAGS+="-DLFS_INLINE_MAX=0"
   - make clean test QUIET=1 CFLAGS+="-DLFS_NO_INTRINSICS"
 
   # compile and find the code size with the smallest configuration
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ script:
   # compile and find the code size with the smallest configuration
   - make clean size
         OBJ="$(ls lfs*.o | tr '\n' ' ')"
-        CFLAGS+="-DLFS_NO{ASSERT,DEBUG,WARN,ERROR}"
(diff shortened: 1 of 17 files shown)


geky:
Closing as this has been superseded by v2 (https://github.com/ARMmbed/littlefs/pull/85)

PR #58, opened 2018-06-08
Added possibility to open multiple files with LFS_NO_MALLOC enabled

dpgeorge:
The `LFS_NO_MALLOC` option is very useful, because some systems have very particular memory allocation needs.  But enabled this option doesn't allow to have multiple files open at the same time.  This PR adds a quick fix for this by allowing to pass in the `file_buffer` directly to `lfs_file_open()`, so the latter doesn't need to call `lfs_malloc()`.

I don't expect this to be merged as-is because it breaks the existing API.  But a feature along these lines would be very useful because it gives full functionality of littlefs without the need for any `malloc()` or `free()` functions.

Alternatives to this PR would be:
- a new function, eg `lfs_file_open_using()` which accepts the 5th parameter added here
- allow to set the global `lfs->cfg->file_buffer` pointer before each call to `lfs_file_open()` so the latter could pick up a new buffer each time

( Background: we are looking to integrate littlefs in MicroPython: https://github.com/micropython/micropython/pull/3847 )

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -1282,8 +1282,9 @@ static int lfs_ctz_traverse(lfs_t *lfs,
 
 
 /// Top level file operations ///
-int lfs_file_open(lfs_t *lfs, lfs_file_t *file,
-        const char *path, int flags) {
+int lfs_file_opencfg(lfs_t *lfs, lfs_file_t *file,
+        const char *path, int flags,
+        const struct lfs_file_config *cfg) {
     // deorphan if we haven't yet, needed at most once after poweron
     if ((flags & 3) != LFS_O_RDONLY && !lfs->deorphaned) {
         int err = lfs_deorphan(lfs);
@@ -1323,6 +1324,7 @@ int lfs_file_open(lfs_t *lfs, lfs_file_t *file,
     }
 
     // setup file struct
+    file->cfg = cfg;
     file->pair[0] = cwd.pair[0];
     file->pair[1] = cwd.pair[1];
     file->poff = entry.off;
@@ -1340,7 +1342,10 @@ int lfs_file_open(lfs_t *lfs, lfs_file_t *file,
     }
 
     // allocate buffer if needed
-    if (lfs->cfg->file_buffer) {
+    file->cache.block = 0xffffffff;
+    if (file->cfg && file->cfg->buffer) {
+        file->cache.buffer = file->cfg->buffer;
+    } else if (lfs->cfg->file_buffer) {
         if (lfs->files) {
             // already in use
             return LFS_ERR_NOMEM;
@@ -1368,6 +1373,11 @@ int lfs_file_open(lfs_t *lfs, lfs_file_t *file,
     return 0;
 }
 
+int lfs_file_open(lfs_t *lfs, lfs_file_t *file,
+        const char *path, int flags) {
+    return lfs_file_opencfg(lfs, file, path, flags, NULL);
+}
+
 int lfs_file_close(lfs_t *lfs, lfs_file_t *file) {
     int err = lfs_file_sync(lfs, file);
 
@@ -1380,7 +1390,7 @@ int lfs_file_close(lfs_t *lfs, lfs_file_t *file) {
     }
 
     // clean up memory
-    if (!lfs->cfg->file_buffer) {
+    if (!(file->cfg && file->cfg->buffer) && !lfs->cfg->file_buffer) {
         lfs_free(file->cache.buffer);
     }
 
--- a/lfs.h
+++ b/lfs.h
@@ -21,7 +21,7 @@ extern "C"
 // Software library version
 // Major (top-nibble), incremented on backwards incompatible changes
 // Minor (bottom-nibble), incremented on feature additions
-#define LFS_VERSION 0x00010004
+#define LFS_VERSION 0x00010005
 #define LFS_VERSION_MAJOR (0xffff & (LFS_VERSION >> 16))
 #define LFS_VERSION_MINOR (0xffff & (LFS_VERSION >>  0))
 
@@ -167,6 +167,12 @@ struct lfs_config {
     void *file_buffer;
 };
 
+// Optional configuration provided during lfs_file_opencfg
+struct lfs_file_config {
+    // Optional, statically allocated buffer for files. Must be program sized.
+    // If NULL, malloc will be used by default.
+    void *buffer;
+};
 
 // File info structure
 struct lfs_info {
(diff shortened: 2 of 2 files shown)


dpgeorge:
I have rebased this PR on top of the latest master, with change in license.

geky:
Hi @dpgeorge, thanks for the pr (and rebasing). I think you're right passing a buffer into open would be a better API.

Tossing ideas around, one idea to make this future compatible would be to take in a sort of config struct. Similar to the lfs_mount/lfs_format functions:
``` cpp
struct lfs_file_config {
    void *buffer;
};

int lfs_file_open(lfs_t *lfs, lfs_file_t *file,
        const char *path, int flags,
        const struct lfs_file_config *cfg);
```

This would let us add additional configuration to the config structure in the future without breaking compatibility or having to come up with additional function names. Thoughts?

dpgeorge:
> one idea to make this future compatible would be to take in a sort of config struct.

It might be simpler to just add a config entry to the `lfs_file_t` struct, which the user can populate if needed.  Then there doesn't need to be any additional args to `lfs_file_open`.

Alternatively, a user could just set the existing `file->cache.buffer` entry manually before calling `lfs_file_open`, in which case the user's buffer would be used.  That also doesn't require adding any args to the function.

geky:
> It might be simpler to just add a config entry to the lfs_file_t struct, which the user can populate if needed. Then there doesn't need to be any additional args to lfs_file_open.

Hmm, although wouldn't that no longer be backwards compatible since the config entry would need to be zeroed by users that don't want to use a buffer?

geky:
Those are all good ideas.

I believe the main benefit of providing a separate config struct is that it makes it clear to the user what needs to be initialized and what doesn't. And additionally the config struct could be stored in 

How about we go the "lfs_file_open_ext" route and add an additional open function as `lfs_file_opencfg`? The simpler `lfs_file_open` function can still be there.

---

One thing I'm considering is the addition of custom attributes (https://github.com/ARMmbed/littlefs/pull/48). Introducing a config struct (or setting members of the file struct) would open up a good way to introduct custom attributes before file open:

``` c
struct my_file {
    lfs_file_t file;
    struct lfs_file_config cfg;
    struct lfs_attr attrs[2];
    uint32_t timestamp;
    uint16_t number;
};

struct my_file file;
file->attrs[0].type = MY_TYPE_TIMESTAMP;
file->attrs[0].size = sizeof(file->timestamp);
file->attrs[0].buffer = &file->timestamp;
file->attrs[1].type = MY_TYPE_NUMBER;
file->attrs[1].size = sizeof(file->number);
file->attrs[1].buffer = &file->number;
file->cfg.buffer = malloc(128);
file->cfg.attrs = &file->attrs;
file->cfg.attr_count = 2;

int err = lfs_file_opencfg(lfs, &file->file, "filename.txt", LFS_O_RDONLY, &file->cfg);
```

dpgeorge:
> How about we go the "lfs_file_open_ext" route and add an additional open function as lfs_file_opencfg? The simpler lfs_file_open function can still be there.

That sounds good.

> I believe the main benefit of providing a separate config struct is that it makes it clear to the user what needs to be initialized and what doesn't.

To allow to add entries to the file config struct in a backwards-compatible way, I guess you'd require the user to always do `memset(&file->cfg, 0, sizeof(file->cfg))`, then populate the used entries, before calling `lfs_file_opencfg()`.

For `const` file configuration structs this would still work because (in C99 at least) you can use named initialisers and the remaining entries will be zero'd out, eg:
```c
static uint8_t my_file_buffer[128];
static const struct lfs_file_config my_file_cfg = {
    .file_buffer = my_file_buffer,
    // all other entries are 0 by default
};
```

> One thing I'm considering is the addition of custom attributes (#48). Introducing a config struct (or setting members of the file struct) would open up a good way to introduct custom attributes before file open

Also sounds good.  Although in your example above, if you open the file with `LFS_O_RDONLY` I assume it'll ignore the attributes in the file config, because it can't write to the file?

geky:
@dpgeorge How does this look to you? I added the `lfs_file_opencfg` onto your pr.

One other question is if these config structs should be copied into the structures during format/mount/opencfg:
1. Copy
  - RAM only allocated for format/mount/opencfg call
  - Allows modifications by format/mount/opencfg
2. Reference by pointer
  - Allows storing config in ROM

I went with the second option for consistency with `lfs_mount`.

> Also sounds good. Although in your example above, if you open the file with LFS_O_RDONLY I assume it'll ignore the attributes in the file config, because it can't write to the file?

It could still read the attributes, though there wouldn't be much advantage over and explicit `lfs_getattr` call.

dpgeorge:
> How does this look to you? I added the lfs_file_opencfg onto your pr.

Thanks, I see it.

> One other question is if these config structs should be copied into the structures during format/mount/opencfg:
> ...
> I went with the second option for consistency with lfs_mount.

That brings up an interesting point: for files there would be two kinds of configurations: global configuration options that are common to all files (like attributes), and local configuration options just for a file (like the cache buffer).  If these are put into the same struct then it would be a waste of memory to allocate a new one of these structs for each file, since all the global config would be common and could have gone in ROM.  To be explicit on this point: the `void *buffer` entry in `lfs_file_config` must be different for each file and so that requires a different instance of this struct for each file that you want open at the same time.  And then you must copy into each of these structs the common global file config values, which could be quite large.  Regardless of if these file config structs go in ROM or RAM it's still wasteful repeating this data.

Maybe the common, global file config (like configuration of attributes and their types) needs to go in the `lft_config` struct.  Then only per-file config goes in `lft_file_config`.

geky:
You're right that any common configuration should go into the global struct.

But I attributes fall into this category. A part of the attribute structure is the buffer that contain's the attribute's data. This would be unique for each file.

There's also the minor point that you may want different attributes for different files, but I'm not sure that's important.

On the plus side, you only pay the cost for the attribute structures if you use custom attributes.

geky:
Rebased due to conflict with master and added version bump to v1.5

Hi @dpgeorge, does this change look good to you? Since it's backwards compatible I think we can go ahead and merge it without any problems (just a minor version bump).

dpgeorge:
Yes it looks good to me, thanks!  It does look backwards compatible.

geky:
Thanks for the initial pr!

dpgeorge:
No problem.  There's not much left of my initial code/commit, I don't know if it's worth keeping it around or not.

geky:
Sure thing, here we go

dpgeorge:
Ok, I'm happy with it.

dpgeorge:
Thanks for the work to get this merged, it's much appreciated!  FYI here is the change I made to support this new `lfs_file_opencfg()` function: https://github.com/micropython/micropython/pull/3847/commits/4b6bbbd134d88588319bf783233d9f46f5134d76

PR #73, opened 2018-07-11
Use PRIu32 and PRIx32 format specifiers to fix warnings

FreddieChopin:
When using `"%d"` or `"%x"` with `uint32_t` types, arm-none-eabi-gcc reports
warnings like below:

```
In file included from lfs.c:8:
lfs_util.h:45:12: warning: format '%d' expects argument of type 'int', but argument 4 has type 'lfs_block_t' {aka 'long unsigned int'} [-Wformat=]
     printf("lfs debug:%d: " fmt "\n", __LINE__, __VA_ARGS__)
            ^~~~~~~~~~~~~~~~
lfs.c:2512:21: note: in expansion of macro 'LFS_DEBUG'
                     LFS_DEBUG("Found partial move %d %d",
                     ^~~~~~~~~
lfs.c:2512:55: note: format string is defined here
                     LFS_DEBUG("Found partial move %d %d",
                                                      ~^
                                                      %ld
```

Fix this by replacing `"%d"` and `"%x"` with `"%" PRIu32` and `"%" PRIx32`.

### DIFF
--- a/emubd/lfs_emubd.c
+++ b/emubd/lfs_emubd.c
@@ -16,6 +16,7 @@
 #include <unistd.h>
 #include <assert.h>
 #include <stdbool.h>
+#include <inttypes.h>
 
 
 // Block device emulated on existing filesystem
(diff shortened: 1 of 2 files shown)


geky:
Looks good to me, thanks for this!

PR #74, opened 2018-07-13
Add C++ guards to public headers

### LINKED ISSUE #32: C++ include
Can we have a:

#ifdef __cplusplus
extern "C" {
#endif

.....

#ifdef __cplusplus
}
#endif

in the include file so it can be cleanly included by a cpp file?

### LINKED ISSUE #53: C++ support in .H files
Hi geky
I would like the .h files to have the C++ support directives, like this:

#ifdef  __cplusplus
extern "C"
{
#endif

....

#ifdef  __cplusplus
}
#endif

Cheers

FreddieChopin:
Even though it is perfectly possible to surround `#include "lfs.h"` with `extern "C" { ... }` directly in the C++ files, the common practice is to have such C++ guards directly in headers, even if this is a C library. The maintenance cost is basically zero (just remember to add includes "outside" the guards and code "inside"), while the gains are important (at least in my opinion):

1. Generally all C libraries do that, so it is a common practice, expected by people developing in C++. This is also true for C system headers like `stddef.h` (see `_BEGIN_STD_C` and `_END_STD_C` macros from newlib as an example) and so on.
2. If the header does not follow this common practice and application does not surround the `#include` with `extern "C" { ... }`, the project compiles perfectly fine, with no error and no warning. You only notice the problem during linking, but only if you actually use any of the lfs functions. You will then get a sort-of a cryptic message that `xxx is undefined`, which can mean a lot of other things beside the problem with `#include`.

Fixes #53 
Fixes #32

### DIFF
--- a/emubd/lfs_emubd.h
+++ b/emubd/lfs_emubd.h
@@ -10,6 +10,11 @@
 #include "lfs.h"
 #include "lfs_util.h"
 
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
 
 // Config options
 #ifndef LFS_EMUBD_READ_SIZE
@@ -74,5 +79,9 @@ int lfs_emubd_erase(const struct lfs_config *cfg, lfs_block_t block);
 // Sync the block device
 int lfs_emubd_sync(const struct lfs_config *cfg);
 
+
+#ifdef __cplusplus
+} /* extern "C" */
+#endif
 
 #endif
--- a/lfs.h
+++ b/lfs.h
@@ -10,6 +10,11 @@
 #include <stdint.h>
 #include <stdbool.h>
 
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
 
 /// Version info ///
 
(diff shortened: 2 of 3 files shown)


FreddieChopin:
I know that in #32 you were against such change. Generally if it would be possible to have a warning or error which would say "hey, you need to surround this `#include`  at line 12 in `xxx.cpp` with `extern "C" { ... }`!', then I would not propose this change, but the reality is as stated above.

So please reconsider, especially because from your point of view it is a win-win with zero cost in future maintenance. Unless you are allergic to C++, but I hope you're not (;

geky:
> Unless you are allergic to C++, but I hope you're not (;

It does give me a bit of a rash when I touch it : )

You're right though, and I'm just being stubborn. Though you have to admit, it's a bit strange that a new language can come along and require changes in the code of another language. Hopefully one of the next language generation forces `extern "C++"` into C++ headers. >: )

Actually once @davidefer opened his issue I was planning on adding this change, but couldn't get to it, forgot about it, and haven't been able to go through old issues yet (sorry @davidefer!). I figure if I'm outnumbered I should accept the changes even if I disagree.

Thanks for beating me too it and putting this up! LGTM

PR #80, opened 2018-07-18
Fix memory leaks

FreddieChopin:
Fix memory leaks in `lfs_init()`, `lfs_format()` and `lfs_mount()`

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -2016,6 +2016,21 @@ int lfs_rename(lfs_t *lfs, const char *oldpath, const char *newpath) {
 
 
 /// Filesystem operations ///
+static void lfs_deinit(lfs_t *lfs) {
+    // free allocated memory
+    if (!lfs->cfg->read_buffer) {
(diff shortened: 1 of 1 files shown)


geky:
Good catch. Mind if I squash your commits?

FreddieChopin:
> Good catch. Mind if I squash your commits?

It's your project (; If you prefer to have them squashed, then no problem with me (;

> Bikeshedding: Could I suggest the name "cleanup"?

Certainly! I was wondering about a good name myself, "cleanup" is indeed nicer (; I've modified the commits.

geky:
Well, on GitHub it does end up changing _your_ branch when I do this, which is unfortunate. Thanks for this!

FreddieChopin:
> Well, on GitHub it does end up changing your branch when I do this, which is unfortunate.

Ha - didn't know about that (; No problem anyway, if I would really like to keep them (no idea what for) I always have my local copy.

PR #81, opened 2018-07-25
Simplified release process based on feedback

geky:
Previously, littlefs had mutable versions. That is, anytime a new commit landed on master, the bot would update the most recent version to contain the patch. The idea was that this would make sure users always had the most recent bug fixes. Immutable snapshots could be accessed through the git hashes.

However, at this point multiple developers have pointed out that this is confusing, with mutable versions being non-standard and surprising.

This new release process adopts SemVer in its entirety, with incrementing patch numbers and immutable versions.

When a new commit lands on master:
1. The major/minor version is taken from lfs.h
2. The most recent patch version is looked up on GitHub and incremented
3. A changelog is built out of the commits to the previous version
4. A new release and tag is created on GitHub

Additionally, any commits that land while CI is still running are coalesced together. Which means multiple PRs can land in a single release.

I've been testing the new script out over on this repo, if anyone wants to mess around with it:
https://github.com/geky/littlefs-test-repo/releases

Let me know your thoughts.

cc @FreddieChopin, @dpgeorge, @davidefer, @bridadan, @theotherjimmy, @kegilbert, @deepikabhavnani
should resolve https://github.com/ARMmbed/littlefs/issues/79, https://github.com/ARMmbed/littlefs/issues/50

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -134,53 +134,44 @@ jobs:
         - STAGE=deploy
         - NAME=deploy
       script:
-        # Update tag for version defined in lfs.h
+        # Find version defined in lfs.h
         - LFS_VERSION=$(grep -ox '#define LFS_VERSION .*' lfs.h | cut -d ' ' -f3)
         - LFS_VERSION_MAJOR=$((0xffff & ($LFS_VERSION >> 16)))
         - LFS_VERSION_MINOR=$((0xffff & ($LFS_VERSION >>  0)))
-        - LFS_VERSION="v$LFS_VERSION_MAJOR.$LFS_VERSION_MINOR"
-        - echo "littlefs version $LFS_VERSION"
+        # Grab latests patch from repo tags, default to 0
+        - LFS_VERSION_PATCH=$(curl -f -u "$GEKY_BOT_RELEASES"
+                https://api.github.com/repos/$TRAVIS_REPO_SLUG/git/refs
+                | jq 'map(.ref | match(
+                    "refs/tags/v'"$LFS_VERSION_MAJOR"'\\.'"$LFS_VERSION_MINOR"'\\.(.*)$")
+                    .captures[].string | tonumber + 1) | max // 0')
+        # We have our new version
+        - LFS_VERSION="v$LFS_VERSION_MAJOR.$LFS_VERSION_MINOR.$LFS_VERSION_PATCH"
+        - echo "VERSION $LFS_VERSION"
         - |
-          curl -u $GEKY_BOT_RELEASES -X POST \
-            https://api.github.com/repos/$TRAVIS_REPO_SLUG/git/refs \
-            -d "{
-              \"ref\": \"refs/tags/$LFS_VERSION\",
-              \"sha\": \"$TRAVIS_COMMIT\"
-            }"
-        - |
-          curl -f -u $GEKY_BOT_RELEASES -X PATCH \
-            https://api.github.com/repos/$TRAVIS_REPO_SLUG/git/refs/tags/$LFS_VERSION \
-            -d "{
-              \"sha\": \"$TRAVIS_COMMIT\"
-            }"
-        # Create release notes from commits
-        - LFS_PREV_VERSION="v$LFS_VERSION_MAJOR.$(($LFS_VERSION_MINOR-1))"
-        - |
-          if [ $(git tag -l "$LFS_PREV_VERSION") ]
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -134,53 +134,44 @@ jobs:
...
       script:
-        # Update tag for version defined in lfs.h
+        # Find version defined in lfs.h
         - LFS_VERSION=$(grep -ox '#define LFS_VERSION .*' lfs.h | cut -d ' ' -f3)
         - LFS_VERSION_MAJOR=$((0xffff & ($LFS_VERSION >> 16)))
         - LFS_VERSION_MINOR=$((0xffff & ($LFS_VERSION >>  0)))
-        - LFS_VERSION="v$LFS_VERSION_MAJOR.$LFS_VERSION_MINOR"
-        - echo "littlefs version $LFS_VERSION"
+        # Grab latests patch from repo tags, default to 0
+        - LFS_VERSION_PATCH=$(curl -f -u "$GEKY_BOT_RELEASES"
```
dpgeorge: Is it a security concern, if someone opens a PR on this repo with a modified .travis.yml, which does an `echo $GEKY_BOT_RELEASES`, they would be able to do anything that geky-bot can do?

.travis.yml
```diff
@@ -134,53 +134,44 @@ jobs:
...
+                https://api.github.com/repos/$TRAVIS_REPO_SLUG/git/refs
+                | jq 'map(.ref | match(
+                    "refs/tags/v'"$LFS_VERSION_MAJOR"'\\.'"$LFS_VERSION_MINOR"'\\.(.*)$")
+                    .captures[].string | tonumber + 1) | max // 0')
```
dpgeorge: Would it be simpler to just use git locally to find out this info via a `git describe`?  I assume you'll add a git tag for each patch, so git describe will tell you the last patch number.

dpgeorge:
> This new release process adopts SemVer in its entirety, with incrementing patch numbers and immutable versions.

That sounds great!

> 4. A new release is created on GitHub

Will you also create git tags for each patch, eg tag v1.5.2?  That would help users to find a specific patch version if they check out a local copy of the repo.  Otherwise where is this information stored?

.travis.yml
geky: Good question, GEKY_BOT_RELEASES is stored in a secure variable in Travis ([more info](https://docs.travis-ci.com/user/environment-variables/#defining-variables-in-repository-settings)). This offers a few protections. Most notably, secure variables are not available on PRs, only on "trusted" builds. So you can't actually get GEKY_BOT_RELEASES on PRs.

Also, I (unintentionally) found out that Travis filters out any secure variables if they get printed accidentally. You could still figure out the variables with a sort of dictionary attack, but still, that would have to land on master.

And finally, the actual value of GEKY_BOT_RELEASES is a GitHub token of the form "geky-bot:12341234123412341234123412341234". So it's easy to disable if it does get compromised.

As a sidenote, this is why GEKY_BOT_STATUSES and GEKY_BOT_RELEASES are two different variables. GEKY_BOT_STATUSES is a separate GitHub token that only has permissions to update commit statuses and is available on PRs for reporting CI results.

kegilbert:
LGTM

.travis.yml
geky: It would be simpler, but querying the tags on GitHub provides a bit more of a catch-all.  It garuntees that all patch numbers are unique even if someone messes with the git history (cough force-pushing cough).

It could also garuntee uniqueness across multiple branches, but I can't think of a reason you would do that.

geky:
> Will you also create git tags for each patch, eg tag v1.5.2? That would help users to find a specific patch version if they check out a local copy of the repo. Otherwise where is this information stored?

Ah! Creating a release in GitHub automatically creates a git tag as well. So all releases will have a `vX.Y.Z` tag. Except the existing releases. I'll probably just leave those as `vX.Y` as they are right now.

.travis.yml
dpgeorge: Ok, thanks for the explanation, that's useful to know!

.travis.yml
dpgeorge: Ok.

dpgeorge:
> Creating a release in GitHub automatically creates a git tag as well. So all releases will have a vX.Y.Z tag

Ok, that sounds good then.  As long as the git history has a tag, and not just github.

geky:
Thanks for the reviews!

PR #101, opened 2018-09-16
Does it mean reading data align lfs->cfg->read_size ?

AubrCool:
Maybe ,this is the correct way!


geky:
Related https://github.com/ARMmbed/littlefs/pull/102

geky:
I'm going to close this as I believe it's the correct behaviour. Feel free to reopen if you disagree.

PR #102, opened 2018-09-16
does it mean write align the  lfs->cfg->prog_size?

AubrCool:
Maybe it is the correct way!


geky:
Hi @AubrCool, thanks for the PR.

If I understand this PR correctly, the current math is by design.

LittleFS may call the block devices with > read/prog sizes, but the data will at least be a multiple of the read/prog sizes. This allows block devices to write out larger chunks of data more efficiently.

https://github.com/ARMmbed/littlefs/issues/96 may have more info.

geky:
I'm going to close this as I believe it's the correct behaviour. Feel free to reopen if you disagree.

PR #106, opened 2018-09-26
If stats file doesn't exist lfs_emubd_create will fail.

conkerkh:
This will create default stats file if it doesn't exist.

### DIFF
--- a/emubd/lfs_emubd.c
+++ b/emubd/lfs_emubd.c
@@ -47,19 +47,24 @@ int lfs_emubd_create(const struct lfs_config *cfg, const char *path) {
 
     // Load stats to continue incrementing
     snprintf(emu->child, LFS_NAME_MAX, "stats");
+
     FILE *f = fopen(emu->path, "r");
-    if (!f) {
+    if (!f && errno != ENOENT) {
         return -errno;
     }
 
-    size_t res = fread(&emu->stats, sizeof(emu->stats), 1, f);
-    if (res < 1) {
-        return -errno;
-    }
+    if (errno == ENOENT) {
+        memset(&emu->stats, 0x0, sizeof(emu->stats));
+    } else {
+        size_t res = fread(&emu->stats, sizeof(emu->stats), 1, f);
+        if (res < 1) {
+            return -errno;
+        }
 
-    err = fclose(f);
-    if (err) {
-        return -errno;
+        err = fclose(f);
+        if (err) {
+            return -errno;
+        }
     }
 
     return 0;


geky:
Ah, it was working because I'm not checking lfs_emubd_create's return code, and the test framework declares the emubd globally, which should always be zero on the first run. Are you using the emubd outside of the test framework?

conkerkh:
@geky Yep I wrote UT using google framework as rest of my UT are implemented this way so it was quicker easier for mu to use in CI. I was indeed checking return code, also OSX is very picky in my case so everything has to be perfect here... I will make all the changes you requested:)

emubd/lfs_emubd.c
```diff
@@ -47,19 +47,33 @@ int lfs_emubd_create(const struct lfs_config *cfg, const char *path) {
...
     FILE *f = fopen(emu->path, "r");
-    if (!f) {
+    if (!f && errno != ENOENT) {
         return -errno;
     }
-
-    size_t res = fread(&emu->stats, sizeof(emu->stats), 1, f);
-    if (res < 1) {
-        return -errno;
+    else if (errno == ENOENT) {
```
geky: I know I'm being nitpicky, but it helps to have a consistent style. Could you adopt either:
``` c
if (blah) {
    return 0;
}

if (blah) {
    blah;
}
```

or

``` c
if (blah) {
    return 0;
} else if (blah) {
    blah;
}
```

I can make this change when squashing the PR if you don't.

emubd/lfs_emubd.c
```diff
@@ -47,9 +47,25 @@ int lfs_emubd_create(const struct lfs_config *cfg, const char *path) {
...
-    if (!f) {
+    if (!f && errno != ENOENT) {
         return -errno;
+    } if (errno == ENOENT) {
+        f = fopen(emu->path, "wb");
+        if (!f) {
+        	return -errno;
+        }
+        memset(&emu->stats, 0x0, sizeof(emu->stats));
+        err = fwrite(&emu->stats, sizeof(emu->stats), 1, f);
```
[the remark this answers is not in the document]
geky: Sorry if it wasn't clear, but we don't need this `fwrite`. Zeroing stats should be fine.

``` c
if (errno == ENOENT) {
    memset(&emu->stats, 0x0, sizeof(emu->stats));
} else {
     // read stats
}
```

conkerkh:
Ok I hope that's all haha

emubd/lfs_emubd.c
```diff
@@ -47,19 +47,27 @@ int lfs_emubd_create(const struct lfs_config *cfg, const char *path) {
...
-    }
 
-    size_t res = fread(&emu->stats, sizeof(emu->stats), 1, f);
-    if (res < 1) {
+    FILE *f = fopen(emu->path, "r");
+    if (!f && errno != ENOENT) {
         return -errno;
-    }
+    } else if (errno == ENOENT) {
+        f = fopen(emu->path, "wb");
```
geky: Thanks for all the changes :+1: 

One last note, we no longer need to open/close the file on this branch.

geky:
Hi @conkerkh, I modified the PR to include the last change I was asking for. I hope that's ok. I'm making a release and was hoping this could make it in.

Thanks for the fix :+1:

PR #131, opened 2018-12-10
Make _tell and _size use const params

earlephilhower:
Since tell and size operations do not modify the underlying file,
these can be made const to avoid trouble interfacing with filesystems
where the file.tell() and file.size() operations are const.

This was hit while I was porting littlefs to the ESP8266 where the
File base class expects these operations to be const methods.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -1083,7 +1083,7 @@ int lfs_dir_seek(lfs_t *lfs, lfs_dir_t *dir, lfs_off_t off) {
     return 0;
 }
 
-lfs_soff_t lfs_dir_tell(lfs_t *lfs, lfs_dir_t *dir) {
+lfs_soff_t lfs_dir_tell(const lfs_t *lfs, const lfs_dir_t *dir) {
     (void)lfs;
     return dir->pos;
 }
@@ -1807,7 +1807,7 @@ int lfs_file_truncate(lfs_t *lfs, lfs_file_t *file, lfs_off_t size) {
     return 0;
 }
 
-lfs_soff_t lfs_file_tell(lfs_t *lfs, lfs_file_t *file) {
+lfs_soff_t lfs_file_tell(const lfs_t *lfs, const lfs_file_t *file) {
     (void)lfs;
     return file->pos;
 }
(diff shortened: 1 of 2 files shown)


FreddieChopin:
It seems that `lfs_t *lfs` in these functions could me made `const lfs_t *lfs` too.

earlephilhower:
Yes, let me update the PR later today.  Our filesystem base class doesn't actually make the method itself a const one so I didn't catch it.

geky:
Thanks for the pr! Sorry for such a late response.

Wouldn't marking these functions as const prevent filesystems from using `seek` to provide them? A lot of filesystems don't provide tell/size and instead expect you to use `seek`. It's even possible (though uncommon) that tell/size need to look at the file's data structure on disk to figure out the size of the file, which is why they could return errors.

FreddieChopin:
> Wouldn't marking these functions as const prevent filesystems from using seek to provide them? A lot of filesystems don't provide tell/size and instead expect you to use seek. It's even possible (though uncommon) that tell/size need to look at the file's data structure on disk to figure out the size of the file, which is why they could return errors.

Isn't it completely opposite here? littlefs is the filesystem which may be wrapped with something higher-level. If this "higher-level something" doesn't use `const` with these arguments then there's no problem at all, as you can always do "non-const" -> "const".

geky:
C++ does let you go "const" -> "non-const" with the [mutable](https://en.cppreference.com/w/cpp/language/cv) modifier. This doesn't exist in C but apparently casting away `const` is fine if you never modify it ([ref](https://stackoverflow.com/a/40968381)).

For littlefs this question comes up because we may end up wanting to change the internal implementation to something that can't be done in a const manner. (especially considering if we're changing the file implementation for large storage improvements ;) ).

FreddieChopin:
> C++ does let you go "const" -> "non-const" with the mutable modifier. This doesn't exist in C but apparently casting away const is fine if you never modify it (ref).

In C or C++ everything is possible, but in my personal style-book casting-away `const` is pretty worrying (;

> For littlefs this question comes up because we may end up wanting to change the internal implementation to something that can't be done in a const manner. (especially considering if we're changing the file implementation for large storage improvements ;) ).

Well, then you can just change it back to non-const (;

Anyway... I did not write littlefs so I don't have any gut feelings you may have in this regard. If you think that there's a chance that tell/size would need to modify something in `lfs_t` or `lfs_file_t` then reject the PR. In C standard there's a similar `long int ftell (FILE* stream);` which is not `const` so it's not that big problem if littlefs is also not `const`.

Conceptually tell and size seem like a const operations, but I don't really like the idea of "conceptual const" where internally the data is mutable, especially on embedded where there actually is a read-only memory for `.text`. If you'd have to cast away `const` then the PR is not worth the trouble.

earlephilhower:
I'm fine closing this given @geky's concerns.  I actually ended up refactoring things to avoid the need for these `const`s a while back, and I wouldn't want to limit flexibility.  Don't want anything to get in the way of a v2-alpha branch! :)

PR #135, opened 2019-01-26
Add missing word (and reflow text)

johnlunney:


geky:
Haha, good catch. Thanks for this!

PR #149, opened 2019-02-17
Fix: Negative length should return error

ajaybhargav:
To make lfs_file_truncate inline with ftruncate function, when -ve
or size more than maximum file size is passed to function it should
return invalid parameter error. In LFS case LFS_ERR_INVAL.

Signed-off-by: Ajay Bhargav <contact@rickeyworld.info>


ajaybhargav:
Found a better cleaner way to check it.. didn't put enough thought on this. Thanks to travis-ci 👍 
Updated pull request #150

PR #150, opened 2019-02-17
Fix: length more than LFS_FILE_MAX should return error

ajaybhargav:
To make lfs_file_truncate inline with ftruncate function, when -ve
or size more than maximum file size is passed to function it should
return invalid parameter error. In LFS case LFS_ERR_INVAL.

Signed-off-by: Ajay Bhargav <contact@rickeyworld.info>

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -2864,6 +2864,10 @@ int lfs_file_truncate(lfs_t *lfs, lfs_file_t *file, lfs_off_t size) {
         return LFS_ERR_BADF;
     }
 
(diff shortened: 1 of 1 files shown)


geky:
Ah this is a good change. Thanks for creating a pr!

Need to rebase and then I will merge this need next chance I get.

ajaybhargav:
Thanks 👍 Glad the patch made it to source. Its not going to mess anything in my source as I created branch specifically for this, will wait for it to be part of release.

PR #151, opened 2019-03-01
Fixed documentation for return lfs_dir_read return value.

Krakonos:
lfs_dir_read breaks the convention of returning non-zero on success,
this feature should be at least documented.

This already caused some confusion in issue https://github.com/ARMmbed/littlefs/issues/114 . The documentation was not clear about return values. This PR clarifies the return values, including those that are not covered by standard error codes.


geky:
Thanks for this! I'm surprised it survived the >100 commit merge that was v2 without needing a rebase.

PR #179, opened 2019-05-14
Fix compilation with -Wundef

rojer:


geky:
I've not ran into this problem myself, but looks reasonable. Thanks for this!

PR #181, opened 2019-05-16
Make lfs1_crc static so it doesn't conflict with prefixed LFS1 code

rojer:
When LFS1 code is present and LFS_MIGRATE is enabled


geky:
Thanks for this!

PR #185, opened 2019-05-17
fix endless loop in dir_compact and nullptr access in lfs_dir_fetchmatch

Mohawi:
-> introduce relocateCounter in dir_compact to catch infinite loop when
   lfs has no free space left anymore
-> call lfs_dir_fetchmatch with ftag=-1 in order to set the invalid bit
   and never let the function match a dir

Fixes:

https://github.com/ARMmbed/littlefs/issues/177
https://github.com/ARMmbed/littlefs/issues/184

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -969,9 +969,9 @@ static lfs_stag_t lfs_dir_fetchmatch(lfs_t *lfs,
 
 static int lfs_dir_fetch(lfs_t *lfs,
         lfs_mdir_t *dir, const lfs_block_t pair[2]) {
-    // note, mask=-1, tag=0 can never match a tag since this
+    // note, mask=-1, tag=-1 can never match a tag since this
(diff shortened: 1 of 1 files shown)


geky:
Hi @Mohawi, thanks for creating a PR. Sorry it's taken so long for me to get to this.

I'm not sure yet the best way to handle for the possible infinite-loops. For now I've split out the null-pointer exception into a separate PR and rebased it onto master:
https://github.com/ARMmbed/littlefs/pull/337

It looks good to me, so I will merge it when I can

geky:
Ok! I've now added proper loop checking in https://github.com/ARMmbed/littlefs/pull/372. Hopefully this takes care of the issues you've ran into.

Sorry about butchering your PR like this. Thanks again for the contribution. It's very useful to know what issues/fixes everyone is looking for.

PR #194, opened 2019-05-28
Fix accidental truncate after seek on inline files

geky:
The cause was mistakenly setting file->ctz.size directly instead of
file->pos, which file->ctz.size gets overwritten with later in
lfs_file_flush.

Also added better seek test cases specifically for inline files. This
should also catch most of the inline corner cases related to
lfs_file_size/lfs_file_tell.

Should fix https://github.com/ARMmbed/littlefs/issues/172

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -194,7 +194,7 @@ static int lfs_bd_prog(lfs_t *lfs,
             off += diff;
             size -= diff;
 
-            pcache->size = off - pcache->off;
+            pcache->size = lfs_max(pcache->size, off - pcache->off);
             if (pcache->size == lfs->cfg->cache_size) {
                 // eagerly flush out pcache if we fill up
                 int err = lfs_bd_flush(lfs, pcache, rcache, validate);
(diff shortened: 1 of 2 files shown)


lfs.c
```diff
@@ -2548,7 +2548,7 @@ static int lfs_file_flush(lfs_t *lfs, lfs_file_t *file) {
                 }
             }
         } else {
-            file->ctz.size = lfs_max(file->pos, file->ctz.size);
+            file->pos = lfs_max(file->pos, file->ctz.size);
```
pabigot: @geky This line is either incorrect or useless, because `file->pos` is overwritten by `pos` on line 2560.  Should it be assigning to `pos`?

pabigot: Probably incorrect, since in what's below `file->pos` is used to replace `file.ctz.size`.

geky: Hi @pabigot, thanks for raising this. To double check is this fixed in https://github.com/ARMmbed/littlefs/pull/257?

pabigot: I didn't touch this material in #257; that fixes a different bug.  I misunderstood what this patch was doing the first couple time I read it.  As far as I know there's nothing wrong left here after this was merged.

geky: Ah ok, good to know

PR #201, opened 2019-06-07
Mark all Python 2 scripts as Python 2

nickray:
On some systems, `python` links to `python3`, so the tests fail to run.

Would there be interest in converting all these to Python 3? Since Python 2 is due for deprecation this year... I could send a PR for this. If so, is Python >=3.6 acceptable? Asking because of f-strings.

### DIFF
--- a/scripts/prefix.py
+++ b/scripts/prefix.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 
 # This script replaces prefixes of files, and symbols in that file.
 # Useful for creating different versions of the codebase that don't
--- a/tests/corrupt.py
+++ b/tests/corrupt.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 
 import struct
 import sys
--- a/tests/stats.py
+++ b/tests/stats.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 
 import struct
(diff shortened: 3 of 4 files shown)


geky:
Thanks for the PR! This breaks on Windows, but I'm not sure we care.

It looks good to me, I'll merge it with a batch of other PRs in a bit.

> Would there be interest in converting all these to Python 3? Since Python 2 is due for deprecation this year... 

Yeah, this is a good idea. I need to shake the Python 2 habit.

nickray:
> Thanks for the PR! This breaks on Windows, but I'm not sure we care.

How so? Is there a way to not break Windows?
I assumed I was just cleaning up inconsistency, since https://github.com/ARMmbed/littlefs/blob/master/tests/debug.py is already marked `python2`.

> Yeah, this is a good idea. I need to shake the Python 2 habit.

Ehehe, it's not such a big deal as I had thought before I made the switch - just bracket your `print`s and in exchange receive f-strings as reward ;)

geky:
Last time I used Python in Windows this would break because the installation did not create the "python2" and "python3" symlinks (perhaps because Windows doesn't have symlinks). But looking around for a reference I can't find anything, so it may not be an issue anymore (or a bad setup on my machine).

> Ehehe, it's not such a big deal as I had thought before I made the switch - just bracket your prints and in exchange receive f-strings as reward ;)

I remember the lack of tuple-unpacking in lambda arguments being a problem, though I think that may have been specific to the program I was working on :)

PR #213, opened 2019-06-23
Protect file operations against reusing lfs_file_t belonging to closed files.

Ar2rL:
Do not allow file operations (except lfs_file_open) on closed files by erroneous reuse lfs_file_t stucture.
It is achived by adding new internal file flag. This flag is set in lfs_file_close and it is tested in each (file related) API call as a precondition.

I also added patch to not allow using of pointers to deallocated memory (some kind of defensive programming).

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -2397,6 +2397,10 @@ int lfs_file_open(lfs_t *lfs, lfs_file_t *file,
 }
 
 int lfs_file_close(lfs_t *lfs, lfs_file_t *file) {
+    if (file->flags & LFS_F_CLOSED) {
+        return LFS_ERR_BADF;
+    }
+
     int err = lfs_file_sync(lfs, file);
 
     // remove from list of mdirs
@@ -2412,10 +2416,16 @@ int lfs_file_close(lfs_t *lfs, lfs_file_t *file) {
         lfs_free(file->cache.buffer);
     }
 
+    file->flags = LFS_F_CLOSED;
+
     return err;
 }
 
 static int lfs_file_relocate(lfs_t *lfs, lfs_file_t *file) {
+    if (file->flags & LFS_F_CLOSED) {
+        return LFS_ERR_BADF;
+    }
+
     while (true) {
         // just relocate what exists into new block
         lfs_block_t nblock;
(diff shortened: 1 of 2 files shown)


lfs.c
```diff
@@ -2410,12 +2414,19 @@ int lfs_file_close(lfs_t *lfs, lfs_file_t *file) {
     // clean up memory
     if (!file->cfg->buffer) {
         lfs_free(file->cache.buffer);
+        file->cache.buffer = NULL;
```
geky: Is this needed with the file protection logic? Counter-intuitively it would actually hide use-after-free errors when tested inside valgrind.

geky:
Hi @Ar2rL, thanks for creating a PR! Sorry it's taken me so long to review this.

I've rebased your branch because of a recent fix to get CI working again: https://github.com/ARMmbed/littlefs/commit/abd90cb84c818a663b584575b019258d01d0065e, let me know if that causes any issues. Also it's curious your commit email points to a different GitHub user.

I'm happy to take in the file protection once we know the code impact. Though some thoughts:

1. Thoughts on inverting the logic to use a LFS_F_OPENED flag? This may better protect against using files that haven't been opened yet, any files declared in the .bss section for instance.

2. Thoughts on using asserts instead of LFS_ERR_BADF? This would let users reduce code size further if they want the minimum. (I realize you were copying the RDWR protection logic, thanks for that. I think we should actually change both to assert).

lfs.c
Ar2rL: Indded protecttion logic is enough and this is not needed.
It is just result of my habit to assign NULL after deallocation.

Ar2rL:
Hi @geky 
1. Adding "inverted logic" was my first follow up after I send this pull request.
I lost a few hours caused by my error of using closed file, so I was focused to protect other "innocent" users ;-) but after some thinking protection against not opened file is also good thing.
2. Good idea.
Anyway, I use NASA principle "Test what you fly, fly what you test", so my systems have always assertions enabled but I understand that for other users size increase can be very problematic.
BTW> Regarding different GitHub user - it is my first GitHub pull request (I ususually use just git and sometimes Bitbucket), so I messed things a little .

geky:
> Regarding different GitHub user - it is my first GitHub pull request (I ususually use just git and sometimes Bitbucket), so I messed things a little .

No worries, I was just curious :)

Closing as this is replaced by https://github.com/ARMmbed/littlefs/pull/237

PR #227, opened 2019-07-11
removed <dirent.h> preventing compile on some archs

haneefmubarak:
On some architectures `<dirent.h>` is not available (preventing the code from compiling); meanwhile, the code does not appear to use any of [the documented functionality of the same](https://pubs.opengroup.org/onlinepubs/7908799/xsh/dirent.h.html):

> ##### [ ]()DESCRIPTION
> 
> The internal format of directories is unspecified.
> 
> > The _&lt;dirent.h&gt;_ header defines the following data type through **typedef**:
> 
> DIR
> 
> A type representing a directory stream.
> 
> > It also defines the structure **dirent** which includes the following members:
> 
> ino_t  d_ino       file serial number
> char   d_name[]    name of entry
> 
> > The type **ino_t** is defined as described in _[&lt;sys/types.h&gt;](https://pubs.opengroup.org/onlinepubs/7908799/xsh/systypes.h.html)_.
> 
> > The character array **d_name** is of unspecified size, but the number of bytes preceding the terminating null byte will not exceed {NAME_MAX}.
> 
> > The following are declared as functions and may also be defined as macros. Function prototypes must be provided for use with an ISO C compiler.
> 
> int            [closedir](https://pubs.opengroup.org/onlinepubs/7908799/xsh/closedir.html)(DIR *);
> DIR           *[opendir](https://pubs.opengroup.org/onlinepubs/7908799/xsh/opendir.html)(const char *);
> struct dirent *[readdir](https://pubs.opengroup.org/onlinepubs/7908799/xsh/readdir.html)(DIR *);
> int            [readdir_r](https://pubs.opengroup.org/onlinepubs/7908799/xsh/readdir_r.html)(DIR *, struct dirent *, struct dirent **);
> void           [rewinddir](https://pubs.opengroup.org/onlinepubs/7908799/xsh/rewinddir.html)(DIR *);
> void           [seekdir](https://pubs.opengroup.org/onlinepubs/7908799/xsh/seekdir.html)(DIR *, long int);
> long int       [telldir](https://pubs.opengroup.org/onlinepubs/7908799/xsh/telldir.html)(DIR *);

The code appears to build fine without it, at least on `x86_64-apple-darwin18.6.0` and on `riscv32-unknown-elf`. It would be rather odd (and possibly depending on undocumented functionality?) if this removal thus prevented the code from compiling on any other architectures / in any other environments.

### DIFF
--- a/emubd/lfs_emubd.c
+++ b/emubd/lfs_emubd.c
@@ -11,7 +11,7 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <limits.h>
-#include <dirent.h>
+//#include <dirent.h>
 #include <sys/stat.h>
 #include <unistd.h>
 #include <assert.h>
--- a/emubd/lfs_emubd.c
+++ b/emubd/lfs_emubd.c
@@ -11,7 +11,6 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <limits.h>
-//#include <dirent.h>
 #include <sys/stat.h>
 #include <unistd.h>
 #include <assert.h>


geky:
Looks like you're right, thanks for creating this PR. At some point emubd.c was probably using dirent.h, but I guess that's no longer the case.

Interesting to hear this was prevent compilation on other platforms. From your output does this mean it's possible to run the tests on Mac?

Interestingly, this may have been the reason for needing this fix https://github.com/ARMmbed/littlefs/commit/abd90cb84c818a663b584575b019258d01d0065e

emubd/lfs_emubd.c
```diff
@@ -11,7 +11,7 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <limits.h>
-#include <dirent.h>
+//#include <dirent.h>
```
geky: Are you able to remove this completely? I don't think we need to keep it around.

haneefmubarak: Ah my bad, I pushed a commit that fixes that.

haneefmubarak:
> Interesting to hear this was prevent compilation on other platforms. From your output does this mean it's possible to run the tests on Mac?

So `<dirent.h>` is a standardized POSIX header, so MacOS has it. I'm working with the `riscv32-unknown-elf` toolchain for `gcc` though, which does not support the same. I imagine when porting toolchains to many new platforms, headers that are more needed for standards-compliance and sophisticated programs (rather than basic low-level programs) tend to be towards the end of the to-do list.

> From your output does this mean it's possible to run the tests on Mac?

It should be possible; if it can be run on Linux using only POSIX APIs, I don't see why you wouldn't be able to run them on Mac. That being said, I tried running them and almost all of them passed - test_corrupt.sh fails with `printf ()` being angry about an invalid number (line 112). I'll look into it later and probably send in a PR to get the tests to pass on Mac, because I think the only issues that prevent the tests from completing are `bash`-isms in the shell script that might be _version-challenged_.

haneefmubarak:
@geky is there anything else that you think needs changes prior to merging?

geky:
Nope! It looks good. Thanks for this!

Just waiting to coalesce a few PRs into a patch release.

> I think the only issues that prevent the tests from completing are bash-isms in the shell script that might be version-challenged.

I was just curious, I don't have access to a Mac so I've never tried. Though the fact I'm running bash commands directly does mean there might be problems...

haneefmubarak:
> Nope! It looks good. Thanks for this!
> 
> Just waiting to coalesce a few PRs into a patch release.

Gotcha. Sounds good!

> > I think the only issues that prevent the tests from completing are bash-isms in the shell script that might be version-challenged.
> 
> I was just curious, I don't have access to a Mac so I've never tried. Though the fact I'm running bash commands directly does mean there might be problems...

I'd be interested in helping to get this figured out a few weeks out from now if you think it's worthwhile (my work backlog is a little full at the moment); I use a Mac as my main machine for development at work.

PR #231, opened 2019-07-17
ignore LFS_NO_INTRINSICS  in noop endianess change

ujeropoc:
The naive way of returning `a` if `a` is already correct should not be gated with LFS_NO_INTRINSICS.

### DIFF
--- a/lfs_util.h
+++ b/lfs_util.h
@@ -142,10 +142,9 @@ static inline int lfs_scmp(uint32_t a, uint32_t b) {
 
 // Convert between 32-bit little-endian and native order
 static inline uint32_t lfs_fromle32(uint32_t a) {
-#if !defined(LFS_NO_INTRINSICS) && ( \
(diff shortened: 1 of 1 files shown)


geky:
Hi @ujeropoc, thanks for creating a pr!

Out of curiosity, are you currently using LFS_NO_INTRINSICS and is this change useful?

The main reason LFS_NO_INTRINSICS exists is to quickly debug compiler issues. One of these issues could be in the emission of the BYTE_ORDER macros themselves, which could cause the compiler to emit the wrong conversion function. The BYTE_ORDER macros are annoyingly non-standard already, which is why there are so many variations.

The LFS_NO_INTRINSICS path is written to be correct C for either byte order.

ujeropoc:
I made the same edit to my LWS_CONFIG file and it seems to work. But if the BYTE_ORDER macros are not reliable, it makes sense to leave it as is.

PR #233, opened 2019-07-17
Change block_cycles disable from 0 to -1

geky:
As it is now, block_cycles = 0 disables wear leveling. This was a mistake. 0 is the "default" value for several other config options and makes it easy to accidentally disable wear-leveling. It's even worse when migrating from v1 as it's easy to miss the addition of block_cycles and end up with a filesystem that is not actually wear-leveling.

Clearly, block_cycles = 0 should do anything but disable wear-leveling.

Here, I've changed block_cycles = 0 to assert. Forcing users to set a value for block_cycles (500 is suggested). block_cycles can be set to -1 to explicitly disable wear leveling if desired.

Related: https://github.com/ARMmbed/littlefs/issues/189

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -1453,7 +1453,7 @@ static int lfs_dir_compact(lfs_t *lfs,
 
     // increment revision count
     dir->rev += 1;
-    if (lfs->cfg->block_cycles &&
+    if (lfs->cfg->block_cycles > 0 &&
             (dir->rev % (lfs->cfg->block_cycles+1) == 0)) {
         if (lfs_pair_cmp(dir->pair, (const lfs_block_t[2]){0, 1}) == 0) {
             // oh no! we're writing too much to the superblock,
@@ -3192,8 +3192,10 @@ static int lfs_init(lfs_t *lfs, const struct lfs_config *cfg) {
     LFS_ASSERT(4*lfs_npw2(0xffffffff / (lfs->cfg->block_size-2*4))
             <= lfs->cfg->block_size);
 
-    // we don't support some corner cases
-    LFS_ASSERT(lfs->cfg->block_cycles < 0xffffffff);
+    // block_cycles = 0 is no longer supported, must either set a number
+    // of erase cycles before moving logs to another block (~500 suggested),
+    // or explicitly disable wear-leveling with -1.
+    LFS_ASSERT(lfs->cfg->block_cycles != 0);
 
     // setup read cache
     if (lfs->cfg->read_buffer) {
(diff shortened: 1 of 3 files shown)


lfs.c
```diff
@@ -3192,8 +3192,10 @@ static int lfs_init(lfs_t *lfs, const struct lfs_config *cfg) {
     LFS_ASSERT(4*lfs_npw2(0xffffffff / (lfs->cfg->block_size-2*4))
             <= lfs->cfg->block_size);
 
-    // we don't support some corner cases
-    LFS_ASSERT(lfs->cfg->block_cycles < 0xffffffff);
+    // block_cycles = 0 is no longer supported, must either set a number
+    // of erase cycles before moving logs to another block (~500 suggested),
+    // or explicitly disable wear-leveling with -1.
+    LFS_ASSERT(lfs->cfg->block_cycles != 0);
```
geky: To be clear this may cause problems on drivers that currently disable wear-leveling.

Alternatively, I could make this default to a suggested value (500). This would maintain backwards compatibility in the common case, but may surprise users if they are intentionally disabling wear-leveling for whatever reason. It would also obscure the fact that wear-leveling was disabled, which is not my intention.

Not sure what the best answer is.

haneefmubarak: I think that asserting upon zero is the correct behavior. I think this is perfect:

 > Here, I've changed block_cycles = 0 to assert. Forcing users to set a value for block_cycles (500 is suggested). block_cycles can be set to -1 to explicitly disable wear leveling if desired.

I think the wording could be cleared up a little to explain the wear-leveling a bit more - I read it as continuously rewriting a particular block up to `block_cycles` times before moving to another block - does that sound correct to you?

pabigot: Also it would help if the suggested value was added to the documentation block in `lfs.h`.  I've been confused about what to use for that parameter.

geky: Does this make more sense? (I've added the term "metadata log" and changed the wording a bit)
``` c
// Number of erase cycles before littlefs evicts metadata logs and moves 
// the metadata to another block. Suggested values are in the range 100-1000,
// with large values having better performance at the cost of less consistent
// wear distribution.
//
// Set to -1 to disable block-level wear-leveling.
int32_t block_cycles;
```

> I read it as continuously rewriting a particular block up to block_cycles times before moving to another block

That is correct. However this only applies to metadata logs, which are also doing their own form of static wear-leveling on the block.

It can end up with block_cycle-sized spikes of wear on the block device, but in the long run these get smoothed out as the spikes get distributed across the block device.

PR #237, opened 2019-07-21
Protect (LFS_ASSERT) file operations against using not opened or closed files.

Ar2rL:
Do not allow some file operations to (re)use lfs_file_t stucture when file is already opened (prohibit open) or not opened/closed (prohibit all operations but open).
It is achieved by adding new internal file flag. This flag is set in lfs_file_open and it is reset in lfs_file_close. It is also tested in each (file related) API call as an ASSERTprecondition.
This is second attempt to add this functionality after discussion in pull request [#213](https://github.com/ARMmbed/littlefs/pull/213).

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -2262,7 +2262,7 @@ int lfs_file_opencfg(lfs_t *lfs, lfs_file_t *file,
     // setup simple file details
     int err;
     file->cfg = cfg;
-    file->flags = flags;
+    file->flags = flags | LFS_F_OPENED;
     file->pos = 0;
     file->cache.buffer = NULL;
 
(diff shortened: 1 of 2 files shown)


geky:
Oh hey, thanks for this! This looks perfect, I will merge it next chance I get 👍

Just to let you know, if you force-push to the same branch, GitHub will automatically update any PRs using that branch.

geky:
Just an FYI I also create a PR to change the rdonly/wronly asserts, these should be able to go in together. I created a separate PR so others would be able to see that this is changing if they are watching the repo.
https://github.com/ARMmbed/littlefs/pull/243

geky:
Just an FYI, I had to remove the assert in `lfs_file_opencfg` as it was causing CI to fail:
https://github.com/ARMmbed/littlefs/commit/e249854858dd28f263e96d96ac3e570b2d591a4a

We unfortunately can't assume the initial value of the lfs_file_t struct in lfs_file_opencfg. But at least this is still useful for protecting against operations on closed files.

PR #243, opened 2019-07-24
Change rdonly/wronly mistakes to assert

geky:
Previously these returned LFS_ERR_BADF. But attempting to modify a file
opened read-only, or reading a write-only flie, is a user error and
should not occur in normal use.

Changing this to an assert allows the logic to be omitted if the user
disables asserts to reduce the code footprint (not suggested unless the
user really really knows what they're doing).

I'm making this change because I can't think of any use cases where an error code would be helpful. But let me know if I'm missing a use case.


geky:
Related https://github.com/ARMmbed/littlefs/pull/237

Issue #75, opened 2018-07-14
Question about scalability and large files


FreddieChopin:
First of all, this is not an "issue", just a question, as I'm not sure whether what I see is normal/expected or not.

I've ported littlefs to my C++ RTOS - https://github.com/DISTORTEC/distortos and I'm trying it out with an SD XC 32 GB card on STM32F429. My test more-or less goes like this:
- dump flash memory directly to SD card (raw write speed) - 10 times,
- format the card with littlefs,
- write the same dump of flash memory to a littlefs file (file system write speed), which always gets truncated when opened - 10 times.

These 3 steps are repeated for write sizes from following sequence: 512 B, 1 kB, 2 kB, ... 512 kB, 1 MB, 2 MB.

After some runs I noticed that difference between speed of raw sequential write to the SD card (with no file system) vs. speed of write with the file system is quite huge, but only for "large" files. The difference grows when the size of file grows - for example when writing just 1 kB, the ratio is just ~2.7, for 128 kB this is already ~7.2, while for 2 MB it's ~14.6 (these are all for average values).

Below is a chart which shows my measurements:

![image](https://user-images.githubusercontent.com/1139038/42723328-0c80334a-875c-11e8-91c0-90a8d1a5eac5.png)
(x axis - write size, y axis - ratio between average fs write speed vs average raw write speed)

Here's another interesting observation - in my test I do 10 runs of raw write, format the card with littlefs and do 10 runs of file system write. While the speed of raw writes are  mostly identical, the speed of file system writes vary:
- for small files - below and equal to 16 kB - the first write is actually the slowest one, while next 9 are a bit faster (for example 384 ms first write vs 320 ms second write),
- for medium-sized files - 32 kB to 128 kB - the speed of file system writes is mostly consistent,
- for large files - 256 kB to 2 MB - the first write is the fastest, following writes are slower, for 2 MB file the last 3 writes get even slower than earlier 6.

Each write goes to the same file, which I truncate to zero on creation.

Here's the chart of these findings:

![image](https://user-images.githubusercontent.com/1139038/42723555-391f5bac-8760-11e8-9d16-22fec615f158.png)
(x axis - number of write, y axis - ratio write speed of this write vs the speed of first write)

I guess it's also worth noting that I use 512 bytes for read, program and erase block size, and 512 for "lookahead" value.

To summarise:
1. Is this expected, that for large files the performance is more than 10x lower than raw performance of the SD card, while for small files the difference is much smaller?
[...]

PR #253, opened 2019-07-31
lfs: correct alignment restriction on lookahead buffer

pabigot:
The buffer need only be 32-bit aligned.

Signed-off-by: Peter A. Bigot <pab@pabigot.com>

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -3412,10 +3412,10 @@ static int lfs_init(lfs_t *lfs, const struct lfs_config *cfg) {
     lfs_cache_zero(lfs, &lfs->rcache);
     lfs_cache_zero(lfs, &lfs->pcache);
 
-    // setup lookahead, must be multiple of 64-bits
+    // setup lookahead, must be multiple of 64-bits, 32-bit aligned
     LFS_ASSERT(lfs->cfg->lookahead_size > 0);
     LFS_ASSERT(lfs->cfg->lookahead_size % 8 == 0 &&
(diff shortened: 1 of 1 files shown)


geky:
Hi @pabigot, thanks for creating a pr!

This was discussed over here: https://github.com/ARMmbed/littlefs/issues/134, though I realize it's not easy to find and should be documented in lfs.h.

> One of the possible solutions for #75 involves storing pairs of pointers (address+size pairs). The lookahead buffer would then need to be a multiple of 64-bits.
>
> Well, it doesn't _need_ to be, but any unaligned size would be wasted. It's cheaper and more transparent to the user to present this as a strict requirement to the user.
>
> If we don't adopt this solution for #75, we can just remove the 64-bit requirement. But adopting it early makes it possible to introduce without breaking compatibility.

It's not needed yet but the alignment is reserved as it is likely it will be needed to solve #75. If it isn't we can remove the assert.

pabigot:
@geky I'm confused again.  I thought with #239 you agreed: yes, the *length* of the lookahead array needs to be a multiple of 8 bytes, but even with the proposed solution of pairs of address and size the *alignment* of the array is to 4-byte values.

So it *is* documented in lfs.h.  I hadn't realized the code enforced the too-strong alignment until I finally ran into the situation.

geky:
Ah! You're right, I'm getting confused again.

After some more coffee this looks great 👍

PR #257, opened 2019-08-03
fix seek position corruption in truncate function

### LINKED ISSUE #252: misbehavior of `lfs_file_truncate`
It appears that lfs_file_truncate() moves the file head without recording that fact.

To reproduce: In a file system with 16 byte block size, and 64 byte cache, create a file that has 64 bytes increasing from 0x00 to 0x3F.  Close the file.

Then open the file and execute these commands:
```
uint8_t buffer[64];
printf("tell %u\n", (unsigned int)lfs_file_tell(lp, fp));
printf("read %d\n", lfs_file_read(lp, fp, buffer, sizeof(buffer)));
printf("\t%02x %02x %02x %02x\n", buffer[0], buffer[16], buffer[32], buffer[48]);
printf("seek %d\n", (int)lfs_file_seek(lp, fp, 0, LFS_SEEK_SET));
printf("tell %u\n", (unsigned int)lfs_file_tell(lp, fp));
printf("truncate %u\n", (unsigned int)lfs_file_truncate(lp, fp, 48));
printf("tell %u\n", (unsigned int)lfs_file_tell(lp, fp));
memset(buffer, -1, sizeof(buffer));
printf("read %d\n", lfs_file_read(lp, fp, buffer, sizeof(buffer)));
printf("\t%02x %02x %02x %02x\n", buffer[0], buffer[15], buffer[16], buffer[47]);
printf("tell %u\n", (unsigned int)lfs_file_tell(lp, fp));
```

The annotated output I get is:
```
tell 0      # file is correctly positioned at start after open
read 64     # file reads 64 bytes that appear to be numbered correctly
        00 10 20 30
seek 0      # Move back to start of file works
tell 0      
truncate 0  # Truncate to 48 bytes succeeds
tell 0      # File position is said to be zero
read 48     # Read returns the length of the file...
        30 3f 70 ff  # .. but the data is the 16 bytes after truncation,
                     # plus garbage past the original end of file
tell 48     #
```

Two anomalous behaviors:
* I did not expect lfs_file_truncate() to move the file position.  If that is expected behavior, `lfs_file_tell()` should have reported it.
* I did not expect to receive data that started at the truncated part of the file, and included garbage beyond the original file end.

EDIT: In this context read and prog size are 16, cache size is 64.

pabigot:
[ftruncate](https://pubs.opengroup.org/onlinepubs/009695399/functions/ftruncate.html) is not supposed to move the seek pointer, but when making a file smaller the littlefs code did this.

The third patch in this set closes #252.

The first two are cleanup patches that (IMO) make it easier for people to understand the code and work with the code: one changing magic numbers to meaningful symbols, and one ensuring the file `off` field is not left uninitialized during open.  I can remove either or both of those if they're not wanted.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -2329,6 +2329,7 @@ int lfs_file_opencfg(lfs_t *lfs, lfs_file_t *file,
     file->cfg = cfg;
     file->flags = flags | LFS_F_OPENED;
     file->pos = 0;
+    file->off = 0;
     file->cache.buffer = NULL;
 
     // allocate entry for file if it doesn't exist
--- a/lfs.c
+++ b/lfs.c
@@ -2981,6 +2981,7 @@ int lfs_file_truncate(lfs_t *lfs, lfs_file_t *file, lfs_off_t size) {
         return LFS_ERR_INVAL;
     }
 
+    lfs_off_t pos = file->pos;
     lfs_off_t oldsize = lfs_file_size(lfs, file);
     if (size < oldsize) {
         // need to flush since directly changing metadata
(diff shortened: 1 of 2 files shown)


geky:
This looks great, I will merge this next chance I get. And thanks for the test case.

I think the move away from magic values is a good idea, I'll have to go update the magic ids/lengths at some point. Happy to take both that and the struct initializer in.

Thanks for this!

PR #259, opened 2019-08-05
Fix if vs ifdef condition on LFS_MIGRATE

geky:
Found by @e107steved: https://github.com/ARMmbed/littlefs/issues/247
> Version 2.1.0 release, line 1503 of lfs.c has "#if LFS_MIGRATE". All other instances are "#ifdef LFS_MIGRATE"


geky:
Superseeded by https://github.com/ARMmbed/littlefs/pull/263

PR #261, opened 2019-08-07
Changed CI to create release note for patches

geky:
This is a result of feedback that the current release notes made it too difficult to see what changes happened on patch releases. From my experience as well it became difficult to chase down which release a commit landed on.

The risk is that this creates additional noise, both for the release page and for user notifications. I am open to feedback if this causes a problem.

Other tweaks on the CI side, these came from iteration with the same scheme for coru and equeue:

- Changed version branch updates to be atomic (vN and vN-prefix). This makes it a bit easier to fix if one of the pushes fails due to a rogue branch with the same name.

- Added GEKY_BOT_DRAFT as a CI macro that can optionally switch between only creating drafts or immediately posting a release. The default is what I will be trying with littlefs which is to draft minor/major releases, but automatically create patch release.

  The real benefit of automatic releases is to use on tiny repos that don't really have an active maintainer. Though this is definitely no longer the case with littlefs, and I'm happy it has gained this much attention.

cc @FreddieChopin

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -248,34 +248,30 @@ jobs:
                   -m "Generated v$LFS_VERSION_MAJOR prefixes")
           git reset --hard
           # Update major version branches (vN and vN-prefix)
-          git push https://$GEKY_BOT_RELEASES@github.com/$TRAVIS_REPO_SLUG.git \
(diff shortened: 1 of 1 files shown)


geky:
This can be seen in action here: https://github.com/geky/littlefs-test-repo/releases

PR #262, opened 2019-08-07
A number of small QOL changes related to lfs_util.h

geky:
- Removed stdlib includes from lfs.h, these should all go through lfs_util.h to let users override these definitions if stdlib is unavailable on their system.

- Changed the name of the LFS_CONFIG macro to LFS_UTIL to avoid confusion with the lfs_config struct. This also hints that LFS_UTIL is related to lfs_util.h.

  LFS_UTIL allows the user to override lfs_util.h so they can provide their own system-level dependencies such as malloc, tracing, builtins, stdint definitions, string.h, and others.

- Moved error code definitions to lfs_util.h. This lets users override the error codes to replace them with their own error codes and avoid a translation layer in some situations. Note the error codes must still be in the range of a negative int.

Note that this is very likely to cause build issues for anyone currently using LFS_CONFIG.

To fix:
- Change LFS_CONFIG to LFS_UTIL
- Make sure you have a copy of the error code definitions in your customized lfs_util.h

Because of these changes this PR will need to wait for at least a minor release.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -5,7 +5,6 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 #include "lfs.h"
-#include "lfs_util.h"
 
 
 /// Caching block device operations ///
--- a/lfs.h
+++ b/lfs.h
@@ -7,8 +7,7 @@
 #ifndef LFS_H
 #define LFS_H
 
-#include <stdint.h>
-#include <stdbool.h>
+#include "lfs_util.h"
 
 #ifdef __cplusplus
 extern "C"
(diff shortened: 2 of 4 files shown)


mr-at-eo:
There is an `#ifndef LFS_CONFIG` in `lfs_util.c` which this PR does not address.

This is not ideal anyway, if I want to use default `lfs_utl.c` with custom `lfs_util.h`. Currently there has to be an `#undef LFS_CONFIG` at the end of `lfs_util_custom.h`. This is kind of ugly and I would prefer something like this:

in lfs_util.c:

    #ifndef LFS_CUSTOM_CRC
     implement lfs_crc
    #endif

That way I can define `LFS_CUSTOM_CRC`  in `lfs_util_custom.h` if I have a better implementation, or use the default one if I don't.

geky:
Good catch :+1:, sorry this has been pending for so long.

Hmm, it's interesting. I hadn't thought of using the default lfs_util.c with a custom lfs_util_custom.h, but it makes sense that this could come up if you wanted to tweak one part of lfs_util.h. You can copy lfs_util.c -> lfs_util_custom.c, but I can see how that may be a bit annoying.

Would it make sense to split out lfs_crc into its own lfs_crc.h and lfs_crc.c files?

lfs_util.h could then include and expose lfs_crc unless you don't include the file in lfs_util_custom.h

geky:
Closing in favor of https://github.com/littlefs-project/littlefs/pull/491

I did consider `LFS_CUSTOM_CRC` again, but decided to leave it out as worst case it isn't too difficult to copy over lfs_util.c for the definition. But if this keeps causing problems for users, maybe we should add `LFS_CUSTOM_CRC` in the future.

PR #263, opened 2019-08-07
Fix build with -Wundef

rojer:
Part of https://github.com/mongoose-os-libs/vfs-fs-lfs/issues/2

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ override CFLAGS += -DLFS_YES_TRACE
 endif
 override CFLAGS += -I.
 override CFLAGS += -std=c99 -Wall -pedantic
-override CFLAGS += -Wextra -Wshadow -Wjump-misses-init
+override CFLAGS += -Wextra -Wshadow -Wjump-misses-init -Wundef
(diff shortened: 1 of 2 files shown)


geky:
Hmm, is there a reason for not putting Wundef in the normal build options?
https://github.com/ARMmbed/littlefs/blob/master/Makefile#L32

Thanks for creating a PR.

geky:
Hi @rojer, thanks for creating a PR. This is a good idea.

Hmm, is there a reason for not putting Wundef in the normal build options?
https://github.com/ARMmbed/littlefs/blob/master/Makefile#L32

rojer:
well, the idea is to prevent this from happening in the future.

geky:
The Makefile is used directly in CI: https://github.com/ARMmbed/littlefs/blob/master/.travis.yml#L18

If we move -Wundef to the Makefile, that just means users will have the warning if they run the tests locally.

The only reason -Werror is special is because halting on warnings can be a pain when testing/debugging locally.

rojer:
ok, sorry, i misunderstood. moved to Makefile.

PR #266, opened 2019-08-09
Revert "Don't bypass cache in `lfs_cache_prog()` and `lfs_cache_read()`"

FreddieChopin:
This reverts commit fdd239fe21a3721dc7fb55a9cf816544af7e9647.

Bypassing cache turned out to be a mistake which causes more problems
than it solves. Device driver should deal with alignment if this is
required - trying to do that in a file system is not a viable solution
anyway.

#160


geky:
@FreddieChopin, sorry about keeping this open for so long, I'll be merging this into v2.2.

PR #278, opened 2019-08-30
lfs: Validate lfs-cfg sizes before performing any arithmetic logics with them

roykuper13:
Since `lfs_init` validates lfs config attributes, it should also check that the attributes are initiated properly, in order to not crash, for example, if `read_size` or `prog_size` or `cache_size` are not set (floating point excpetion).

Also, those attributes are really critical for lfs, and if the user didn't set those values, he must be aware of that (assertion error is pretty indicative, unlike a weird floating point exception..).

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -3369,6 +3369,12 @@ static int lfs_init(lfs_t *lfs, const struct lfs_config *cfg) {
     lfs->cfg = cfg;
     int err = 0;
 
+    // validate that the lfs-cfg sizes were initiated properly before
+    // performing any arithmetic logics with them
+    LFS_ASSERT(lfs->cfg->read_size != 0);
+    LFS_ASSERT(lfs->cfg->prog_size != 0);
+    LFS_ASSERT(lfs->cfg->cache_size != 0);
+
     // check that block size is a multiple of cache size is a multiple
     // of prog and read sizes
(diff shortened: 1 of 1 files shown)


geky:
Looks good to me, should we also assert that read/prog/erase/sync are non-null?

roykuper13:
> Looks good to me, should we also assert that read/prog/erase/sync are non-null?

we should, but maybe not in `lfs_init`, because if, for example, someone wants to implement a read-only filesystem using littlefs, a prog() function is irrelevant (I hope this example makes any sense, correct me if i'm wrong). therefore we can't assert that prog is non-null in `lfs_init`.

I suggest to assert each function where they are being used. for example, `lfs_bd_flush` should first check that `lfs->cfg->prog != NULL`.

tell me what you think.

geky:
I've gone ahead and merged the original PR, thanks for pushing it up. Feel free to create another PR if you think we should add more checks.

The main issue with asserting where used is that I think that is basically every entry-point to the filesystem, which is quite a number of functions. Including mount. So we might as well assert during mount/format and assume the user won't change the config struct while the filesystem is mount (the user could also change geometry such as `block_size`, which would likely break the filesystem).

We don't want too many asserts as then we risk making asserts to expensive to enable on smaller MCUs.

That's a good point about a read-only filesystem though. Given the way the filesystem API works with files and flags, we already would need a build-time define to configure read-only mode, so we could use that to decide if we assert on write functions.

PR #282, opened 2019-09-02
Corrections for typos and grammar

runderwo:
While reading through the doc I kept breaking out of analysis mode when I ran across phrasing that didn't make sense.  These are just some proposals and you can do what you want with them :-)


geky:
Thanks for this! These all look great.

Reading through, some of the typos are quite funny. I can see why they would be distracting. :)

PR #286, opened 2019-09-09
build: Fix warnings about shift count width difference for 16 bit com…

sipke:
Build warnings exist on a gcc based 16 bit compiler. Cast relevant types
to fix.

littlefs/lfs.c: In function 'lfs_gstate_xororphans':
littlefs/lfs.c:355:5: warning: left shift count >= width of type
littlefs/lfs.c: In function 'lfs_dir_fetchmatch':
littlefs/lfs.c:849:17: warning: left shift count >= width of type
littlefs/lfs.c: In function 'lfs_dir_commitcrc':
littlefs/lfs.c:1278:9: warning: left shift count >= width of type

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -352,7 +352,7 @@ static inline bool lfs_gstate_hasmovehere(const struct lfs_gstate *a,
 
 static inline void lfs_gstate_xororphans(struct lfs_gstate *a,
         const struct lfs_gstate *b, bool orphans) {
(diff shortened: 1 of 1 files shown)


geky:
This looks great! Thanks for creating a PR.

We really should add a test job for a 16-bit MCU. Do you know any 16-bit MCUs that have qemu support?

But there's no reason to not merge this. I will merge this next chance I get.

PR #298, opened 2019-09-30
fix types for 16bit machines

sipke:
In implementing more of the littlefs functionality on a 16 bit machine, found errors in directory traversal, which were a result of data type conversion issues between 16 bit int and the 32 bit size and tag types in littlefs.
Also found that the issue https://github.com/ARMmbed/littlefs/issues/178 was resolved through correcting the type conversions, though can't say which of the changes exactly as I blanket replaced all that I could find.
In hindsight it may have been more straight forward to use a typedef for return type (e.g. typedef int err_ret_t), rather than trying to address each case one by one.
Either way for 16 bit machines, these changes or similar will be needed, so let me know your thoughts as we would like these changes upstream.
I did not address the values in the LFS_MIGRATE section, though technically they should be done too.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -1372,12 +1372,12 @@ static lfs_stag_t lfs_dir_drop(lfs_t *lfs, lfs_mdir_t *dir, lfs_mdir_t *tail) {
     return 0;
 }
 
-static int lfs_dir_split(lfs_t *lfs,
+static lfs_stag_t lfs_dir_split(lfs_t *lfs,
         lfs_mdir_t *dir, const struct lfs_mattr *attrs, int attrcount,
         lfs_mdir_t *source, uint16_t split, uint16_t end) {
     // create tail directory
     lfs_mdir_t tail;
-    int err = lfs_dir_alloc(lfs, &tail);
+    lfs_stag_t err = lfs_dir_alloc(lfs, &tail);
     if (err) {
         return err;
     }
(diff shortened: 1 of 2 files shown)


geky:
Hi @sipke, thanks for creating a PR!

I think it's safe to omit LFS_MIGRATE for now, any progress towards better 16-bit support is good progress.

One of the first things is that the top-level public API (anything in lfs.h) needs to stay as an `int` return type. I realize it may not be clear in the code, but any tag types getting returned out of the public API should be stripped down to the POSIXish error codes. These are expected to fit in `int` on any system since that is what is expected by `errno.h`. I realize I may eat my words later, but I think we should keep it to `int` for now.

The `lfs_stag_t` is an entirely internal type, so if it gets exposed through the public API then something is wrong. I believe it's more correct to cast these to `int` once we've tested that they only contain errors (< 0 for `lfs_stag_t` types)

``` diff
    lfs_stag_t tag = lfs_dir_find(lfs, &cwd, &path, NULL);
    if (tag < 0) {
        LFS_TRACE("lfs_stat -> %d", tag);
-        return tag;
+        return (int)tag;
    }
```

sipke:
That works. Will probably take a bit of effort to work out each case, which is why I was hoping for the blanket change, but fully understand the requirement.
Will someone else with more internal knowledge look at this?
If not I will probably look at individual cases and send smaller PRs as I find/verify fixes.

sipke:
Did a v2 of this pull request based on feedback: https://github.com/ARMmbed/littlefs/pull/299

sipke:
Closing as new pull request v2 https://github.com/ARMmbed/littlefs/pull/299 created.

PR #299, opened 2019-10-01
fix types for 16bit machines v2

sipke:
Version two for original pull request https://github.com/ARMmbed/littlefs/pull/298, modified based on feedback.
Commit https://github.com/ARMmbed/littlefs/commit/0b5a78e2cd5c0fc927a6828e01a0aee08d1f5e0d seems to resolve the issue https://github.com/ARMmbed/littlefs/issues/178.


geky:
Hi @sipke, thanks for this!

After reviewing the changes it looks good to me. Thanks for updating.

Sorry it has been pending for so long, I was hoping to get some fork of qemu working to add 16-bit support to CI, but I think that will have to wait. It shouldn't block this PR at least.

Issue #156, opened 2019-03-14
Making littlefs thread-safe?


rbphilip:
Hey! I just found littlefs and am porting it this morning to a FreeRTOS system with 128MB NOR flash. 

I see that littlefs doesn't have any sort of thread-safe mechanism built in. That's fine and I've seen references to wrapping littlefs calls with semaphores. Also fine. 

The question is *which* calls need to be wrapped. open/close/read/write/??

Anything that has a prototype in lfs.h? That'll likely be my first stab at it, but perhaps this is an issue that someone else has resolved in an elegant manner..

Thanks for any insights.

FreddieChopin:
> The question is which calls need to be wrapped. open/close/read/write/??

Everything which is an user-facing public API.

For an example see this port:
https://github.com/DISTORTEC/distortos/tree/master/source/FileSystem/littlefs
https://github.com/DISTORTEC/distortos/tree/master/include/distortos/FileSystem/littlefs

rbphilip:
Unfortunately I have zero experience with C++. Is "const std::lock_guard<LittlefsFile> lockGuard {*this};" a semaphore "take"? 

Is this "return" the semaphore "give" "return {{}, ret};"

FreddieChopin:
> Is "const std::lock_guard lockGuard {*this};" a semaphore "take"?

Yes.

> Is this "return" the semaphore "give" "return {{}, ret};"

Not exactly. Any possible exit from this function will destroy this `lockGuard` object, which will cause its destructor to be executed, which in turn releases the mutex. Generally any exit from this function causes automatic release of the mutex.

Please also don't confuse mutex and semaphore - the fact that in FreeRTOS they are "almost the same thing" (one of many stupid things in FreeRTOS...) doesn't mean that they are meant for the same purpose.

rbphilip:
Thanks. 

It appears that there is a different mutex used for directories than for files. Do you perhaps know if this is true (again, I don't really grok C++) and if so, why?

FreddieChopin:
There are no separate mutexes for file and directory objects - there is only one recursive mutex per file system and locking the file or directory is achieved by locking the mutex in the associated file system.

`lockGuard` is only a RAII (look it up) object which locks/unlocks the mutex/object that you provide. The "thing" you provide just has to have `lock()` and `unlock()` functions which will be called in constructor and destructor respectively.

Here's the only mutex in these clases:
[...]

PR #317, opened 2019-11-02
Configurable thread-safe littlefs

roykuper13:
Configurable thread-safe littlefs.
Closes https://github.com/ARMmbed/littlefs/issues/156

Not sure if that's the best approach. Other options that I thought of raised some problems, for example:
1. Queuing littlefs callbacks and invoke them one by one from a different thread, but:
a. I don't want to make the user create a thread for littlefs due to a possible limitations in embedded devices.
b.  In order to queue a callback and invoke it later, I need to save the buffers (of `lfs_file_write`, `lfs_setattr` etc), which may be a problem.

2. use rwlock on `lfs`, but that won't really solve the problem.

So eventually I just decided to lock an entire littlefs function execution using a spinlock. To do so, I make sure that `lfs_malloc` and `lfs_free` won't be executed inside littlefs function (except `lfs_format` and `lfs_mount`). Also I assume that the read,prog,sync and erase callbacks that are given to `lfs` struct can be called inside a spinlock context.

This "feature" requires pointers to spinlock_lock and spinlock_unlock functions (As we don't want + can't add a "generic" spinlock implementation to the project), and exposes a nice `lfs_acquire_lock` and `lfs_release_lock` to wrap littlefs functions.

I test it locally simply by creating two threads, each one reads and writes to the same file thousands of times. By disabling this feature, the program aborted in one of the `lfs_file_opencfg` call from one of the threads (it happens everytime, not just once or twice). Enabling this feature and wrap the open, read write and close calls with the `lfs_acquire_lock` and `lfs_release_lock` solved the problem.
Do you have any suggestion how can I add a generic test for this project?

Btw, the diff between the sizes of my test binary with and without this feature is 152 bytes.

Anyway, If you have a better solution for a thread-safe littlefs I would be glad to hear.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,10 @@ endif
 ifdef TRACE
 override CFLAGS += -DLFS_YES_TRACE
 endif
+ifdef THREAD_SAFE
+override CFLAGS += -DLFS_THREAD_SAFE
+endif
+
 override CFLAGS += -I.
 override CFLAGS += -std=c99 -Wall -pedantic
 override CFLAGS += -Wextra -Wshadow -Wjump-misses-init -Wundef
--- a/lfs.c
+++ b/lfs.c
@@ -2330,7 +2330,6 @@ int lfs_file_opencfg(lfs_t *lfs, lfs_file_t *file,
     file->flags = flags | LFS_F_OPENED;
     file->pos = 0;
     file->off = 0;
-    file->cache.buffer = NULL;
 
     // allocate entry for file if it doesn't exist
     lfs_stag_t tag = lfs_dir_find(lfs, &file->m, &path, &file->id);
@@ -2338,6 +2337,22 @@ int lfs_file_opencfg(lfs_t *lfs, lfs_file_t *file,
         err = tag;
         goto cleanup;
     }
+
+#ifdef LFS_THREAD_SAFE
+    // We don't want to peform memory allocation inside a spinlock context.
+    // Therefore, the user should provide an initialized configuration struct
+    // with an allocated buffer.
+    LFS_ASSERT(NULL != file->cfg->buffer);
+    file->cache.buffer = file->cfg->buffer;
+#else
+    file->cache.buffer = NULL != file->cfg->buffer ?
+        file->cfg->buffer : lfs_malloc(lfs->cfg->cache_size);
+
+    if (NULL == file->cache.buffer) {
+        err = LFS_ERR_NOMEM;
+        goto cleanup;
+    }
+#endif // LFS_THREAD_SAFE
 
     // get id, add to list of mdirs to catch update changes
     file->type = LFS_TYPE_REG;
@@ -2413,17 +2428,6 @@ int lfs_file_opencfg(lfs_t *lfs, lfs_file_t *file,
         }
     }
 
-    // allocate buffer if needed
-    if (file->cfg->buffer) {
-        file->cache.buffer = file->cfg->buffer;
-    } else {
-        file->cache.buffer = lfs_malloc(lfs->cfg->cache_size);
-        if (!file->cache.buffer) {
-            err = LFS_ERR_NOMEM;
-            goto cleanup;
-        }
-    }
-
     // zero to avoid information leak
     lfs_cache_zero(lfs, &file->cache);
 
@@ -2485,10 +2489,12 @@ int lfs_file_close(lfs_t *lfs, lfs_file_t *file) {
         }
     }
 
+#ifndef LFS_THREAD_SAFE
     // clean up memory
     if (!file->cfg->buffer) {
         lfs_free(file->cache.buffer);
     }
+#endif
 
     file->flags &= ~LFS_F_OPENED;
     LFS_TRACE("lfs_file_close -> %d", err);
@@ -3083,6 +3089,20 @@ int lfs_stat(lfs_t *lfs, const char *path, struct lfs_info *info) {
     LFS_TRACE("lfs_stat -> %d", err);
     return err;
 }
+
+#ifdef LFS_THREAD_SAFE
+int lfs_acquire_lock(lfs_t *lfs) {
+    LFS_TRACE("lfs_acquire_lock(%p)", (void*)lfs);
+
+    return lfs->cfg->spinlock_lock(lfs->cfg->lock);
+}
+
+void lfs_release_lock(lfs_t *lfs) {
+    LFS_TRACE("lfs_release_lock(%p)", (void*)lfs);
+
+    lfs->cfg->spinlock_unlock(lfs->cfg->lock);
+}
+#endif
 
 int lfs_remove(lfs_t *lfs, const char *path) {
     LFS_TRACE("lfs_remove(%p, \"%s\")", (void*)lfs, path);
--- a/lfs.h
+++ b/lfs.h
@@ -174,6 +174,15 @@ struct lfs_config {
     // are propogated to the user.
     int (*sync)(const struct lfs_config *c);
 
+#ifdef LFS_THREAD_SAFE
+    void *lock;
+
+    // Continously tries to acquire the given lock
+    int (*spinlock_lock)(void *lock);
+    // Releases the given lock
+    int (*spinlock_unlock)(void *lock);
+#endif
+
     // Minimum size of a block read. All read operations will be a
     // multiple of this value.
     lfs_size_t read_size;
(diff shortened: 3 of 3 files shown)


geky:
Hi @roykuper13, thanks for creating a PR. Sorry about the late response.

littlefs isn't thread-safe, so it makes sense that you see aborts without external locking.

The question I have is could you not provide thread-safety by wrapping the littlefs functions with your own thread safe wrapper?
```
struct lfssafe {
    void *lock;
    lfs_t lfs;
};

int lfssafe_acquire_lock(struct lfssafe *lfssafe) {
    return spinlock_lock(lfssafe->lock);
}

int lfssafe_release_lock(struct lfssafe *lfssafe) {
    return spinlock_unlock(lfssafe->lock);
}

int mythread(struct lfssafe *lfssafe) {
    lfssafe_acquire_lock(lfssafe);
    int err = lfs_mkdir(&lfssafe->lfs, "hello");
    if (err) {
        lfssafe_release_lock(lfssafe);
        return err;
    }
    lfssafe_release_lock(lfssafe);
}
```

littlefs won't call malloc if buffers are provided, so if you provide all of the buffers you should be safe. You can insure this by defining LFS_NO_MALLOC during compilation to turn lfs_malloc into an error.

---

littlefs doesn't use multiple-threads, so I have a difficult time accepting that it needs hooks for a multi-threading API.

Implementing synchronization through a wrapper also gives you the freedom to move the synchronization up through any other shared logic you have on top of the filesystem.

roykuper13:
@geky 
This is actually what i did in my project.
But the problem that I see is that it might not be that clear what are the steps to make littlefs thread-safe. One can just think that wrap every call to lfs function is enough, But that's not true! if one wants to wrap lfs function correctly, he would have to dig into littlefs code and notice that it uses `malloc` and `free` in some of the cases, and he needs to understand how to "avoid" that logic (i.e - understand that he's the one who responsible for allocating lfs buffers). That makes littlefs interface not so friendly imo. The users who uses littlefs might not notice the `malloc` and `free` - and without noiticing, they'll perform memory operations inside lock contexts!

imo, that makes littlefs interface less friendly (i.e if the user must dig into the code of littlefs to validate that everything that it does can be done inside a lock context). 

 In this PR, we warns the user that he should allocated lfs buffers in order to make lfs thread-safe.
If you don't want that, I think that at least we should document what should be done in order to make lfs thread-safe (in README or SPEC files), though I believe that the suggested solution in this PR is the right thing to do. Also, this "feature" was (kind of) requested here: https://github.com/ARMmbed/littlefs/issues/156

it is not a PR that solves a problem or something, it just makes littlefs interface more friendly, clear and simple to use, when dealing with threads.

geky:
> But the problem that I see is that it might not be that clear what are the steps to make littlefs thread-safe.

> If you don't want that, I think that at least we should document what should be done in order to make lfs thread-safe (in README or SPEC files), though I believe that the suggested solution in this PR is the right thing to do. Also, this "feature" was (kind of) requested here: #156

This is fair, and I'd be happy to take in a PR that improves this. Though I am going to be a stickler that this should not increase code-size for non-multi-threaded code.

> One can just think that wrap every call to lfs function is enough, But that's not true! if one wants to wrap lfs function correctly, he would have to dig into littlefs code and notice that it uses malloc and free in some of the cases, and he needs to understand how to "avoid" that logic

Though, for most multi-threaded systems, malloc+free is safe to call inside locked code.

I believe it's only an issue for the spinlocks, though I'm not sure which system you're on.

> How's it contradict my suggestion? You have this freedom in my suggestion as well (eventually i'm exporting wrappers as well..)

It only makes a difference if you always have locks at the filesystem API. If you also have locks on, say, a relational database built on top of littlefs, then you now have duplicated locks. This duplication is not an issue for larger devices, but on small RAM constrained devices, the overhead of these locks can add up.

---

Here's my concerns:

1. Would it make more sense for lfs_mutex_lock/unlock to be static, instead of going through the lfs_config struct? It's possible for there to be multiple block devices in a system, but I don't think the same is true for mutexes. Static lfs_mutex_lock/unlock functions would be possible to optimize out in more cases.

   For example:
   ``` c 
   // in lfs_util.h
   int lfs_mutex_create(lfs_mutex_t *mutex) {
       (void*)mutex;
       return 0;
   }
   void lfs_mutex_destroy(lfs_mutex_t *mutex) {
       (void*)mutex;
   }
   void lfs_mutex_lock(lfs_mutex_t *mutex) {
       (void*)mutex;
   }
   void lfs_mutex_unlock(lfs_mutex_t *mutex) {
       (void*)mutex;
   }
   ```
   ``` diff
   // in lfs.c
   lfs_soff_t lfs_file_size(lfs_t *lfs, lfs_file_t *file) {
       LFS_TRACE("lfs_file_size(%p, %p)", (void*)lfs, (void*)file);
   +    lfs_mutex_lock(&lfs->mutex);
       LFS_ASSERT(file->flags & LFS_F_OPENED);
       int res;
       if (file->flags & LFS_F_WRITING) {
   -        LFS_TRACE("lfs_file_size -> %"PRId32,
   -                lfs_max(file->pos, file->ctz.size));
   -        return lfs_max(file->pos, file->ctz.size);
   +        res = lfs_max(file->pos, file->ctz.size);
       } else {
   -        LFS_TRACE("lfs_file_size -> %"PRId32, file->ctz.size);
   -        return file->ctz.size;
   +        res = file->ctz.size;
       }
   +    lfs_mutex_unlock(&lfs->mutex);
   +    LFS_TRACE("lfs_file_size -> %"PRId32, res);
   +    return res;
   }
   ```

   Note the code size isn't a one time cost, but risks adding overhead for every lfs function.

2. The benefit of adding a thread-safe API is that users would only need to provide lfs_mutex_lock/unlock, and in theory this means all lfs functions can be called from multiple threads.

   So I think this sort of addition would also need the hooks in every lfs function. I realize it's tedius, but not needing to call lfs_mutex_lock/unlock is the main benefit of integrating thread-safety into littlefs.

roykuper13:
> This is fair, and I'd be happy to take in a PR that improves this. Though I am going to be a stickler that this should not increase code-size for non-multi-threaded code.

That was my intention :) that's why I wrapped my changes with ifdefs, so it can be configurable. When disabled, the code size doesn't increase.


> Though, for most multi-threaded systems, malloc+free is safe to call inside locked code.
I believe it's only an issue for the spinlocks, though I'm not sure which system you're on.

Why do you think mutex should be use instead of spinlock? when taking the malloc and free out, littlefs functions are executed in reasonably short amount of time, that putting another thread to "sleep" is not worth it (I'm thinking about situations, for example, where you have 2 threads that loop through a big buffer and `lfs_file_write` it in chunks, so everytime one thread acquires a mutex lock, the second thread "sleeps"). I think spinlock should be used, correct me if i'm wrong.

About the concerns:

1. > Would it make more sense for lfs_mutex_lock/unlock to be static, instead of going through the lfs_config struct? It's possible for there to be multiple block devices in a system, but I don't think the same is true for mutexes

I guess the lock/unlock could be static, but the lock object itself should be part of the lfs_config struct.

2. > So I think this sort of addition would also need the hooks in every lfs function. I realize it's tedius, but not needing to call lfs_mutex_lock/unlock is the main benefit of integrating thread-safety into littlefs

 There's a problem using the lock/unlock functions inside lfs functions (the "hooks"). For example, If one thread acquires the lock in the begining of `lfs_file_truncate`, and a second thread tries to acquire the lock, the second will fail. But `lfs_file_truncate` calls `lfs_file_write` in one of its branches! So you will end up getting into `lfs_file_write` when your lock is acquired.. that's a problem (You also can't release the lock right before the `lfs_file_write` call, because another thread might acquire the lock before you enter `lfs_file_write`, and mess your lfs structs). Therefore I think that locks should be used before and after calling lfs functions.

Also, in binary size standpoint - I think that adding calls to lock/unlock functions in each lfs function increases the size more then  just wrap lfs functions (For example, taking the example code from README.md - you will just need to lock before `lfs_mount` and unlock after `lfs_unmount`).

roykuper13:
> This is fair, and I'd be happy to take in a PR that improves this. Though I am going to be a stickler that this should not increase code-size for non-multi-threaded code.

That was my intention :) that's why I wrapped my changes with ifdefs, so it can be configurable. When disabled, the code size doesn't increase.

Anyway, your suggestion can be made if the user provides a recursive lock/unlock implementations (simple locks will cause deadlocks). I'll update the PR and let you know when it's ready.

geky:
Side note: LWIP has similar sys_mutex call for synchronization (though multithreading is much more important there): https://www.nongnu.org/lwip/2_1_x/multithreading.html

It also expects malloc to be thread-safe.

Also if the mutex lock/unlock are static you may be able to introduce the hooks with zero code-size without ifdefs (though if not ifdefs work).

I think it's possible to restructure the littlefs functions to not require recursive mutexs, which would also be a bit nice for tracing. Though I don't know if it's worth it. It could always be added later without breaking compatibility.

beadon-SAT:
> > This is fair, and I'd be happy to take in a PR that improves this. Though I am going to be a stickler that this should not increase code-size for non-multi-threaded code.
> 
> That was my intention :) that's why I wrapped my changes with ifdefs, so it can be configurable. When disabled, the code size doesn't increase.
> 
> Anyway, your suggestion can be made if the user provides a recursive lock/unlock implementations (simple locks will cause deadlocks). I'll update the PR and let you know when it's ready.

@roykuper13 sounds good, we are running into a similar problem here.  Would love to see lfs thread safe !

geky:
Just an update, it looks like this has been picked up here: https://github.com/littlefs-project/littlefs/pull/470

PR #322, opened 2019-11-05
Allow debug prints without __VA_ARGS__ in non-MSVC

hemmick:
The terminating comma does not get removed in non-MSVC compilers.  Use ## to consume the comma in this case, allowing zero-arg prints

### DIFF
--- a/lfs_util.h
+++ b/lfs_util.h
@@ -50,31 +50,35 @@ extern "C"
 
 // Logging functions
 #ifdef LFS_YES_TRACE
-#define LFS_TRACE(fmt, ...) \
-    printf("lfs_trace:%d: " fmt "\n", __LINE__, __VA_ARGS__)
+#define LFS_TRACE_(fmt, ...) \
+    printf("lfs_trace:%d: " fmt "%s\n", __LINE__, __VA_ARGS__)
+#define LFS_TRACE(...) LFS_TRACE_(__VA_ARGS__, "")
 #else
-#define LFS_TRACE(fmt, ...)
+#define LFS_TRACE(...)
 #endif
 
 #ifndef LFS_NO_DEBUG
-#define LFS_DEBUG(fmt, ...) \
-    printf("lfs_debug:%d: " fmt "\n", __LINE__, __VA_ARGS__)
+#define LFS_DEBUG_(fmt, ...) \
+    printf("lfs_debug:%d: " fmt "%s\n", __LINE__, __VA_ARGS__)
+#define LFS_DEBUG(...) LFS_DEBUG_(__VA_ARGS__, "")
 #else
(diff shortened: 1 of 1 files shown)


geky:
Hi @hemmick, thanks for creating a PR. Unfortunately this doesn't work in standard c99 which means if we start adopting zero-arg prints we'll start writing code that _can't_ be ported to other standard compilers.

So the current solution is to avoid zero-arg prints. It's not a good solution but it's the only solution the standard offers us.

---

That was the current state of things, but it's been getting more and more annoying in a few other projects I've been working on so I looked around for more info and did find one portable hack that would get around this. It's not the greatest, but it does have the expected semantics:

``` c
#ifndef LFS_NO_DEBUG
#define LFS_DEBUG_(fmt, ...) \
    printf("waka_debug:%d: " fmt "%s\n", __LINE__, __VA_ARGS__)
#define LFS_DEBUG(...) \
    LFS_DEBUG_(__VA_ARGS__, "")
#else
#define LFS_DEBUG(fmt, ...)
#endif
```

I think that's what we should go with. And if the `##__VA_ARGS__` is available users can override the LFS_DEBUG declaration with their own locally.

We could detect some compilers and use `##__VA_ARGS__` where available, but messing around with this in GCC I kept getting a "ISO C99 requires at least one argument" warning. Searching around as far as I can tell this is impossible to disable with `-pedantic`, which is too useful to turn off.

It's still possible for users to override LFS_DEBUG locally if they're compiling without `-pedantic`, so I think this is the best solution available to us.

dpgeorge:
FWIW, the way we deal with this kind of problem in MicroPython is to have 2 macros, one that takes no variable arguments and one that does (and must take at least one): https://github.com/micropython/micropython/blob/master/py/compile.c#L67-L68

geky:
@dpgeorge, that is a good way to handle it! It hadn't even crossed my mind to just make a separate macro for single args. I'll have to pocket this for the future.

I still want to consider the `printf("debug: " fmt "%s\n", args, "")` option above for the ergonomics. 

As far as I can tell the only downside is the runtime overhead of parsing another argument. But the logging functions are exactly expected to be performance sensitivity or lightweight. And if the overhead is a concern, the macros can be locally overwritten to use toolchain specific tricks like @hemmick's PR.

Interested in other thoughts though.

hemmick:
@geky Thanks! Excellent write-up -- I will use your new method

geky:
Hi @hemmick, did you have a chance to try the LFS_DEBUG indirection? Is it possible to update this PR? If not I can bring it in and make the suggested change in the next release.

geky:
@hemmick Thanks for the original PR! The improved debug statements should now be available on master.

Issue #295, opened 2019-09-25
assert block != LFS_BLOCK_NULL in lfs_bd_read failing in lfs_ctz_traverse


Ralim:
Hi,

Currently working on getting Littlefs running on a NAND flash.
The flash is 256MB, with erase pages of 128K and and read/write sizes of 2048 bytes.

Running into an issue, were after a long running series of reads and writes to the flash, something is becoming unhappy and the assert [here](https://github.com/ARMmbed/littlefs/blob/fd204ac2fb4e05ffce6db6d1be67e0342c267cd7/lfs.c#L32) is failing.

When this assert fails, this is the current call stack:

- lfs_bd_read
- lfs_ctz_traverse
- lfs_alloc
- lfs_dir_alloc
- lfs_dir_split
- lfs_dir_compact
- lfs_dir_commit
- lfs_file_opencfg

Looking in `lfs_ctz_traverse`, at the time of the read call before the one that fails the assert, the read is from block 1820 (`head`), and the result that is read into the buffer is all 0xFF (validated read is correct by manually reading this block as well).

Any ideas on what could have lead to this?
This appears to be showing up on all units after a decent number of file actions, (around 100 files created and then deleted). Files are mix of small (2k) and large (25MB).

I can look into dumping the entire flash memory out if that would be of use?

This is the config structure:
```

const struct lfs_config Filesystem::LittleFScfg = {
// block device operations
		context:0, //
		read : Filesystem::raw_read,  //
		prog : Filesystem::raw_prog,  //
		erase : Filesystem::raw_erase,  //
		sync : Filesystem::raw_sync,  //

		// block device configuration
		read_size : RawFlash::get_read_size(),  //minimum read size
		prog_size : RawFlash::get_program_size(),  // minimum programming size
		block_size : RawFlash::get_erase_size(),  //Erase block size
		block_count : RawFlash::size() / RawFlash::get_erase_size(),  // Number of eraseable blocks
		block_cycles:1000,  // Cycle blocks around every 1k writes
		cache_size :LFS_BUFFER_SIZE,  //Size of the cache buffer
		lookahead_size : LFS_LOOKAHEAD_SIZE,  // Size of the lookahead scratch buffer
		read_buffer : lfs_read_buffer,  // Static read buffer
		prog_buffer : lfs_prog_buffer,  // Static programming buffer
		lookahead_buffer:lfs_lookahead_buffer,  // Static lookahead buffer
		name_max: 65,  // Max length of file names
[...]

PR #329, opened 2019-11-14
Fixed issue with directories falling out of date after block relocation

geky:
This is caused by dir->head not being updated when dir->m.pair may be. This causes the two to fall out of sync and later dir rewinds to fail.

~**Note:** Needs tests, there may be other issues related to out-of-sync directory heads and seek/rewinds.~

EDIT: Has tests, but now it needs those tests to _work_

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,6 +55,7 @@ test: \
 	test_attrs \
 	test_move \
 	test_orphan \
+	test_relocations \
 	test_corrupt
 	@rm test.c
 test_%: tests/test_%.sh
--- a/lfs.c
+++ b/lfs.c
@@ -1503,9 +1503,13 @@ static int lfs_dir_compact(lfs_t *lfs,
                 }
             }
 #ifdef LFS_MIGRATE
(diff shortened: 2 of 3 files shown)


Ralim:
Would this be a possible cause for #295 ?
If so, happy to look into testing this

geky:
@ralim, skimming #295, I don't think so. I think this bug really only happens if you call lfs_dir_rewind.

Unfortunately, the line that causes `assert(block == LFS_BLOCK_NULL)` to fail is very generic and can be caused by all sorts of issues. It's sort of like a hardfault. Sorry I haven't been able to look into it yet.

Ralim:
@geky
No issue, just was curious if it could result in that down the line.
In my code I do, do a fair amount of seeking over files opened read only, while also (possibly) writing to other files at the same time. I have locking on all access to the lfs struct, but wasnt sure if this could be something that resulted in that assert failing.

I understand that issue is a pain to debug :cry:

geky:
> seeking over files opened read only, while also (possibly) writing to other files at the same time.

I believe this PR only fixes issues related to directory seeking (not files, the two pieces of logic are actually unrelated).

PR #339, opened 2019-11-27
according to spec tag 0x00000000 is not valid

fsulima:
I have encountered crash in lfs_dir_fetchmatch when retrieved tag was 0x00000000 and callback was 0x0. This is the suggested fix.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -266,7 +266,7 @@ typedef int32_t lfs_stag_t;
     (((lfs_tag_t)(type) << 20) | ((lfs_tag_t)(id) << 10) | (lfs_tag_t)(size))
 
 static inline bool lfs_tag_isvalid(lfs_tag_t tag) {
-    return !(tag & 0x80000000);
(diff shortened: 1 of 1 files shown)


geky:
Hi @fsulima, thanks for creating a PR. This actually looks like it is a duplicate of https://github.com/ARMmbed/littlefs/pull/337 (discussion here https://github.com/ARMmbed/littlefs/issues/184).

Let me know if I'm incorrect.

fsulima:
Hello @geky , you are correct, but it looks like the fix I suggest is little more general.

geky:
That's true, though I'm not sure it's needed. As of #337, littlefs does not rely on tag 0x00000000 being invalid.

It isn't a valid tag in the sense that littlefs should never emit it (as a side-effect of the tag encoding), but this is just intended as a convenience for external tools and future filesystem development.

geky:
Closing as resolved by https://github.com/ARMmbed/littlefs/pull/337, feel free to reopen if I've made a mistake.

PR #342, opened 2019-11-30
Make littlefs magic configurable

roykuper13:
Make littlefs superblock magic configurable.

Can be useful, for example, when:
1. You want to distinguish between 2 or more "visible" storage (For example, when using [littlefs-fuse](https://github.com/ARMmbed/littlefs-fuse) multiple times, and you want to distinguish between the multiple files that are created).

2. You want to make your storage completely unreadable (For example, in case of using the littlefs-fuse, encrypting the file that is used to manage your storage is not enough to make the file completely unreadable, as simple `hd` on the file will reveal the "littlefs" string..)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,16 @@ endif
 ifdef TRACE
 override CFLAGS += -DLFS_YES_TRACE
 endif
+
+ifdef LFS_MAGIC
+MAGIC_LEN=$(shell echo -n $(LFS_MAGIC) | wc -c)
+ifeq ($(MAGIC_LEN), 8)
+override CFLAGS += -DLFS_MAGIC=\"$(LFS_MAGIC)\"
+else
+$(warning Invalid littlefs magic (length must be 8)! default littlefs magic will be used.)
+endif
+endif
+
 override CFLAGS += -I.
 override CFLAGS += -std=c99 -Wall -pedantic
 override CFLAGS += -Wextra -Wshadow -Wjump-misses-init -Wundef
--- a/lfs.c
+++ b/lfs.c
@@ -10,6 +10,11 @@
 #define LFS_BLOCK_NULL ((lfs_block_t)-1)
 #define LFS_BLOCK_INLINE ((lfs_block_t)-2)
 
+#ifndef LFS_MAGIC
+#define LFS_MAGIC "littlefs"
+#endif
+
+
 /// Caching block device operations ///
 static inline void lfs_cache_drop(lfs_t *lfs, lfs_cache_t *rcache) {
     // do not zero, cheaper if cache is readonly or only going to be
@@ -3542,7 +3547,7 @@ int lfs_format(lfs_t *lfs, const struct lfs_config *cfg) {
         lfs_superblock_tole32(&superblock);
         err = lfs_dir_commit(lfs, &root, LFS_MKATTRS(
                 {LFS_MKTAG(LFS_TYPE_CREATE, 0, 0), NULL},
-                {LFS_MKTAG(LFS_TYPE_SUPERBLOCK, 0, 8), "littlefs"},
+                {LFS_MKTAG(LFS_TYPE_SUPERBLOCK, 0, 8), LFS_MAGIC},
                 {LFS_MKTAG(LFS_TYPE_INLINESTRUCT, 0, sizeof(superblock)),
                     &superblock}));
         if (err) {
@@ -3602,7 +3607,7 @@ int lfs_mount(lfs_t *lfs, const struct lfs_config *cfg) {
                 LFS_MKTAG(LFS_TYPE_SUPERBLOCK, 0, 8),
                 NULL,
                 lfs_dir_find_match, &(struct lfs_dir_find_match){
-                    lfs, "littlefs", 8});
+                    lfs, LFS_MAGIC, 8});
         if (tag < 0) {
             err = tag;
             goto cleanup;
@@ -4472,7 +4477,7 @@ static int lfs1_mount(lfs_t *lfs, struct lfs1 *lfs1,
             lfs->lfs1->root[1] = superblock.d.root[1];
         }
 
-        if (err || memcmp(superblock.d.magic, "littlefs", 8) != 0) {
+        if (err || memcmp(superblock.d.magic, LFS_MAGIC, 8) != 0) {
             LFS_ERROR("Invalid superblock at %d %d", 0, 1);
             err = LFS_ERR_CORRUPT;
             goto cleanup;
@@ -4712,7 +4717,7 @@ int lfs_migrate(lfs_t *lfs, const struct lfs_config *cfg) {
         lfs_superblock_tole32(&superblock);
         err = lfs_dir_commit(lfs, &dir2, LFS_MKATTRS(
                 {LFS_MKTAG(LFS_TYPE_CREATE, 0, 0), NULL},
-                {LFS_MKTAG(LFS_TYPE_SUPERBLOCK, 0, 8), "littlefs"},
+                {LFS_MKTAG(LFS_TYPE_SUPERBLOCK, 0, 8), LFS_MAGIC},
                 {LFS_MKTAG(LFS_TYPE_INLINESTRUCT, 0, sizeof(superblock)),
                     &superblock}));
         if (err) {


geky:
Hi @roykuper13, thanks for creating a PR.

Something to watch out for: The magic number may not be present in block 0.

Why this is:

1. Like directories, the superblock is a "metadata pair". This means that the superblock is actually a log residing on blocks 0 and 1, and the data alternates between the two blocks. This is so the filesystem can erase the other block as a part of compaction.

   The reason the superblock is not entirely static is because it contains a pointer to the root directory. Which, because of copy-on-write operations, changes more often than you'd think. Because of this we actually just store the superblock as a part of the root directory itself and avoid the indirection (as of v2).

2. There is also an extra mechanism to prevent exhausting the superblock where if littlefs notices excessive superblock compacts (> cfg.block_cycles) it will expand a chain of indirect superblocks. This means the superblock may not even be in the anchor pair {0,1}, but instead further down the chain.

   Sidenote: I've wanted to create extra superblock entries on each pair in this chain, since it could help recover a corrupted filesystem. But littlefs currently doesn't do this. It would also create some issues with out-of-date superblocks if you could change the filesystem configuration (https://github.com/ARMmbed/littlefs/issues/279).

Even if you look at the problem generically, we can't keep the magic number in the superblock, since we'd need to eventually erase block 0 before updating the block, and if you lose power, it'd lose the magic number. In theory we could make block 0 only static, but this wouldn't work on very small devices, where it's possible to put littlefs on a small two-block section of internal memory.

This is actually really annoying, it means we can't determine the block_size at mount time, since we need the block_size to read the metadata pairs! Any suggestions on how to solve this are welcome.

So building any tools that rely on the magic-number is risky.

--- 

If you're ok determining the filesystem after mount, lfs_setattr/lfs_getattr may be useful. You can assign custom attributes to the root directory, effectively adding custom fields to the filesystem:

``` c
#define MY_MAGIC_ATTR 0x6d
#define MY_MAGIC_STRING "magi"

// format
int err = lfs_format(&lfs, &cfg);
if (err) { return err; }
err = lfs_mount(&lfs, &cfg);
if (err) { return err; }
err = lfs_setattr(&lfs, "/", MY_MAGIC_ATTR, MY_MAGIC_STRING, 4);
if (err) { return err; }
err = lfs_unmount(&lfs);
if (err) { return err; }

// mount
int err = lfs_mount(&lfs, &cfg);
if (err) { return err; }
char magic[4];
err = lfs_getattr(&lfs, "/", MY_MAGIC_ATTR, MY_MAGIC_STRING, 4);
if (err != 4 || strcmp(magic, MY_MAGIC_STRING) != 0) {
    lfs_unmount(&lfs, &cfg);
    return LFS_ERR_INVAL;
}
```

---

As a sidenote:

> You want to make your storage completely unreadable (For example, in case of using the littlefs-fuse, encrypting the file that is used to manage your storage is not enough to make the file completely unreadable, as simple hd on the file will reveal the "littlefs" string..)

I don't think I understand this point. If you want to make your storage completely unreadable, you need to do encryption on the entire block device, at the block level. Shouldn't this encrypt the superblock as well and hide the "littlefs" string?

roykuper13:
@geky Thanks for the response.
Actually I didn't try to mess with the superblock magic goal/usage.. I just wanted to make the hardcoded value of "littlefs" configurable.

About the unreadable storage (Correct me if i'm wrong/missing anything) - So I'm currently using the littlefs-fuse project, and wanted to encrypt my files stored in the file created in `lfs_fuse_bd_create`, and generally make the file completly unreadable. I need to encrypt using AES-ECB.
Now, If i'll encrypt in the "block level", i.e encrypt the buffer that `lfs_fuse_bd_prog` recieves (correct me if that wasn't what you meant in "block level"), the "littlefs" would be encrypted as well.
But - imo, the better approach would be to encrypt the entire buffer passed to `lfs_file_write`, because as far as i understand (from reading the code and from actual testing it), `lfs_fuse_bd_prog` recieves chunks of the buffer that is passed to `lfs_file_write`. The size of each chunk is `prog_size`. It means that i'll have to add padding to each chunk (because `lfs_fuse_bd_prog` is "standalone" prog function, it shouldn't be aware if the chunk that it handles is the final chunk of the file, or the middle chunk. Also `prog_size` might not be dividable by AES block size.. ) -> and therefore:
* the size of the file increases dramatically (With my suggested approach, I'll have to pad the buffer passed to `lfs_file_write` just once, so the file size only increases by the padded bytes which is up to AES_BLOCK_SIZE bytes).
* the `off` parameter would be incorrect because it doesn't take the padding that we added to the previous chunk into consideration, so i'll have to add code to fix it.

Also from the design standpoint, i don't think that encryption/decryption code should be part ofthe block device functions, and in littlefs functions. In my project for example, I have multiple storage "contexts" (e.g - A file generated by `lfs_fuse_bd_create` which stores encrypted files **AND** a file generated by `lfs_fuse_bd_create` which stores plain text files, also I have a context which stores files in an allocated buffer, etc), and I have a top-level layer which wrapps them all, and it deals with encryption, decryption etc, so these contexts wouldn't have to mess with that (and therefore it'll be easy to add more storage contexts).

So with my suggested approach, the only thing that is left readable is the "littlefs" hardcoded value. Generally I believe that hardcoded values should be configurable.. it doesn't increase the size of the final binary, so i can't see why not doing that..

roykuper13:
@geky I've changed every usage of the work "superblock magic" to simply "lfs magic" (in the code, in the commit and in the title of this PR). As far as I know I can't change the name of my branch in github (after PR is opened). Tell me if you want me to open a new PR with an updated src branch name.

thrasher8390:
@roykuper13, 

I'm not sure how the proposed solution solves issue #1. It will require you to build lfs multiple times with your different `LFS_MAGIC` defined. I think utilizing the attributes as @geky pointed out makes sense to me. 

In regard to #2, I've thought about that a bit and am curious why you want to make the WHOLE of external memory encrypted. e.g. directory information and file names. It does make sense to encrypt the file information and at that point I agree with one of your options to encrypt the file data before calling `lfs_file_write`. One option that I was considering for a file that is "encrypted" is to add an attribute to that file that contains information saying that the file is encrypted. 

Similar to #2: Integrity checking. You could add an attribute that holds the integrity CRC for a given file.

roykuper13:
> @roykuper13,
> 
> I'm not sure how the proposed solution solves issue #1. It will require you to build lfs multiple times with your different `LFS_MAGIC` defined. I think utilizing the attributes as @geky pointed out makes sense to me.
> 
> In regard to #2, I've thought about that a bit and am curious why you want to make the WHOLE of external memory encrypted. e.g. directory information and file names. It does make sense to encrypt the file information and at that point I agree with one of your options to encrypt the file data before calling `lfs_file_write`. One option that I was considering for a file that is "encrypted" is to add an attribute to that file that contains information saying that the file is encrypted.
> 
> Similar to #2: Integrity checking. You could add an attribute that holds the integrity CRC for a given file.

Hey,

So yeah I agree with you about the 1st thing.

But this PR originally was created because of the 2nd issue: In my project, I need the entire fs to be unreadable. It doesn't necessarily mean that I need to encrypt everything, I can take other approaches regarding file names (but yeah, I need them to be unreadable as well..). But the main thing that I have no control on is the "littlefs" magic string. It seems wrong to me that a project has hard-coded values (that can't be modified) especially if they can be exposed to the user interface. Most of the times it won't affect anything, but in my specific case, IT IS a problem.
placing this magic as a lfs attribute is an option, even though it seems redundant to me, so I believe that my changes should be merged for now..

@geky @thrasher8390 what do you guys think

About the integrity&encryption attribute flags - seems to be a good idea (i think), but it is out of this PR's scope..

geky:
Hi, sorry about the slow response, your PRs deserve better than that.

---

> In regard to #2, I've thought about that a bit and am curious why you want to make the WHOLE of external memory encrypted.

It seems reasonable to me. Encrypting the entire disk offers more security by reducing the amount of information that is available. By only encrypting the file's data, an attacker can still see the file's size, name, and do tricky things like deleting files.

The way I see it, if you're an application developer and you know what files should be there and it won't change from user to user, file-level encryption offers a performant method for securing the user's data.

If you're an OS developer or applications where you don't control the files, only encrypting the file data risks leaking information the user may be sensitive about.

It's also interesting to look at performance. On one hand, file-level encryption is faster for small files, since you encrypt the bare minimum. On the other hand, disk-level encryption (per block) can take advantage of the filesystem to split up large files across multiple encryption blocks. Though I believe file-level encryption will be faster in general.

Both approaches have merits, so littlefs is unopinionated here.

> Similar to #2: Integrity checking. You could add an attribute that holds the integrity CRC for a given file.

It's worth noting you can also store an HMAC or CMAC alongside the data to provide integrity for the disk. littlefs doesn't need power-of-two block sizes, so you can slip an HMAC/CMAC in each page.

---

> But the main thing that I have no control on is the "littlefs" magic string. It seems wrong to me that a project has hard-coded values (that can't be modified) especially if they can be exposed to the user interface.

The magic string is more similar to the version number stored next to it. It's used to sanity check that the filesystem is a littlefs filesystem and isn't corrupted. Changing either of these requires a code change because the filesystem no longer adheres to the littlefs specification. It's no longer a "littlefs".

This is what the fs-level user attribute was added for, user-specific attributes at the filesystem level that can be used to store additional info about the filesystem.

> placing this magic as a lfs attribute is an option, even though it seems redundant to me, so I believe that my changes should be merged for now..

Unfortunately, as a maintainer, I need to be adverse to new features without a good strong reason. If I merge this change, I need to support it for the foreseeable future, or risk breaking backwards compatibility.

Sorry, but I don't think this should land. It's still possible to use the fs-level attributes to mark the filesystem as needed.

---

> As far as I know I can't change the name of my branch in github 

Yeah, this is a limit on GitHub's end. I wouldn't worry about it, I've seen weirder branch names.

roykuper13:
> Hi, sorry about the slow response, your PRs deserve better than that.

don't worry, it's fine :)

> The magic string is more similar to the version number stored next to it. It's used to sanity check that the filesystem is a littlefs filesystem and isn't corrupted. Changing either of these requires a code change because the filesystem no longer adheres to the littlefs specification. It's no longer a "littlefs".

I'm aware that this magic string is used for the corruption check.
I don't understand why is the magic string must be "littlefs" and not a value that is configured at build-time? It does not matter what's the value of the magic string as long as the value doesn't change at run-time.

My changes just make sure that the filesystem uses that configurable magic string, and not a hard-coded value. The corruption check uses the same value, without being aware what's the value. I can't see what's the problem with that..

> This is what the fs-level user attribute was added for, user-specific attributes at the filesystem level that can be used to store additional info about the filesystem.
>

I'm not sure how's that related to my problem.. My goal is to use other magic string instead of "littlefs", using build time configuration (so, without changing the code each time I compile lfs). How can I do that?
 
> Unfortunately, as a maintainer, I need to be adverse to new features without a good strong reason. If I merge this change, I need to support it for the foreseeable future, or risk breaking backwards compatibility.
> 
> Sorry, but I don't think this should land. It's still possible to use the fs-level attributes to mark the filesystem as needed.

Oh, I didn't want to force you merge this of course :) I just want to find a good solution to my problem, and as of right now I can't find one..

PR #356, opened 2019-12-19
Update SPEC.md

zqb-all:
fix size in Layout of the CRC tag

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -289,8 +289,8 @@ Layout of the name tag:
 ```
         tag                          data
 [--      32      --][---        variable length        ---]
-[1| 3| 8 | 10 | 10 ][---            (size)             ---]
- ^  ^  ^    ^    ^- size               ^- file name
+[1| 3| 8 | 10 | 10 ][---          (size * 8)           ---]
+ ^  ^  ^    ^    ^- size                   ^- file name
  |  |  |    '------ id
  |  |  '----------- file type
  |  '-------------- type1 (0x0)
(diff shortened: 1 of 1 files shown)


SPEC.md
```diff
@@ -764,9 +764,9 @@ Layout of the CRC tag:
 ```
         tag                                    data
 [--      32      --][--      32      --|---        variable length        ---]
-[1| 3| 8 | 10 | 10 ][--      32      --|---            (size)             ---]
+[1| 3| 8 | 10 | 10 ][--      32      --|---          (size - 4)           ---]
```
geky: Should this actually be `(size - 32)` since the other fields are in bits?

SPEC.md
```diff
@@ -764,9 +764,9 @@ Layout of the CRC tag:
...
  ^  ^  ^    ^    ^            ^- crc                      ^- padding
- |  |  |    |    '- size (12)
```
geky: What does this 12 mean here? The size could be any value.

SPEC.md
zqb-all: yes, this field should in bits, but size is in bytes, maybe `(size * 8 - 32)` is better ?

geky: Ah! Good point, you're right that is better.

geky:
Hi @zqb-all, sorry about the absence. Does the updated comment make sense to you?

If you're not able to update the PR I can bring it in and make the suggested changes on the next release.

zqb-all:
Hi @geky, thanks for you reply, I am glad to update the PR.

geky:
Thanks for the contribution!

PR #359, opened 2019-12-30
Change assert to runtime check.

cdesjardins:
I had a system that was constantly hitting this assert, after making
this change it recovered immediately.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -29,8 +29,7 @@ static int lfs_bd_read(lfs_t *lfs,
         lfs_block_t block, lfs_off_t off,
         void *buffer, lfs_size_t size) {
     uint8_t *data = buffer;
-    LFS_ASSERT(block != LFS_BLOCK_NULL);
-    if (off+size > lfs->cfg->block_size) {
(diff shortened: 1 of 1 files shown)


geky:
Hi @cdesjardins, thanks for creating a PR!

I'm of the opinion this check should actually be done when we first find the bad block addresses (so in lfs_file_open, lfs_dir_open, lfs_stat, or similar).

I want to see if that can accomplished before the next release, if not I'll going ahead merge this in.

geky:
Ok! I've merged this into https://github.com/ARMmbed/littlefs/pull/372 and added tests and a couple of other checks.

I looked into different implementations of this check, but now think your original proposal is the best solution. Thanks again for the PR.

I will close this when https://github.com/ARMmbed/littlefs/pull/372 is merged. I don't think GitHub is smart enough to track cherry-picks across PRs.

geky:
Closing as this has been merged, unfortunately I don't think I merged it in a way that is reflected here in GitHub.

geky:
Thanks for the contribution!

PR #360, opened 2020-01-02
Fix incorrect comment on `lfs_npw2`

jpdoyle:
`lfs_npw2` returns a value v such that `2^v >= a` and `2^(v-1) < a`, but
the previous comment incorrectly describes it as "less than or equal to
a".


geky:
Hi @jpdoyle, sorry about the delay. Thanks for pointing this out, I will merge this in for the next release.

The rewritten description is also much more understandable 👍

geky:
Thanks for the contribution!

PR #372, opened 2020-01-27
Rework test framework, fix a number of related bugs

geky:
This took a bit longer than I thought it would, some because of the number of bugs that fell out, some because I underestimated how much time it would take to edit all of the tests.

EDIT: I did not realize how long the description of this PR would be, sorry for the wall of text. Also sorry about being out-of-touch for a while.

### Test framework rework

This PR reworks the testing framework to better isolate and fix filesystem bugs. This mainly involves generalizing the tests to let them run under different contexts (power-loss, wear, valgrind), while also adding extra scripts to make test results easier to reproduce and visualize.

- Reworked how test.py works. Now instead of just dispatching `tests/test_*.sh` scripts, it loads tests stored in `tests/test_*.toml` files. This lets us store extra attributes describing tests, such as "leaky" or "reentrant", which we can use to decide how to run tests later.

  - `./scripts/test.py` - Runs tests normally. Each test case is a single function that should terminate without asserting.
  - `./scripts/test.py --reentrant` - Run tests with a simulated power-loss occuring every _n_ cycles where _n_ = _n_+1 every power-loss. Only test cases marked with `reentrant = true` are ran, since the test itself also needs to be able to handle power-loss.
  - `./scripts/test.py --valgrind` - Run tests under valgrind, checking for memory constraint errors. Tests marked with `leaky = true` are skipped here.

  There are also the addition of "internal" tests, where test cases can specify `in = "lfs.c"` or some other file to be compiled as though it was appended to the file "lfs.c". This gives the test access to static functions and other internals. I'm planning to use these for more unit-style tests in the future.

  It's worth noting that tests can no longer span multiple sequential processes, but this feature wasn't that helpful and created problems for reproducibility and porting to devices.

  The real kicker is how test.py integrates with gdb. By adding --gdb, test.py will drop you you into gdb at the exact assert where a test fails. This includes reproducing any power-cycles necessary to get the test into the failing state. This is very helpful for reproducing complex test failures.

  ``` bash
  $ ./scripts/test.py --gdb
  ====== building ======
  built 15 test suites, 116 test cases, 740 permutations
  ====== testing ======
  test_alloc ✓✓✓✓✓✓✓✓✓✓✓✓✓
  test_attrs ✓✓✓✓
  test_badblocks ✓✓✓✓✓✓✓✓✓✓✓✓
  test_dirs ✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓
  ✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓
  test_entries ✓✓✓✓✓✓✓✓
  test_exhaustion ✓✓✓✓✓✓✓✓✓✓✓✓
  test_format ✓✓✓✓✗
  ====== results ======
  tests/test_format.toml:29:failure: test_format#5#1 (BLOCK_CYCLES=32, N=10) failed with -6
  tests/test_format.toml:39:assert: assert failed with 2, expected eq 1
          assert(info.type == LFS_TYPE_REG);

  ======= gdb ======
  GNU gdb (Ubuntu 7.11.1-0ubuntu1~16.5) 7.11.1
  Copyright (C) 2016 Free Software Foundation, Inc.
  License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
  This is free software: you are free to change and redistribute it.
  There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
  and "show warranty" for details.
  This GDB was configured as "x86_64-linux-gnu".
  Type "show configuration" for configuration details.
  For bug reporting instructions, please see:
  <http://www.gnu.org/software/gdb/bugs/>.
  Find the GDB manual and other documentation resources online at:
  <http://www.gnu.org/software/gdb/documentation/>.
  For help, type "help".
  Type "apropos word" to search for commands related to "word"...
  Reading symbols from ./tests/test_format.toml.test...done.
  Starting program: /home/geky/littlefs3/tests/test_format.toml.test 5 1
  tests/test_format.toml:39:assert: assert failed with 2, expected eq 1

  Program received signal SIGABRT, Aborted.
  0x00007ffff7a42428 in __GI_raise (sig=6) at ../sysdeps/unix/sysv/linux/raise.c:54
  54      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
  #2  0x0000000000401f17 in test_case5 (BLOCK_CYCLES=32, N=10) at tests/test_format.toml:39
  39              assert(info.type == LFS_TYPE_REG);
  (gdb)
  ```

- Restructured the test block devices. Now there is testbd, filebd, and rambd.

  The way that emubd created a separate file for each block wasn't that useful. It was marginally easier? But didn't match the single-file-block-device that was easy to create when dumping flash from an actual device.

  I've replaced emubd with simpler, and hopefully more useful, block devices:
  - filebd - Simulate a block device on a single file.

    As an added plus this block device can be used to mount disk images extracted from a device's flash. You can even use it to mount /dev/sdb devices into a C program.

  - rambd - Simulate a block device in RAM.

    The main reason for adding this is performance. Now that we don't allow separate-process tests, we can test using a block device stored in RAM. This speeds up the tests by ~7x.

    - test_dirs w/ filebd - 0m7.170s
    - test_dirs w/ rambd  - 0m0.966s

  On top of these is testbd, a wrapper for filebd/rambd with extra hooks for testing. This is where a lot of the test magic happens.

  - Simulated block wear - testbd can be configured to keep track of block wear, and force blocks to start reporting LFS_ERR_CORRUPT when exhausted. The current wear modified during a test to tightly control how the block device will behave under littlefs.

    Block wear can also be used to explicitly force geometry of the underlying block device, and lets us change the size of the block device on the fly while testing.

    We can also read the state of wear on the block device after a test, which may be useful for future benchmarks.

  - Simulated power loss - testbd can be give a number of "write cycles" to run for. This creates a counter that is decremented every prog/erase cycle. When the counter hits zero the program is forcefully terminated.

    Note this is much faster that using gdb to kill the program externally.

- Added a handful of debugging scripts for visualizing the state of littlefs.

  - `./scripts/readblock.py disk block_size block` - Prints a hex dump of a given block on disk. It's basically just `dd if=disk bs=block_size count=1 skip=block | xxd -g1 -` but with less typing.

    ``` bash
    $ ./scripts/readblock.py disk 512 0xa
    off       data
    00000000: 71 01 00 00 f0 0f ff f7 6c 69 74 74 6c 65 66 73  q.......littlefs
    00000010: 2f e0 00 10 00 00 02 00 00 02 00 00 00 04 00 00  /...............
    00000020: ff 00 00 00 ff ff ff 7f fe 03 00 00 20 00 04 19  ...............
    00000030: 61 00 00 0c 00 62 20 30 0c 09 a0 01 00 00 64 00  a....b 0......d.
    ...
    ```

  - `./scripts/readmdir.py disk block_size block1 block2` - Prints info about the tags in a metadata pair on disk. It can print the currently active tags as well as the raw log of the metadata pair.

    ``` bash
	$ ./scripts/readmdir.py disk 512 0xa 0xb
    off       tag       type            id  len  data (truncated)
    0000003b: 0020000a  dir              0   10  63 6f 6c 64 63 6f 66 66 coldcoff
    00000049: 20000008  dirstruct        0    8  02 02 00 00 03 02 00 00 ........
    00000008: 00200409  dir              1    9  68 6f 74 63 6f 66 66 65 hotcoffe
    00000015: 20000408  dirstruct        1    8  fe 01 00 00 ff 01 00 00 ........
	```

  - `./scripts/readtree.py disk block_size` - Parses the littlefs tree and prints info about the semantics of what's on disk. This includes the superblock, global-state, and directories/metadata-pairs.

    ``` bash
	$ ./scripts/readtree.py disk 512
    superblock "littlefs"
      version v2.0
      block_size 512
      block_count 1024
      name_max 255
      file_max 2147483647
      attr_max 1022
    gstate 0x000000000000000000000000
    dir "/"
    mdir {0x0, 0x1} rev 3
    v id 0 superblock "littlefs" inline size 24
    mdir {0x77, 0x78} rev 1
      id 0 dir "coffee" dir {0x1fc, 0x1fd}
    dir "/coffee"
    mdir {0x1fd, 0x1fc} rev 2
      id 0 dir "coldcoffee" dir {0x202, 0x203}
      id 1 dir "hotcoffee" dir {0x1fe, 0x1ff}
    dir "/coffee/coldcoffee"
    mdir {0x202, 0x203} rev 1
    dir "/coffee/warmcoffee"
    mdir {0x200, 0x201} rev 1
	```

- Added explode_asserts.py.

  What is explode_asserts.py? It is an extra C preprocessing-step that makes asserts aggressively readable.

  explode_asserts.py takes in a C file and parses any asserts it finds, transforming them into an equivalent statement that can also report the original arguments to the logic operation in the assert. This is very useful for debugging and removes the need for assert_int_eq, assert_str_lt, etc that are a common sight in C testing frameworks.

  And because explode_asserts.py works with traditional asserts, it can be applied to any existing project to get nicer assert results.

  Here's a test failure showing how explode_asserts.py can extract assert arguments:

  ``` bash
  geky@s:littlefs3$ ./scripts/test.py
  ====== building ======
  built 15 test suites, 116 test cases, 740 permutations
  ====== testing ======
  test_alloc ✓✓✓✓✓✓✓✓✓✓✓✓✓
  test_attrs ✓✓✓✓
  test_badblocks ✓✓✓✓✓✓✓✓✓✓✓✓
  test_dirs ✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✗
  ====== results ======
  tests/test_dirs.toml:48:failure: test_dirs#3#1 (N=3) failed with -6
  tests/test_dirs.toml:72:assert: assert failed with "removeme000", expected eq "removeme0"
          assert(strcmp(info.name, path) == 0);

  tests passed: 64
  tests failed: 1
  ```

### Bug fixes!

What was the point of all these test changes? ~It was to make my life easier~ Finding and fixing bugs! of which there were quite a few:

- Fixed bug in dir splitting when there's a large number of open files 9453ebd15d590778a5207920b2671a4fff91a9e0

  Updates open files were depending on the mutable dir rather than the immutable copy intended to avoid sync issues.

- Fixed issues with neighbor updates during moves f4b6a6b3284d8b721ba3db0bba93ee45fde7ead9

  This was found by @eastmoutain and initial fix proposed in https://github.com/ARMmbed/littlefs/issues/343 though there were a couple lingering issues that needed to fixed.

  This lingering issues turned out to be quite involved. There's a better writeup in f4b6a6b3284d8b721ba3db0bba93ee45fde7ead9's commit message.

  Basically the order and number of create/delete tags was assumed to behave a certain way when passed to lfs_dir_commit and this was wrong.

  The fix was to restructure how lfs_dir_commit tracks create/delete tags and how implicit moves in global state are handled (by making them no-longer implicit).

  I believe this was the source of most of the "LFS_ASSERT(block != LFS_BLOCK_NULL)" issues. Will be working through open issues to make sure.

- Fixed parent going out of sync when relocated in lfs_mkdir a5d614fbfbf19b8605e08c28a53bc69ea3179a3e

  This can happen if we're creating a new directory in a multi-pair directory, our predecessor has to be relocated as we thread ourselves into the linked-list.

  Fortunately this can be fixed by adding our temporary mdir to the list of open files/dirs, which are automatically updated on metadata-pair changes. Note we need to take care to unhook our temporary before lfs_mkdir returns.

- Fixed to-be-removed child going out of sync when relocated in lfs_remove and lfs_rename a5d614fbfbf19b8605e08c28a53bc69ea3179a3e

  Surprisingly this can happen if a nested directory is removed and it is its own parent's predecessor (which can happen after moving the directory around, since this does not change the threaded linked-list). If this happens, and our parent is relocated, we could go out of sync, causing us to update are predecessor with the incorrect tail pointer.

- Fixed lfs_rename with same source and destination a5d614fbfbf19b8605e08c28a53bc69ea3179a3e

  An uncommon operation, lfs_rename("deja/vu", "deja/vu"). This was broken and would result in two deletes to the same directory entry. Fixed and added test case specifically for this situation.

- Fixed how we manage gstate deltas when we lose power during nested relocations a5d614fbfbf19b8605e08c28a53bc69ea3179a3e

  Managing gstate deltas when we lose power during relocations was broken. And unfortunately complicated.

  The issue happens when we remove a directory, have to relocate its parent, and lose power.

  When we remove a directory, we need to move the contents of its gstate delta to another directory or we'll corrupt littlefs gstate. (gstate is an xor of all deltas on the filesystem). This used to go to the parent, but this actually isn't correct since the threaded linked-list is used to read the gstate, not the directory tree.

  The fix here is to take special care about where we put the gstate delta. Delta depending on removes needs to go to the predecessor, while orphan-related delta describing the remove itself needs to go to the parent.

- Fixed lfs_deorphan so it can handle more than one orphan/half-orphan during a power-loss a5d614fbfbf19b8605e08c28a53bc69ea3179a3e

  Two bugs here:
  1. lfs_deorphan couldn't handle handle multiple orphans in a single power-loss. We can't have multiple _orphans_, but we can have multiple _half-orphans_ (where only one block address is out-of-date in the metadata pair). This happens when we have nested relocations.

  2. lfs_deorphan needs to relocate the current pair if it is a half-orphan. This is because if both a directory and its immediate successor is relocated, the tail pointer we fetch isn't valid until after we fix the half-orphan.

  Fortunately the fix here is simpler than the bug and we just need to carefully handle how lfs_deorphan scans the threaded linked-list of metadata-pairs.

- Fixed reproducability issue when we can't read a directory revision 517d3414c5e04eedb07be2e58107c1f96b8b8684

  An interesting subtlety of the block-device layer is that the block-device is allowed to return LFS_ERR_CORRUPT on reads to untouched blocks. This can easily happen if a user is using ECC or some sort of CMAC on their blocks. Normally we never run into this, except for the optimization around directory revisions where we use uninitialized data to start our revision count.

  We correctly handle this case by ignoring whats on disk if the read fails, but end up using unitialized RAM instead. This is not an issue for normal use, though it can lead to a small information leak. However it creates a big problem for reproducability, which is very helpful for debugging.

  I ended up running into a case where the RAM values for the revision count was different, causing two identical runs to wear-level at different times, leading to one version running out of space before a bug occured because it expanded the superblock early.

- Fixed incorrect erase assumption if lfs_dir_fetch exceeds block size 517d3414c5e04eedb07be2e58107c1f96b8b8684

  This could be caused if the previous tag was a valid commit and we lost power causing a partially written tag as the start of a new commit.

  Fortunately we already have a separate condition for exceeding the block size, so we can force that case to always treat the mdir as unerased.

- Fixed cleanup issue caused by lfs_fs_relocate failing when trying to outline a file in lfs_file_sync 517d3414c5e04eedb07be2e58107c1f96b8b8684

  Most operations involving metadata-pairs treat the mdir struct as entirely temporary and throw it out if any error occurs. Except for lfs_file_sync since the mdir is also a part of the file struct.

  This is relevant because of a cleanup issue in lfs_dir_compact that usually doesn't have side-effects. The issue is that lfs_fs_relocate can fail. It needs to allocate new blocks to relocate to, and as the disk reaches its end of life, it can fail with ENOSPC quite often.

  If lfs_fs_relocate fails, the containing lfs_dir_compact would return immediately without restoring the previous state of the mdir. If a new commit comes in on the same mdir, the old state left there could corrupt the filesystem.

  It's interesting to note this is forced to happen in lfs_file_sync, since it always tries to outline the file if it gets ENOSPC (ENOSPC can mean both no blocks to allocate and that the mdir is full). I'm not actually sure this bit of code is necessary anymore, we may be able to remove it.

- Fixed missing half-orphans when allocating blocks during lfs_fs_deorphan 517d3414c5e04eedb07be2e58107c1f96b8b8684

  This was a really interesting bug. Normally, we don't have to worry about allocations, since we force consistency before we are allowed to allocate blocks. But what about the deorphan operation itself? Don't we need to allocate blocks if we relocate while deorphaning?

  It turns out the deorphan operation can lead to allocating blocks while there's still orphans and half-orphans on the threaded linked-list. Orphans aren't an issue, but half-orphans may contain references to blocks in the outdated half, which doesn't get scanned during the normal allocation pass.

  Fortunately we already fetch directory entries to check CTZ lists, so we can also check half-orphans here. However this causes lfs_fs_traverse to duplicate all metadata-pairs, not sure what to do about this yet.

- Fixed commit issue if block cannot be written to and contains valid commits  77e3078b9f28ff689e5a623250eae5c85ae8b3aa

  We can miss bad writes if the underlying block device contains a valid commit with the exact same size as in the exact same offset.

  The fix here is to check the CRC on disk with the CRC we have in RAM. This avoids the issue without needed to save our entire commit in RAM.

  It could be an issue if the disk contains a valid commit with exact same size at the exact same offset _and_ the exact same CRC. But this is unlikely enough to never happen.

- Fixed broken wear-leveling when block_cycles = 2n-1 fe957de892edbefd7af54acbe19e9a4e0f0fb1d0

  See https://github.com/ARMmbed/littlefs/issues/369

  Blocks in the metadata-pair are relocated every "block_cycles", or, more mathy, when rev % block_cycles == 0 as long as rev += 1 every block write.

  But there's a problem, rev isn't += 1 every block write. There are two blocks in a metadata-pair, so looking at it from each blocks perspective, rev += 2 every block write.

  This leads to a sort of aliasing issue, where, if block_cycles is divisible by 2, one block in the metadata-pair is always relocated, and the other block is _never_ relocated. Causing a complete failure of block-level wear-leveling.

  Fortunately, because of a previous workaround to avoid block_cycles = 1 (since this will cause the relocation algorithm to never terminate), the actual math is rev % (block_cycles+1) == 0. This means the bug only shows its head in the much less likely case where block_cycles is a multiple of 2 plus 1, or, in more mathy terms, block_cycles = 2n+1 for some n.

  To workaround this we can bitwise or our block_cycles with 1 to force it to never be a multiple of 2n.

- Fixed lfs_fs_size doubling metadata-pairs 6530cb3a6129f28dda117fdcb0231291b0eff680

  This was caused by the previous fix for allocations during lfs_fs_deorphan in this branch. To catch half-orphans during block allocations we needed to duplicate all metadata-pairs reported to lfs_fs_traverse. Unfortunately this causes lfs_fs_size to report 2x the number of metadata-pairs, which would undoubtably confuse users.

  The fix here is inelegantly simple, just do a different traversale for allocations and size measurements. It reuses the same code but touches slightly different sets of blocks.

  Unfortunately, this causes the public lfs_fs_traverse and lfs_fs_size functions to split in how they report blocks. This is technically allowed, since lfs_fs_traverse may report blocks multiple times due to CoW behavior, however it's undesirable and I'm sure there will be some confusion.

  But I don't have a better solution, so from this point lfs_fs_traverse will be reporting 2x metadata-blocks and shouldn't be used for finding the number of available blocks on the filesystem.

- Fixed lfs_alloc lookahead corruption if littlefs hits a bd error during traversal

   This issue was caused by errors returned from the block device during a lookahead scan. littlefs aborts the scan to return the errors, but at that point has already cleared the lookahead buffer, putting the allocator into a state where is thinks blocks are free when they may be in use.

   The fix is to mark the lookahead buffer as empty if the allocator must abort. This will force a new lookahead scan on the next alloc call. This means a filesystem may be unable to make progress, but at least gives the user a chance to recover.

   This was found and fixed by @thrasher8390 in https://github.com/ARMmbed/littlefs/pull/401

### Related issues

WIP: I'm currently going through open issues to figure out what is impacted. I'm currently behind on this.

- https://github.com/ARMmbed/littlefs/issues/343
- https://github.com/ARMmbed/mbed-os/issues/11704
- https://github.com/ARMmbed/littlefs/issues/295

### TODO

- [x] Add power-cycling tests
- [x] Add better support for debugging single-file block devices
- [x] RAM block device?
- [x] Add more tests around relocations under power-loss
- ~Fix non-DAG wear-leveling~
- [x] Investigate https://github.com/ARMmbed/littlefs/issues/369
- ~Investigate https://github.com/ARMmbed/littlefs/issues/379~
- [x] Fix CI to use new framework
- [x] Add tests for recovering from extra bad states
  - Recover from finding bad block addresses on disk
  - Loop detection?
- ~Code coverage? (optional)~
- [x] Add minor bump to this


lfs.c
```diff
@@ -480,7 +475,7 @@ static int lfs_alloc(lfs_t *lfs, lfs_block_t *block) {
 
         // find mask of free blocks from tree
         memset(lfs->free.buffer, 0, lfs->cfg->lookahead_size);
-        int err = lfs_fs_traverse(lfs, lfs_alloc_lookahead, lfs);
+        int err = lfs_fs_traverseraw(lfs, lfs_alloc_lookahead, lfs, true);
         if (err) {
             return err;
```
thrasher8390: Should reset .i, .size, .off incase of err, we want to make sure that we force a traverse next time lfs_alloc is called. 
I'm going to make a PR related to https://github.com/ARMmbed/littlefs/issues/394 but may need help updating the testing suite as I'm not certain the best way to make that change.

geky:
Ok, I believe it's time to bring this PR in. There's still a number of open issues that would be nice to fix (specifically non-DAG trees and https://github.com/ARMmbed/littlefs/issues/379), but this PR has been open for long enough already and additional fixes can hopefully come in on later patch releases.

v2.2 will include this PR along with any open PRs tagged [`v2.2`](https://github.com/ARMmbed/littlefs/pulls?q=is%3Aopen+is%3Apr+label%3Av2.2)

There's going to be a delay as we wait for CI (I've lost permissions to kill unnecessary Travis jobs and need to sort that out), but hopefully the v2.2 release will be done in the next couple of days.

PR #373, opened 2020-01-28
Indicate C99 standard as target for LittleFS code

### LINKED ISSUE #358: SPEC.md or DESIGN.md should specify target version of C standard (c99)
### Summary

It appears that LFS is targeting the `C99` standard as the minimum C standard that is required to compile the library.  However, this is not mentioned in the [README.md](https://github.com/ARMmbed/littlefs/blob/ab56dc5a8b9b06e3cd2b84099c140ade636f3de5/README.md), [DESIGN.md](https://github.com/ARMmbed/littlefs/blob/f7fd7d966a9d8c6e783a95cd289401f42045a73c/DESIGN.md), nor [SPEC.md](https://github.com/ARMmbed/littlefs/blob/bdff4bc59eb69bc403e0f5878e72a96a0a21a190/SPEC.md) documents.

It would be helpful to explicitly list this requirement in at least one of those documents.
My vote would be to add it as a goal in the top of the README.md (as part of the list including `Power-loss resilience`, `Dynamic wear leveling`, and `Bounded RAM/ROM`)

### Why `c99`?

I realized that, although I could offer suggestions (or perhaps even PRs) for issues in LFS, I could not do so without knowing at least the minimum C version that it would need to work with.  Since none of the docs (listed above) indicated this, I dug in to find the minimum version.   The Travis.yml file, as part of the CI checks, compiles with `-std=c99`.  This leaves only `c90` (aka `c89` or `ansi`), `iso9899:199409` (aka `c94` or `c95`) as the earlier versions.  Neither `c90` nor `c95` support `// C++-style comments`, and therefore the minimum target version must be `c99`.

henrygab:
Resolve #358

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,9 @@ the filesystem until sync or close is called on the file.
 
 ## Other notes
 
+Littlefs is written in C, and specifically should compile with any compiler
+that conforms to the `C99` standard.
+
 All littlefs calls have the potential to return a negative error code. The
 errors can be either one of those found in the `enum lfs_error` in
 [lfs.h](lfs.h), or an error returned by the user's block device operations.


geky:
Ah thanks for this! This looks great. Will merge next release.

PR #384, opened 2020-02-26
emubd: fix MSVC compilation

Mixaill:
fixes emubd compilation with pure MSVC compiler.

### DIFF
--- a/emubd/lfs_emubd.c
+++ b/emubd/lfs_emubd.c
@@ -12,11 +12,25 @@
 #include <stdio.h>
 #include <limits.h>
 #include <sys/stat.h>
-#include <unistd.h>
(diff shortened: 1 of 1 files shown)


geky:
Hi @Mixaill, thanks for the PR!

The emubd layer is actually changing significantly here:
https://github.com/ARMmbed/littlefs/pull/372

Though I'm not sure how https://github.com/ARMmbed/littlefs/pull/372 would effect MSVC.

geky:
Thanks for the PR. But I'm going to go ahead and close this as the bd layer was rewritten in v2.2. It no longer calls mkdir, so I'm not sure what changes need to be made for MSVC support.

PR #397, opened 2020-03-15
build: Use LTO

roykuper13:
reduced my binary size by ~13k


thrasher8390:
@roykuper13, what are you reporting a 13k decrease too? I had done a analysis about a year ago and lfs was about 11k in total size so I'd like to understand exactly what you are using as your metric. 

Also, why do you rely on the lfs makefile for your personal project?

geky:
Hi @roykuper13 thanks for creating a PR!

@thrasher8390 You can kinda see what's going on if you click the "show all checks" on the CI status.

> Passed, code size is 0B (-100.00%) 

With LTO it looks like CI is optimizing out littlefs completely. This is probably a quirk of how the CI finds the build size.

Humorously, most of littlefs is in a single .c file which sort of acts as a poor man's LTO (though this wasn't the intention). So I don't think there would be much gain from compiling the library with LTO.

roykuper13:
@geky @thrasher8390 Hey, thanks for the responses!

> @roykuper13, what are you reporting a 13k decrease too? I had done a analysis about a year ago and lfs was about 11k in total size so I'd like to understand exactly what you are using as your metric.
> 
> Also, why do you rely on the lfs makefile for your personal project?

I only created a simple test.c file that basically contains the example of README.md and compiled it with `make lfs`.

For your second question, I rely on the lfs makefile when I want to run simple tests/samples. For these kind of things, I don't create static lfs library and link it with a small sample, I just rely on the makefile. The makefile already have a `lfs` target which seems to be intended for those kind of situations (`ifneq ($(wildcard test.c main.c),)` in the 2nd line...). I expect my samples/tests to be as small as possible as well, without unused symbols, at all. and therefore I believe lto should be used.

Also, I use lfs as a submodule in my project. I still want to use lfs makefile, and compile it as a static library (`make lfs.a`). For this case, I just added a commit which puts the symbols in different sections, and in my project's makefile I eventually use `--gc-sections` which removes the unused symbols of littlefs (Compile lfs files with lfs Makefile seems the right approach to me, rather then compile lfs files in my own build files, especially if lfs's makefile "offers" a lfs.a target).

In both cases btw, If i just use `lfs_mount` and `lfs_unmount`, it saves me more then 30k. it is a huge different.

geky:
I'm happy to bring this in if someone can fix this so it doesn't report 0 for `make size`. The CI invocation for this is here:
https://github.com/ARMmbed/littlefs/blob/ce2c01f098f4d2b9479de5a796c3bb531f1fe14c/.travis.yml#L42-L45

I do think it's fairly atypical to use littlefs's Makefile for anything other than running tests or examples locally.

It would be better to copy the Makefile into the root of your project and make project-specific changes outside of littlefs. Otherwise there is the risk changes we make to the Makefile here could break your project's compilation.

Makefile
```diff
@@ -33,6 +35,7 @@ override CFLAGS += -Wextra -Wshadow -Wjump-misses-init -Wundef
 # Remove missing-field-initializers because of GCC bug
 override CFLAGS += -Wno-missing-field-initializers
 
+override CFLAGS += $(EXTRA_CFLAGS)
```
geky: LTO should be moved into the `ifdef DEBUG` section above, otherwise it would make debugging difficult.

roykuper13: ok, done

roykuper13:
> I'm happy to bring this in if someone can fix this so it doesn't report 0 for `make size`. The CI invocation for this is here:
> https://github.com/ARMmbed/littlefs/blob/ce2c01f098f4d2b9479de5a796c3bb531f1fe14c/.travis.yml#L42-L45
> 

alright, fixed

> I do think it's fairly atypical to use littlefs's Makefile for anything other than running tests or examples locally.
> 
> It would be better to copy the Makefile into the root of your project and make project-specific changes outside of littlefs. Otherwise there is the risk changes we make to the Makefile here could break your project's compilation.

I see what you mean, but generally i prefer to not "patch" files of an imported projects. Moreover, since you already exposing a target for static libs, why not offer an option for that?

since I agree with you that we cannot force the user to compile with these flags, I made it optional (with `AR_SECTIONS`) tell me if it makes sense, or do you want me to remove it completely (even though I believe this should be an option)

geky:
Hmm, I'm not sure why but it seems like the CI is still failing to get the code size with LTO
> test/littlefs-arm — Passed, code size is 0B (-100.00%) 

You should be able to see the CI output if you click the "Expand all checks" link in GitHub.

Makefile
```diff
@@ -20,13 +24,24 @@ ifdef DEBUG
...
 ifdef WORD
 override CFLAGS += -m$(WORD)
 endif
 ifdef TRACE
 override CFLAGS += -DLFS_YES_TRACE
 endif
+
+ifeq ($(TARGET), lfs.a)
+	ifdef AR_SECTIONS
+		override CFLAGS += -fdata-sections --function-sections
```
geky, line 41: Does --fdata-sections need two --s?

roykuper13: No. It should be `-fdata-sections` or `-ffunction-sections` according to the man page

roykuper13:
> Hmm, I'm not sure why but it seems like the CI is still failing to get the code size with LTO
> 
> > test/littlefs-arm — Passed, code size is 0B (-100.00%)
> 
> You should be able to see the CI output if you click the "Expand all checks" link in GitHub.

@geky 
alright, now i believe it's fixed :)
`test/littlefs-arm — Passed, code size is 14464B (+0.00%)`

however, the `-flto` addition caused segfault in test_evil (only when SIZE=4*LFS_BLOCK_SIZE and only in mips and ppc) for some reason. I'll try to understand why it happens (and if discarding `-flto` is/isn't needed for target `test`).

roykuper13:
> > Hmm, I'm not sure why but it seems like the CI is still failing to get the code size with LTO
> > > test/littlefs-arm — Passed, code size is 0B (-100.00%)
> > 
> > 
> > You should be able to see the CI output if you click the "Expand all checks" link in GitHub.
> 
> @geky
> alright, now i believe it's fixed :)
> `test/littlefs-arm — Passed, code size is 14464B (+0.00%)`
> 
> however, the `-flto` addition caused segfault in test_evil (only when SIZE=4*LFS_BLOCK_SIZE and only in mips and ppc) for some reason. I'll try to understand why it happens (and if discarding `-flto` is/isn't needed for target `test`).

@geky 
So apparently due to (i guess) lto's optimizations, the corrupted block is never read (for some reason).. As a result, we reach `lfs_rambd_prog` and write `SIZE` bytes from `from_here` file to the `buffer` (and its size is only 1024... and it is on the stack) since `cfg` is also on the stack, we corrupt its memory as well, which then leads to segfault.

Adding `memcpy(&bbuffer[8], &bad, sizeof(bad))` solves the problem, not sure why.

I replaced lto with gc-sections for `lfs` target (in addition to `lfs.a` target).

PR #401, opened 2020-03-24
Lookahead corruption fix given an IO Error during traversal

thrasher8390:
(bugfix) adding line function to clear out all the global 'free' info…rmation so that we can reset it after a failed traversal

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -10,6 +10,8 @@
 #define LFS_BLOCK_NULL ((lfs_block_t)-1)
 #define LFS_BLOCK_INLINE ((lfs_block_t)-2)
 
+static void lfs_alloc_ack(lfs_t *lfs);
+
 /// Caching block device operations ///
 static inline void lfs_cache_drop(lfs_t *lfs, lfs_cache_t *rcache) {
     // do not zero, cheaper if cache is readonly or only going to be
@@ -22,6 +24,15 @@ static inline void lfs_cache_zero(lfs_t *lfs, lfs_cache_t *pcache) {
     // zero to avoid information leak
     memset(pcache->buffer, 0xff, lfs->cfg->cache_size);
     pcache->block = LFS_BLOCK_NULL;
+}
+
+/// Invalidate the lookahead buffer. This is done during mounting and failed traversals ///
+static inline void lfs_setup_invalid_lookahead_buffer(lfs_t *lfs)
+{
+    lfs->free.off = lfs->seed % lfs->cfg->block_size;
+    lfs->free.size = 0;
+    lfs->free.i = 0;
+    lfs_alloc_ack(lfs);
 }
 
 static int lfs_bd_read(lfs_t *lfs,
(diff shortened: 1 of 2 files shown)


thrasher8390:
@geky, here is the PR intended to fix https://github.com/ARMmbed/littlefs/issues/394. Please have a look and give me feedback. I wasn't sure how to fix the testing, also I know you are reworking the testing a bunch in another PR so I didn't want to create a ton of merge conflicts.

lfs.c
```diff
@@ -24,6 +26,15 @@ static inline void lfs_cache_zero(lfs_t *lfs, lfs_cache_t *pcache) {
     pcache->block = LFS_BLOCK_NULL;
 }
 
+/// Invalidate the lookahead buffer. This is done during mounting and failed traversals ///
+static inline void lfs_setup_invalid_lookahead_buffer(lfs_t *lfs)
```
geky: Minor nits:
- Could this be named something like `lfs_alloc_reset` so the alloc functions (lfs_alloc, lfs_alloc_ack) all have a similar prefix?
- Could this be moved next to lfs_alloc_ack, it seems related. If you ordered the function definitions lfs_alloc_ack, lfs_alloc_reset, lfs_alloc, you wouldn't need to predeclare lfs_alloc_ack.
- Also I don't think inline is beneficial here, since it's static the compiler is very likely to inline if it does save code size

lfs.c
```diff
@@ -4467,6 +4476,7 @@ static int lfs1_mount(lfs_t *lfs, struct lfs1 *lfs1,
         lfs->lfs1->root[1] = LFS_BLOCK_NULL;
 
         // setup free lookahead
+        // TODO should this also call lfs_setup_invalid_lookahead_buffer(lfs); the free.off is different in the current version of lfs
```
geky: Yeah all three entry points are a bit uniquely special, that's because we need to trick the allocator into always giving the superblock blocks 0 and 1 (off = 0).

You could replace this with `lfs_setup_invalid_lookahead_buffer` if you init lfs->seed with 0, but that may be too clever. It's likely not worth changing since modifying v1 code carries has more risks than rewards.

lfs.c
thrasher8390: I'm okay with all of these comments! Definitely appreciate consistency :)

lfs.c
thrasher8390: Thanks for the feedback, I'll leave it as is an remove the TODO

thrasher8390:
@geky, I hit some merge conflicts when cherry-picking 1c5bae6. Since you have a better idea of how the tests have progressed I gave you access to thrasher8390/littlefs if you want to make the change to my branch. I've pushed all of my changes and don't plan on pushing anything else unless requested.

Thanks again for your feedback!

geky:
Ok! I have added the test case and retargeted this PR to be against https://github.com/ARMmbed/littlefs/pull/372. Now we'll see if it passes CI.

Then I'll go ahead an merge this into https://github.com/ARMmbed/littlefs/pull/372, and I'm hoping to bring https://github.com/ARMmbed/littlefs/pull/372 into master this week.

Also a quick note you don't actually need to give me permissions to thrasher8390/littlefs. Creating a PR gives me permission to modify the PR's branch unless you disable it explicitly.

That being said I generally avoid modifying other people's PR branches because it can disrupt their work if they're not expecting their branch to change.

geky:
Merged into https://github.com/ARMmbed/littlefs/pull/372, thanks for creating this PR!

geky:
Thanks for this! It should now be on master.

thrasher8390:
Awesome! Thanks @geky.

Issue #394, opened 2020-03-12
LFS Corruption - Invalid Look Ahead Buffer (super block directory pointer and CTZ pointer conflict)


thrasher8390:
@geky,
**This is more or less a Cry for Help.** I have experienced my file system being reformatted due to -84 `LFS_ERR_CORRUPT` and I've been trying to understand what is going wrong for the last week.

Our file system is running on 511 blocks of size 4096. I have our block wear-leveling set to **500** and the CACHE_SIZE is 1024. Our max reads can be 1024 but our maximum program size is 256.

I'd like to start by learning a bit about the super block (This is not the piece that is causing the corruption on mount). I'll post 2 images for our block 0 and block 1. What I see is that block 0 has a revision of 0x0B but seems to be "corrupted". Block 1 has a revision of 0x0C and has all of the data that I would expect. 
**Do you know why block 0 looks the way it does?**

Block 0 **looks corrupted**
![image](https://user-images.githubusercontent.com/2731241/76533010-cc3c8880-644d-11ea-9988-aa5f64a23e94.png)
Block 1 **seems to be what I'd expect. Circled red is the directory that failed to mount**
![image](https://user-images.githubusercontent.com/2731241/76533302-39501e00-644e-11ea-90c4-e5cd9da725b3.png)

**Getting into the corruption piece**
I can see that the superblock 1 said that directory "db" lives at 0xC2 / 0xC3. It turns out that when I look at those 2 blocks, I see the left overs of a CTZ file that is frequently written (I know that blocks are erased before being written to, therefore it is expected that a CTZ file would be littered through external flash). ~I are currently investigating whether or not the file system ever placed the CTZ file in blocks 0xC2/ 0xC3.~ Using debug.py  I looked at the directory of the CTZ file in question and have confirmed that the last 2 valid points do indeed point to 0xC2 and 0xC3. Therefore the file system in its current state believe that the CTZ used to live at 0xC2 and currently lives at 0xC3 and the superblock believes that our db directory metadata pair lives at 0xC2 / 0xC3. Seeming to be more like an LFS issue...
![image](https://user-images.githubusercontent.com/2731241/76536668-09574980-6453-11ea-9c7c-22e7c274257d.png)

**HERE IS THE GOTCHA**
[...]

PR #403, opened 2020-03-24
(improvement) adding contribution file

thrasher8390:


thrasher8390:
@geky, @pjsg, I created a new PR that didn't include the bug fix from issue #394

geky:
Thanks for the PR and getting this started. This will be useful to know what questions potential contributors ask, I'm biased after using GitHub for so long.

I'll have to think about what this should look like.

PR #405, opened 2020-04-03
Fix -Wmissing-field-initializers

rojer:


geky:
Hi @rojer, thanks for creating a PR. Did this change in v2.2?

I'm happy to bring this in, though the flag should be probably be added to the Makefile to prevent another regression. It looks like it was omitted because of a compiler bug (https://github.com/ARMmbed/littlefs/commit/7d8f8ced03ea7b0afd7f26c340fc1c2474360ab5), but it would be interesting to see if the fixed GCC has been adopted in Travis yet.

rojer:
> Did this change in v2.2?
yes, i think so

> see if the fixed GCC has been adopted in Travis yet.
added a commit that removes the flag, let's see if it passes

rojer:
looks like it's ok

geky:
Time to merge. Thanks for this!

Issue #297, opened 2019-09-29
Python wrapper for littlefs: Permisson to use littlefs in name


jrast:
I'm currently working on a thin python wrapper for littlefs. The goal is to provide a simple, platform independent library which does not require any compilation steps on the system of the enduser. Therefore playing around with littlefs, testing if it would be a good fit or preparing/read/change flash memories on a computer would be a lot easier.

The current state of the implementation can be found here:
https://github.com/jrast/littlefs-python

I'm planing to release the library under the name "littlefs-python" and wanted to ask for permission to use this name before a initial release. The library will use the same BSD-3 Clause licence as littlefs itself. Is this ok or should I change the name?

geky:
That looks like a great project, thanks for sharing! Feel free to use the name, it would be more confusing without it : )

jrast:
Thanks! Hopefully the package will be of use for some people!

PR #412, opened 2020-04-13
Added littlefs-python to the related projects section

jrast:
As introduced in #297, I created a python wrapper for littlefs. The wrapper supports two API's: A C-like API which is the same as in C and a more pythonic API which is easier to use if you are more the python guy. The wrapper is built with littlefs 2.2.1 at the moment.


geky:
Hi @jrast, sorry about leaving this PR hanging for so long. I haven't been able to make a release for a while, but will bring this in on the next patch release.

PR #424, opened 2020-05-09
lfs.c: Replace strlen() with strnlen()

abderraouf-adjal:
**lfs.c: Replace strlen() with strnlen():**  
This modification is a security and performance optimization  
to make string length counting stop when we rich the limits.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -1925,7 +1925,7 @@ int lfs_mkdir(lfs_t *lfs, const char *path) {
     }
 
     // check that name fits
-    lfs_size_t nlen = strlen(path);
+    lfs_size_t nlen = strnlen(path, 1+(size_t)lfs->name_max);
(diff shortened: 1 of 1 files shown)


abderraouf-adjal:
Hmm, I see.  
This works with `-std=gnu99` but not `-std=c99`, what to do?

e107steved:
If not available as a function, strnlen() can be implemented as a macro something like min(strlen(), maxlen)
Must say I've not felt the need for this particular function myself, although I do use strlcpy() and similar in preference to strcpy() or strncpy() - although that group are also specific to gcc and originated elsewhere (BSD?)

abderraouf-adjal:
@e107steved a `min()` function would be better than a macro (in-case of an `len++` input for ex.), my point was to not access a memory outside a buffer, a `min()` would work but it still do an access.

Issue #442, opened 2020-06-17
API to check if file is currently open


keck-in-space:
I have a need to check if a file is currently open. Is it OK to do so as indicated below?

```c
lfs_file_t* file;
bool isOpen = ((file->flags & LFS_F_OPENED) == LFS_F_OPENED);
```

This is a useful check to be made in cases where files are written autonomously and could fail to close before the next attempt to open is made.

LFS defines the `LFS_F_OPENED` flag as internally used, so I wanted to check if this access was safe, or if there are some conditions when the flag isn't immediately updated.

geky:
Oh interesting. Could you not use the return value from `lfs_file_close` to set a flag in your application logic?

The reason LFS_F_OPENED is noted as a internal only is to keep us free to change its behavior in any release. It's often hard enough to fix things while maintaining backwards compatibility with the public API.

keck-in-space:
Yes, that is one way to handle this. I wanted to avoid tracking it externally if possible, since it would be another variable to keep track of, and we'd need to write to it every time an opencfg or close executed (successfully or otherwise). We're locked in on LFS 2.1.4, if that matters.

geky:
> This is a useful check to be made in cases where files are written autonomously and could fail to close before the next attempt to open is made.

Actually, thinking about it, are you sure this flag gives you this information?

lfs_file_close is a bit strange in that it does two things:
- Syncs data to disk.
- Cleans up resources.

The resource clean up can not fail. If an error occurs on write, the pending data is discarded, but the local resources (file cache, anything allocated) are deallocated and lost. This way common code that does open->write->close does not memory leak if a filesystem error occurs (such as a full disk).

If you need to take recovery action and not lose the data, lfs_file_sync syncs data to disk without cleaning up resources. An lfs_file_sync followed by lfs_file_close is guaranteed to do no extra disk operations.

---

So if an error happens, this should change the state of LFS_F_OPENED. Instead LFS_F_ERRED gets set, which indicates if a write error occured. If it's set, all future writes on the file are discarded.

LFS_F_OPENED is really only used as a debug assertion to help prevent some programming mistakes such as opening a file twice or writing after close. It also may be replaced by https://github.com/ARMmbed/littlefs/pull/443 which looks like it has the same affect without needing an extra bit.

keck-in-space:
[...]

PR #443, opened 2020-06-17
Assert that the file isnt open in lfs_file_opencfg

### LINKED ISSUE #429: infinite loop during file close
if lfs_file_open is called two times (with identical arguments) it will return ok to both calls. That will create self referencing linked list (lfs->mlist->next will look to itself). During file closing (in function lfs_dir_commit) linked list will create infinite loop and never return.

I'm aware that described usage is not valid but maybe you could add assertion or return error result on second call. I discovered that behavior when I've tested lfs_file_opencfg bug forgot to remove previous lfs_file_open.

NoahGorny:
Closes #429

Also fixed a magic number in `opencfg`

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -2403,7 +2403,7 @@ int lfs_file_opencfg(lfs_t *lfs, lfs_file_t *file,
     LFS_ASSERT(!lfs_mlist_isopen(lfs->mlist, (struct lfs_mlist*)file));
 
     // deorphan if we haven't yet, needed at most once after poweron
-    if ((flags & 3) != LFS_O_RDONLY) {
+    if ((flags & LFS_O_RDWR) != LFS_O_RDONLY) {
         int err = lfs_fs_forceconsistency(lfs);
         if (err) {
             LFS_TRACE("lfs_file_opencfg -> %d", err);
@@ -2478,7 +2478,7 @@ int lfs_file_opencfg(lfs_t *lfs, lfs_file_t *file,
 
     // fetch attrs
     for (unsigned i = 0; i < file->cfg->attr_count; i++) {
-        if ((file->flags & 3) != LFS_O_WRONLY) {
+        if ((file->flags & LFS_O_RDWR) != LFS_O_WRONLY) {
             lfs_stag_t res = lfs_dir_get(lfs, &file->m,
                     LFS_MKTAG(0x7ff, 0x3ff, 0),
                     LFS_MKTAG(LFS_TYPE_USERATTR + file->cfg->attrs[i].type,
@@ -2490,7 +2490,7 @@ int lfs_file_opencfg(lfs_t *lfs, lfs_file_t *file,
             }
         }
 
-        if ((file->flags & 3) != LFS_O_RDONLY) {
+        if ((file->flags & LFS_O_RDWR) != LFS_O_RDONLY) {
             if (file->cfg->attrs[i].size > lfs->attr_max) {
                 err = LFS_ERR_NOSPC;
                 goto cleanup;
@@ -2807,7 +2807,7 @@ lfs_ssize_t lfs_file_read(lfs_t *lfs, lfs_file_t *file,
     LFS_TRACE("lfs_file_read(%p, %p, %p, %"PRIu32")",
             (void*)lfs, (void*)file, buffer, size);
     LFS_ASSERT(file->flags & LFS_F_OPENED);
-    LFS_ASSERT((file->flags & 3) != LFS_O_WRONLY);
+    LFS_ASSERT((file->flags & LFS_O_RDWR) != LFS_O_WRONLY);
 
     uint8_t *data = buffer;
     lfs_size_t nsize = size;
@@ -2887,7 +2887,7 @@ lfs_ssize_t lfs_file_write(lfs_t *lfs, lfs_file_t *file,
     LFS_TRACE("lfs_file_write(%p, %p, %p, %"PRIu32")",
             (void*)lfs, (void*)file, buffer, size);
     LFS_ASSERT(file->flags & LFS_F_OPENED);
-    LFS_ASSERT((file->flags & 3) != LFS_O_RDONLY);
+    LFS_ASSERT((file->flags & LFS_O_RDWR) != LFS_O_RDONLY);
 
     const uint8_t *data = buffer;
     lfs_size_t nsize = size;
@@ -3052,7 +3052,7 @@ int lfs_file_truncate(lfs_t *lfs, lfs_file_t *file, lfs_off_t size) {
     LFS_TRACE("lfs_file_truncate(%p, %p, %"PRIu32")",
             (void*)lfs, (void*)file, size);
     LFS_ASSERT(file->flags & LFS_F_OPENED);
-    LFS_ASSERT((file->flags & 3) != LFS_O_RDONLY);
+    LFS_ASSERT((file->flags & LFS_O_RDWR) != LFS_O_RDONLY);
 
     if (size > LFS_FILE_MAX) {
         LFS_TRACE("lfs_file_truncate -> %d", LFS_ERR_INVAL);
--- a/lfs.c
+++ b/lfs.c
@@ -411,6 +411,17 @@ static inline void lfs_superblock_tole32(lfs_superblock_t *superblock) {
     superblock->attr_max    = lfs_tole32(superblock->attr_max);
 }
 
+static inline bool lfs_mlist_isopen(struct lfs_mlist *head,
+        struct lfs_mlist *node) {
+    for (struct lfs_mlist **p = &head; *p; p = &(*p)->next) {
+        if (*p == (struct lfs_mlist*)node) {
+            return true;
+        }
+    }
+
+    return false;
+}
+
 
 /// Internal operations predeclared here ///
 static int lfs_dir_commit(lfs_t *lfs, lfs_mdir_t *dir,
@@ -2007,6 +2018,8 @@ int lfs_mkdir(lfs_t *lfs, const char *path) {
 
 int lfs_dir_open(lfs_t *lfs, lfs_dir_t *dir, const char *path) {
     LFS_TRACE("lfs_dir_open(%p, %p, \"%s\")", (void*)lfs, (void*)dir, path);
+    LFS_ASSERT(!lfs_mlist_isopen(lfs->mlist, (struct lfs_mlist*)dir));
+
     lfs_stag_t tag = lfs_dir_find(lfs, &dir->m, &path, NULL);
     if (tag < 0) {
         LFS_TRACE("lfs_dir_open -> %"PRId32, tag);
@@ -2387,6 +2400,7 @@ int lfs_file_opencfg(lfs_t *lfs, lfs_file_t *file,
                  ".buffer=%p, .attrs=%p, .attr_count=%"PRIu32"})",
             (void*)lfs, (void*)file, path, flags,
             (void*)cfg, cfg->buffer, (void*)cfg->attrs, cfg->attr_count);
+    LFS_ASSERT(!lfs_mlist_isopen(lfs->mlist, (struct lfs_mlist*)file));
 
     // deorphan if we haven't yet, needed at most once after poweron
     if ((flags & 3) != LFS_O_RDONLY) {
--- a/lfs.c
+++ b/lfs.c
@@ -422,6 +422,15 @@ static inline bool lfs_mlist_isopen(struct lfs_mlist *head,
     return false;
 }
 
+static inline void lfs_mlist_remove(lfs_t *lfs, struct lfs_mlist *mlist) {
+    for (struct lfs_mlist **p = &lfs->mlist; *p; p = &(*p)->next) {
+        if (*p == mlist) {
+            *p = (*p)->next;
+            break;
+        }
+    }
+}
+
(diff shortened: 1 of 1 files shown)


e107steved:
I think I'd prefer the operation failed by returning an error, rather than doing an assert

NoahGorny:
> I think I'd prefer the operation failed by returning an error, rather than doing an assert

I think that checking every time for something that is really a user error is not needed, but I will accept this if more people prefer checks. If your program hangs in littlefs I expect that you will compile littlefs in debug and then you will see the problem.

In a different matter, I really need to find a way to check if the struct was initialized, because right now I access uninitialized memory :(

judgeh:
I definitely prefer the return code. I think asserts are heavily overused in lfs. there are many cases where my application would benefit from handling return codes rather than the asserting.

lfs.c
```diff
@@ -2387,9 +2398,10 @@ int lfs_file_opencfg(lfs_t *lfs, lfs_file_t *file,
                  ".buffer=%p, .attrs=%p, .attr_count=%"PRIu32"})",
             (void*)lfs, (void*)file, path, flags,
             (void*)cfg, cfg->buffer, (void*)cfg->attrs, cfg->attr_count);
+    LFS_ASSERT(is_node_not_in_list(lfs->mlist, (struct lfs_mlist*)file));
```
diederikvdv, line 2401: This check can also be added to the lfs_dir_open function:
LFS_ASSERT(is_node_not_in_list(lfs->mlist, (struct lfs_mlist*)dir));

This requires the static function to be declared earlier. Other static functions are declared at the top of the file

NoahGorny: added, thanks for the feedback!

NoahGorny:
Guys, I am not sure why you should return error on such cases. Iterating over all open files in each open call seems costly, so I think we can stick to an assert. You can open the same file twice, just not with the same struct...

judgeh:
> You can open the same file twice, just not with the same struct...

Oh in our testing, we were unable to open a file twice with different structures.
we
1. opened a file with hdl1
2. wrote data using hdl1
3. synced hdl1 (but not close) 
4. opened same file with hdl2
5. read hd2 (this always reported zero bytes) 

Do you expect, in this scenario, the hdl2 read to return the bytes written by hdl1 ? Maybe our test case was buggy, but we concluded that we could not read a file that was already open. If this is not the case, then thats great.

judgeh:
Also, isnt it still checking every open file in the assert? How is asserting any different than returning in terms on complexity ?

NoahGorny:
I will check the test case you mentioned, thanks!
I currently only check that the struct is not reused, in order to test that the file is not already opened, I will need to add a new assert.

And assert is better because it is not complied in, unless on debug

geky:
Hi @NoahGorny, thanks for creating a PR! Sorry to leave this hanging for so long. This is a good addition.

I wonder if it would also be good to replace LFS_F_OPENED (https://github.com/littlefs-project/littlefs/pull/237) so we have the same check for open/close files that doesn't get tripped up by uninitialized memory. Thoughts?

---

My opinion is this check should assert. For a code size reason, not a usability reason. Asserts can be removed during compilation, return codes can not. And for microcontroller-scale programs, you really don't want to pay for code that should never run.

Also the current behavior for these types of user checks is to assert, if we want to change this behavior it should be done via a different issue or PR.

lfs.c
```diff
@@ -2387,9 +2400,10 @@ int lfs_file_opencfg(lfs_t *lfs, lfs_file_t *file,
...
 
     // deorphan if we haven't yet, needed at most once after poweron
-    if ((flags & 3) != LFS_O_RDONLY) {
+    if ((flags & LFS_O_RDWR) != LFS_O_RDONLY) {
```
geky, line 2406: 👍

lfs.c
```diff
@@ -411,6 +411,17 @@ static inline void lfs_superblock_tole32(lfs_superblock_t *superblock) {
     superblock->attr_max    = lfs_tole32(superblock->attr_max);
 }
 
+static inline bool is_node_not_in_list(struct lfs_mlist *head,
+    struct lfs_mlist *node) {
```
geky, line 415: style nit: Two indents when inside parens (I realize I don't have this documented anywhere...)

lfs.c
```diff
@@ -411,6 +411,17 @@ static inline void lfs_superblock_tole32(lfs_superblock_t *superblock) {
     superblock->attr_max    = lfs_tole32(superblock->attr_max);
 }
 
+static inline bool is_node_not_in_list(struct lfs_mlist *head,
```
geky, line 414: nit: Prefix with `lfs_`. Thoughts on `lfs_mlist_isopen` for consistency with the other functions?

diederikvdv:
> I wonder if it would also be good to replace LFS_F_OPENED (#237) so we have the same check for open/close files that doesn't get tripped up by uninitialized memory. Thoughts?

If LFS_F_OPENED (#237) is replaced, it might also be a good addition to add the isOpened API mentioned in #442. But then based on lfs_is_node_not_in_list instead of LFS_F_OPENED

lfs.c
NoahGorny: I changed in all places now :smile:

lfs.c
NoahGorny: done

lfs.c
NoahGorny: done, changed to `lfs_mlist_isused`

NoahGorny:
> Hi @NoahGorny, thanks for creating a PR! Sorry to leave this hanging for so long. This is a good addition.
> 
> I wonder if it would also be good to replace LFS_F_OPENED (#237) so we have the same check for open/close files that doesn't get tripped up by uninitialized memory. Thoughts?

I agree, and added commits that deprecate LFS_F_OPENED. It required me to sometimes add files to the mlist when we did not do so before, so you might wanna take a look at 0250015 and 0ca3012 to tell me whether my solution is okay or not.  
In any case, I did not add support for asserting of file that was opened twice using a different `file` struct, so we might want to take a look at it. Reusing an existing struct will now raise assert errors :smile:  
By the way, are there any ways to "catch' asserts in the tests? I want to deliberately see the asserts trigger in the tests.

Thanks alot for your feedback btw @geky

NoahGorny:
> > I wonder if it would also be good to replace LFS_F_OPENED (#237) so we have the same check for open/close files that doesn't get tripped up by uninitialized memory. Thoughts?
> 
> If LFS_F_OPENED (#237) is replaced, it might also be a good addition to add the isOpened API mentioned in #442. But then based on lfs_is_node_not_in_list instead of LFS_F_OPENED

I think @geky should voice his opinion here. I think that currently with the assert not fully catching all cases (opened with different struct) it should not be public API yet... maybe after we catch/handle all such cases

NoahGorny:
> > You can open the same file twice, just not with the same struct...
> 
> Oh in our testing, we were unable to open and read a file twice with different structures.
> we
> 
>     1. opened a file with hdl1
> 
>     2. wrote data using hdl1
> 
>     3. synced hdl1 (but not close)
> 
>     4. opened same file with hdl2
> 
>     5. read hd2 (this always reported zero bytes)
> 
> 
> Do you expect, in this scenario, the hdl2 read to return the bytes written by hdl1 ? Maybe our test case was buggy, but we concluded that we could not read a file that was already open. If this is not the case, then thats great.

@judgeh It indeed fails with my tests, I created a test (on a separate PR) that fails on this (ad995ceb9bf72f5c87690e2c63f70ed585fabb29)  
I am not sure what is the correct approach to this, but I think @geky can have some insights :smile:

geky:
> Oh in our testing, we were unable to open and read a file twice with different structures.
> > @judgeh It indeed fails with my tests, I created a test (on a separate PR) that fails on this (ad995ce)

Ah, this looks like the same issue being discussed in https://github.com/littlefs-project/littlefs/issues/483. Thumbing up https://github.com/littlefs-project/littlefs/issues/483 would help know who's ran into this. Also thanks for the test case!

TLDR: Currently, littlefs is transactional in the sense that each open file handle gets its own snapshot. Each file handle can be read/written independently (and independently from disk). Something that came up in https://github.com/littlefs-project/littlefs/issues/483, we could go through all open files and throw out other changes on sync/close, it would just take more code size. More info in https://github.com/littlefs-project/littlefs/issues/483.

lfs.c
```diff
@@ -411,6 +411,22 @@ static inline void lfs_superblock_tole32(lfs_superblock_t *superblock) {
...
+    for (struct lfs_mlist **p = &head; *p; p = &(*p)->next) {
+        if (*p == (struct lfs_mlist*)node) {
+            return true;
+        }
+    }
+
+    return false;
+}
+
+static inline void lfs_mlist_add(lfs_t *lfs, struct lfs_mlist *mlist) {
```
geky, line 425: If we add this, `lfs_mlist_remove` would be more valuable:
[https://github.com/littlefs-project/littlefs/blob/4c9146ea539f72749d6cc3ea076372a81b12cb11/lfs.c#L2553-L2559 ](https://github.com/littlefs-project/littlefs/blob/4c9146ea539f72749d6cc3ea076372a81b12cb11/lfs.c#L2553-L2559)

Also a nit: I think `lfs_mlist_append` captures the data structure operation better.

lfs.h
```diff
@@ -136,7 +136,7 @@ enum lfs_open_flags {
     LFS_F_READING = 0x040000, // File has been read since last flush
     LFS_F_ERRED   = 0x080000, // An error occured during write
     LFS_F_INLINE  = 0x100000, // Currently inlined in directory entry
-    LFS_F_OPENED  = 0x200000, // File has been opened
+    LFS_F_OPENED  = 0x200000, // DEPRECATED: File has been opened
```
geky, line 139: Since these are marked as "internally used flags", I think it's safe to remove without deprecation.

lfs.c
```diff
@@ -411,6 +411,22 @@ static inline void lfs_superblock_tole32(lfs_superblock_t *superblock) {
     superblock->attr_max    = lfs_tole32(superblock->attr_max);
 }
 
+static inline bool lfs_mlist_isused(struct lfs_mlist *head,
```
geky, line 414: nit: I think this would be better as `lfs_mlist_isopen`, it is directly related to `lfs_file_open` and `lfs_dir_open`.

lfs.c
geky: Ah I misread this. Well, I don't really have an opinion on this change. It is a curious mask.

lfs.c
```diff
@@ -2400,23 +2418,22 @@ int lfs_file_opencfg(lfs_t *lfs, lfs_file_t *file,
...
-    file->flags = flags | LFS_F_OPENED;
     file->pos = 0;
     file->off = 0;
+    file->flags = flags;
     file->cache.buffer = NULL;
 
+    // get id, add to list of mdirs to catch update changes
+    file->type = LFS_TYPE_REG;
+    lfs_mlist_add(lfs, (struct lfs_mlist *)file);
+
```
geky, line 2429: I wouldn't move this either, the `lfs_mdir_t` is not initialized until the `lfs_dir_find`, and I'm not sure having an uninitialized mdir on in mlist is safe.

Unfortunately the mlist is a bit sensitive because all wear-leveling/allocation logic depend on it. It would be better if we prefer changing the LFS_F_OPENED behavior slightly to avoid changing the mlist logic, since LFS_F_OPENED asserts are just protection against already invalid API usage.

geky:
> it might also be a good addition to add the isOpened API mentioned in #442

I'm still not convinced on https://github.com/littlefs-project/littlefs/issues/442. The behavior of an `lfs_file_isopened` could be accomplished with an external flag without requiring an additional API into the filesystem. It could also be added later if there is interest.

geky:
> By the way, are there any ways to "catch' asserts in the tests?

Almost missed this, unfortunately no, the test script currently only treats asserts as failures. The test script should also be documented better...

lfs.c
NoahGorny: I think this is better than a magic num

lfs.c
NoahGorny: done both!

lfs.h
NoahGorny: removed the change entirely

lfs.c
NoahGorny: its okay, changed it

lfs.c
NoahGorny: its honestly better to leave alone the LFS_F_OPENED flag honestly. It causes those problems with the assert and the `isused` check can be utilized only when need (in `file_opencfg` and `dir_open`) and the rest of the API can stay with the flag

lfs.c
geky: It looks like this is still `lfs_mlist_isused`, or am I missing a commit?

lfs.c
geky: Can't argue with that

lfs.c
NoahGorny: sorry, updated it now :smile:

geky (approved):
This looks great, thanks for this! Will merge into the next release.

geky:
I didn't notice until after merging that this reverted the `LFS_F_OPENED` flag, so I went ahead and re-added that change to #495. But can re-revert if needed.

I think I ran into some of the issues you did with the change, but one thing that helped was the new separate API-layer, I could move the check there and it solved some of complex internal situations, as well as reduced the calls to `lfs_mlist_isopen` to the API layer.

PR #459, opened 2020-08-18
Fix assert

sgupta20:
Fix for a minor annoyance where the testbd device uses `assert` rather than `LFS_ASSERT`


geky:
Good catch, thanks for this! Will merge into the next patch release.

geky:
Retriggering for CI

geky:
Of course that broke it, reopening this PR here: https://github.com/littlefs-project/littlefs/pull/486. Sorry for the noise.

PR #470, opened 2020-09-17
Add thread safe wrappers

renesas-billgesner:
I'm working on an optional thread safe wrapper for the top level LittleFS APIs. I'm currently trying to sort out a few outstanding issues but the general design is there. When LFS_THREAD_SAFE is enabled all lfs calls will be routed through the locking wrappers otherwise all that code will be compiled out and function as normal.

We are testing using AWS IoT qualification tests but have not added any tests to the test suite.

Please let me know your thoughts.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -1525,7 +1525,7 @@ static int lfs_dir_compact(lfs_t *lfs,
         if (lfs_pair_cmp(dir->pair, (const lfs_block_t[2]){0, 1}) == 0) {
             // oh no! we're writing too much to the superblock,
             // should we expand?
-            lfs_ssize_t res = lfs_fs_size(lfs);
+            lfs_ssize_t res = _lfs_fs_size(lfs);
             if (res < 0) {
                 return res;
             }
@@ -1906,7 +1906,7 @@ compact:
 
 
 /// Top level directory operations ///
-int lfs_mkdir(lfs_t *lfs, const char *path) {
+int _lfs_mkdir(lfs_t *lfs, const char *path) {
     LFS_TRACE("lfs_mkdir(%p, \"%s\")", (void*)lfs, path);
     // deorphan if we haven't yet, needed at most once after poweron
     int err = lfs_fs_forceconsistency(lfs);
@@ -2005,7 +2005,7 @@ int lfs_mkdir(lfs_t *lfs, const char *path) {
     return 0;
 }
 
-int lfs_dir_open(lfs_t *lfs, lfs_dir_t *dir, const char *path) {
+int _lfs_dir_open(lfs_t *lfs, lfs_dir_t *dir, const char *path) {
     LFS_TRACE("lfs_dir_open(%p, %p, \"%s\")", (void*)lfs, (void*)dir, path);
     lfs_stag_t tag = lfs_dir_find(lfs, &dir->m, &path, NULL);
     if (tag < 0) {
@@ -2056,7 +2056,7 @@ int lfs_dir_open(lfs_t *lfs, lfs_dir_t *dir, const char *path) {
     return 0;
 }
 
-int lfs_dir_close(lfs_t *lfs, lfs_dir_t *dir) {
+int _lfs_dir_close(lfs_t *lfs, lfs_dir_t *dir) {
     LFS_TRACE("lfs_dir_close(%p, %p)", (void*)lfs, (void*)dir);
     // remove from list of mdirs
     for (struct lfs_mlist **p = &lfs->mlist; *p; p = &(*p)->next) {
@@ -2070,7 +2070,7 @@ int lfs_dir_close(lfs_t *lfs, lfs_dir_t *dir) {
     return 0;
 }
 
-int lfs_dir_read(lfs_t *lfs, lfs_dir_t *dir, struct lfs_info *info) {
+int _lfs_dir_read(lfs_t *lfs, lfs_dir_t *dir, struct lfs_info *info) {
     LFS_TRACE("lfs_dir_read(%p, %p, %p)",
             (void*)lfs, (void*)dir, (void*)info);
     memset(info, 0, sizeof(*info));
@@ -2123,11 +2123,11 @@ int lfs_dir_read(lfs_t *lfs, lfs_dir_t *dir, struct lfs_info *info) {
     return true;
 }
 
-int lfs_dir_seek(lfs_t *lfs, lfs_dir_t *dir, lfs_off_t off) {
+int _lfs_dir_seek(lfs_t *lfs, lfs_dir_t *dir, lfs_off_t off) {
     LFS_TRACE("lfs_dir_seek(%p, %p, %"PRIu32")",
             (void*)lfs, (void*)dir, off);
     // simply walk from head dir
-    int err = lfs_dir_rewind(lfs, dir);
+    int err = _lfs_dir_rewind(lfs, dir);
     if (err) {
         LFS_TRACE("lfs_dir_seek -> %d", err);
         return err;
@@ -2166,14 +2166,14 @@ int lfs_dir_seek(lfs_t *lfs, lfs_dir_t *dir, lfs_off_t off) {
     return 0;
 }
 
-lfs_soff_t lfs_dir_tell(lfs_t *lfs, lfs_dir_t *dir) {
+lfs_soff_t _lfs_dir_tell(lfs_t *lfs, lfs_dir_t *dir) {
     LFS_TRACE("lfs_dir_tell(%p, %p)", (void*)lfs, (void*)dir);
     (void)lfs;
     LFS_TRACE("lfs_dir_tell -> %"PRId32, dir->pos);
     return dir->pos;
 }
 
-int lfs_dir_rewind(lfs_t *lfs, lfs_dir_t *dir) {
+int _lfs_dir_rewind(lfs_t *lfs, lfs_dir_t *dir) {
     LFS_TRACE("lfs_dir_rewind(%p, %p)", (void*)lfs, (void*)dir);
     // reload the head dir
     int err = lfs_dir_fetch(lfs, &dir->m, dir->head);
@@ -2380,7 +2380,7 @@ static int lfs_ctz_traverse(lfs_t *lfs,
 
 
 /// Top level file operations ///
-int lfs_file_opencfg(lfs_t *lfs, lfs_file_t *file,
+int _lfs_file_opencfg(lfs_t *lfs, lfs_file_t *file,
         const char *path, int flags,
         const struct lfs_file_config *cfg) {
     LFS_TRACE("lfs_file_opencfg(%p, %p, \"%s\", %x, %p {"
@@ -2529,26 +2529,26 @@ int lfs_file_opencfg(lfs_t *lfs, lfs_file_t *file,
 cleanup:
     // clean up lingering resources
     file->flags |= LFS_F_ERRED;
-    lfs_file_close(lfs, file);
+    _lfs_file_close(lfs, file);
     LFS_TRACE("lfs_file_opencfg -> %d", err);
     return err;
 }
 
-int lfs_file_open(lfs_t *lfs, lfs_file_t *file,
+int _lfs_file_open(lfs_t *lfs, lfs_file_t *file,
         const char *path, int flags) {
     LFS_TRACE("lfs_file_open(%p, %p, \"%s\", %x)",
             (void*)lfs, (void*)file, path, flags);
     static const struct lfs_file_config defaults = {0};
-    int err = lfs_file_opencfg(lfs, file, path, flags, &defaults);
+    int err = _lfs_file_opencfg(lfs, file, path, flags, &defaults);
     LFS_TRACE("lfs_file_open -> %d", err);
     return err;
 }
 
-int lfs_file_close(lfs_t *lfs, lfs_file_t *file) {
+int _lfs_file_close(lfs_t *lfs, lfs_file_t *file) {
     LFS_TRACE("lfs_file_close(%p, %p)", (void*)lfs, (void*)file);
     LFS_ASSERT(file->flags & LFS_F_OPENED);
 
-    int err = lfs_file_sync(lfs, file);
+    int err = _lfs_file_sync(lfs, file);
 
     // remove from list of mdirs
     for (struct lfs_mlist **p = &lfs->mlist; *p; p = &(*p)->next) {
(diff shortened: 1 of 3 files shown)


davidefer:
Hi it seems ok to me. Thank you.

geky:
Hi @renesas-billgesner, thanks for creating a PR! Sorry about the late response.

This looks promising and would be valuable to bring in (previous discussion https://github.com/ARMmbed/littlefs/pull/317).

This would also be a good target for moving the tracing code into, since it clutters up the internal logic as is. I can make this change later though so I don't clog this PR.

lfs.c
```diff
@@ -1525,7 +1525,7 @@ static int lfs_dir_compact(lfs_t *lfs,
         if (lfs_pair_cmp(dir->pair, (const lfs_block_t[2]){0, 1}) == 0) {
             // oh no! we're writing too much to the superblock,
             // should we expand?
-            lfs_ssize_t res = lfs_fs_size(lfs);
+            lfs_ssize_t res = _lfs_fs_size(lfs);
```
geky, line 1528: I was thinking a good name for these internal functions may be "lfs_fs_size_raw", "lfs_dir_read_raw", etc. Thoughts?

lfs.c
```diff
@@ -4911,3 +4911,124 @@ int lfs_migrate(lfs_t *lfs, const struct lfs_config *cfg) {
 }
 
 #endif
+
+#if LFS_THREAD_SAFE
+
+#define CREATE_LFS_TS_1(ret, function, b_type, b) \
+    ret _ts ## function(b_type b)                 \
+    {                                             \
+        int err = lfs->cfg->lock(lfs->cfg);       \
```
geky, line 4920: This is a good way to implement this. A future feature I want to look at is optionally collapsing configuration at compile time, but that will be something for the future.

lfs.c
```diff
@@ -4911,3 +4911,124 @@ int lfs_migrate(lfs_t *lfs, const struct lfs_config *cfg) {
 }
 
 #endif
+
+#if LFS_THREAD_SAFE
+
+#define CREATE_LFS_TS_1(ret, function, b_type, b) \
```
geky, line 4917: I would prefer these to not be macro wrapped, I understand the motivation but they hurt debugability. It's tedius but I think it would be best to write these out individually.

I can expand these if you're ok with that. Vim macros makes it relatively painless.

lfs.h
```diff
@@ -84,6 +85,9 @@ enum lfs_error {
     LFS_ERR_NOMEM       = -12,  // No more memory available
     LFS_ERR_NOATTR      = -61,  // No data/attr available
     LFS_ERR_NAMETOOLONG = -36,  // File name too long
+#if LFS_THREAD_SAFE
+    LFS_ERR_LOCK        = -23,  // Failed to aquire lock
```
geky, line 89: Is this needed? littlefs's current strategy is to let any negative error codes go through (for example the block device callbacks). This lets the OS return more specific errors such as EBUSY or EDEADLK (looking at https://linux.die.net/man/3/pthread_mutex_lock).

lfs_util.h
```diff
@@ -43,6 +43,9 @@ extern "C"
 {
 #endif
 
+#ifndef LFS_THREAD_SAFE
```
geky, line 46: A simple comment here such as "Enables thread-safe wrappers using the lock/unlock callbacks in lfs_config" could go a long way

lfs.c
renesas-billgesner: I have them expanded in a previous commit. I'll grab those changes and update.

lfs.h
```diff
@@ -9,6 +9,7 @@
 
 #include <stdint.h>
 #include <stdbool.h>
+#include "lfs_util.h"
```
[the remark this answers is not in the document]
renesas-billgesner: This wasn't actually the issue, please disregard.

geky:
Hi @renesas-billgesner, this is looking pretty good. Do you think it is ready to bring in? Sorry about not being able to get to this for a while.

I have a handful of nits/style requests, but can change those during merge if you're not able to change them.

lfs.c
```diff
@@ -5,7 +5,7 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 #include "lfs.h"
-#include "lfs_util.h"
+#include <lfs_util.h>
```
geky, line 8: Do these need to be angle-brackets?

lfs.h
```diff
@@ -9,6 +9,7 @@
 
 #include <stdint.h>
 #include <stdbool.h>
+#include <lfs_util.h>
```
geky, line 12: Oh I was surprised this wasn't already included. Does this need to use angle-brackets?

lfs.h
```diff
@@ -53,7 +54,7 @@ typedef uint32_t lfs_block_t;
 
 // Maximum size of a file in bytes, may be redefined to limit to support other
 // drivers. Limited on disk to <= 4294967296. However, above 2147483647 the
-// functions lfs_file_seek, lfs_file_size, and lfs_file_tell will return
+// functions lfs_file_seek_raw, lfs_file_size_raw, and lfs_file_tell_raw will return
```
geky, line 57: Looks like documentation got caught in a sed

lfs.h
```diff
@@ -174,6 +175,16 @@ struct lfs_config {
     // are propogated to the user.
     int (*sync)(const struct lfs_config *c);
 
+    #if LFS_THREAD_SAFE
```
geky, line 178: nit: unindent these (can change on merge)

lfs.c
```diff
@@ -4911,3 +4911,398 @@ int lfs_migrate(lfs_t *lfs, const struct lfs_config *cfg) {
 }
 
 #endif
+
+#if LFS_THREAD_SAFE
+
+int lfs_format_ts(lfs_t *lfs, const struct lfs_config *config)
+{
```
geky, line 4918: nit: consistent style - squiggly-brackets on same line

lfs.h
```diff
@@ -644,7 +655,106 @@ int lfs_fs_traverse(lfs_t *lfs, int (*cb)(void*, lfs_block_t), void *data);
...
+#define lfs_mkdir lfs_mkdir_raw
+#define lfs_dir_open lfs_dir_open_raw
+#define lfs_dir_close lfs_dir_close_raw
+#define lfs_dir_read lfs_dir_read_raw
+#define lfs_dir_seek lfs_dir_seek_raw
+#define lfs_dir_tell lfs_dir_tell_raw
+#define lfs_dir_rewind lfs_dir_rewind_raw
+#define lfs_fs_size lfs_fs_size_raw
+#define lfs_fs_traverse lfs_fs_traverse_raw
+#define lfs_migrate lfs_migrate_raw
```
geky, line 756: I'm not a fan of redefining functions, as they can cause problems for debuggers. It may be tedious, but I believe each of these should get its own wrapper:
``` c
int lfs_file_close(lfs_t *lfs, lfs_file_t *file) {
    return lfs_file_close_raw(lfs, file);
}
```

Thinking ahead, I'm planning to have logging overhead in each of these, so I don't think any will end up as a simple passthrough. It should be easy enough for the compiler to inline these anyways.

lfs.c
```diff
@@ -4911,3 +4911,398 @@ int lfs_migrate(lfs_t *lfs, const struct lfs_config *cfg) {
...
+    if(err)
+    {
+        return err;
+    }
+    err = lfs_format_raw(lfs, config);
+    config->unlock(config);
+
+    return err;
+}
+int lfs_mount_ts(lfs_t *lfs, const struct lfs_config *config)
```
geky, line 4929: nit: I don't think we need the `_ts` suffix. If the user specifies LFS_THREAD_SAFE, they should be able to assume all functions are thread safe.

lfs_util.h
```diff
@@ -43,6 +43,10 @@ extern "C"
 {
 #endif
 
+// Enables thread-safe wrappers using the lock/unlock callbacks in lfs_config
+#ifndef LFS_THREAD_SAFE
```
geky, line 47: Very minor, but I'm of the opinion we should change this to LFS_THREADSAFE (no space) to match the incoming LFS_READONLY (https://github.com/littlefs-project/littlefs/pull/480)

geky:
Also the CI was broken but should be fixed now. It should trigger on the next push to this branch.

lfs.h
renesas-billgesner: fixed

lfs.c
renesas-billgesner: fixed

lfs.h
renesas-billgesner: fixed

lfs.c
renesas-billgesner: fixed

lfs_util.h
renesas-billgesner: fixed

lfs.c
renesas-billgesner: The angle brackets allow the user to add a custom lfs_util.h by giving their version precedence in the include order. If they don't have a custom lfs_util.h it will still use the one in this repo. I found this slightly useful but I can definitely workaround reverting.

geky: Oh interesting, is this a case that LFS_CONFIG does not work for?
https://github.com/littlefs-project/littlefs/blob/4c9146ea539f72749d6cc3ea076372a81b12cb11/lfs_util.h#L10-L20

It's a minor concern, but it is possible that the angle brackets could cause the wrong file to get included for users.

geky (commented):
Thanks for fixing those, I have just a couple more nits.

This PR makes me wonder if it would be a good time to bring in config improvements (https://github.com/littlefs-project/littlefs/issues/158#issuecomment-494891516), but that can be on a different PR.

lfs.h
```diff
@@ -647,6 +657,36 @@ int lfs_fs_traverse(lfs_t *lfs, int (*cb)(void*, lfs_block_t), void *data);
...
+int lfs_mkdir_raw(lfs_t *lfs, const char *path);
+int lfs_dir_open_raw(lfs_t *lfs, lfs_dir_t *dir, const char *path);
+int lfs_dir_close_raw(lfs_t *lfs, lfs_dir_t *dir);
+int lfs_dir_read_raw(lfs_t *lfs, lfs_dir_t *dir, struct lfs_info *info);
+int lfs_dir_seek_raw(lfs_t *lfs, lfs_dir_t *dir, lfs_off_t off);
+lfs_soff_t lfs_dir_tell_raw(lfs_t *lfs, lfs_dir_t *dir);
+int lfs_dir_rewind_raw(lfs_t *lfs, lfs_dir_t *dir);
+lfs_ssize_t lfs_fs_size_raw(lfs_t *lfs);
+int lfs_fs_traverse_raw(lfs_t *lfs, int (*cb)(void*, lfs_block_t), void *data);
+int lfs_migrate_raw(lfs_t *lfs, const struct lfs_config *cfg);
```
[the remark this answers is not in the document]
geky: Oh, `*_raw` functions should all be made static for this reason, which should help ensure the extra layer is optimized away as much as possible.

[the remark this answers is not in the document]
renesas-billgesner: good call. I had some build issues with implicit declarations and added declarations for the functions that are called internally. I initially added declarations for all the functions but realized that didn't fit the rest of the module. The new declarations could likely be removed after moving around the definitions but I didn't want to muddy up the PR.

lfs.c
```diff
@@ -4911,3 +4911,521 @@ int lfs_migrate(lfs_t *lfs, const struct lfs_config *cfg) {
...
+
+int lfs_removeattr (lfs_t * lfs, const char * path, uint8_t type) {
+    return lfs_removeattr_raw(lfs, path, type);
+}
+
+int lfs_file_open (lfs_t * lfs, lfs_file_t * file, const char * path, int flags) {
+    return lfs_file_open_raw(lfs, file, path, flags);
+}
+
+int lfs_file_opencfg_ts(lfs_t *lfs, lfs_file_t *file, const char *path, int flags, const struct lfs_file_config *config) {
```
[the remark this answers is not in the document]
renesas-billgesner: fixed

lfs.c
```diff
@@ -4911,3 +4911,521 @@ int lfs_migrate(lfs_t *lfs, const struct lfs_config *cfg) {
 }
 
 #endif
+
+#if LFS_THREADSAFE
+
+int lfs_format (lfs_t * lfs, const struct lfs_config * config) {
```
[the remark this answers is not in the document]
renesas-billgesner: fixed

lfs.c
```diff
@@ -4911,3 +4911,521 @@ int lfs_migrate(lfs_t *lfs, const struct lfs_config *cfg) {
...
+        return err;
+    }
+
+    err = lfs_format_raw(lfs, config);
+    config->unlock(config);
+
+    return err;
+}
+
+int lfs_mount (lfs_t * lfs, const struct lfs_config * config) {
```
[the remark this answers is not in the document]
renesas-billgesner: fixed

lfs.c
```diff
@@ -4911,3 +4920,521 @@ int lfs_migrate(lfs_t *lfs, const struct lfs_config *cfg) {
...
+    }
+
+    err = lfs_format_raw(lfs, config);
+    config->unlock(config);
+
+    return err;
+}
+
+int lfs_mount(lfs_t *lfs, const struct lfs_config *config) {
+    int err = config->lock(config);
```
geky, line 4939: Hmm, does it actually make sense to lock here? In theory no one else should be able to access the filesystem before it's initialized.

geky:
Hi @renesas-billgesner, I went ahead and added some additional style tweaks and moved over LFS_TRACE into the new wrappers (nicely cleans some of the intertwining LFS_TRACE calls).

- Stayed on non-system include for lfs_util.h for now
- Named internal functions "lfs_functionraw"
- Merged lfs_fs_traverseraw
- Added LFS_LOCK/UNLOCK macros
- Changed LFS_THREADSAFE from 1/0 to defined/undefined to match LFS_READONLY
- Moved LFS_TRACE calls to API wrapper function

Branch here: https://github.com/littlefs-project/littlefs/tree/thread-safe, PR here: https://github.com/renesas/littlefs/pull/3

Let me know if this looks good to you.

geky:
I've also added a Travis build job to test the compilation with this configuration, it should show up here when Travis gets to it:
https://travis-ci.org/github/littlefs-project/littlefs/jobs/746469616

geky:
I've also been looking at some unrelated config-related changes in https://github.com/littlefs-project/littlefs/pull/491. I haven't rebased onto the LFS_THREADSAFE/LFS_READONLY changes yet, but it raises some questions:

1. > The angle brackets allow the user to add a custom lfs_util.h by giving their version precedence in the include order.

   Is this preferable to `-DLFS_CONFIG=path/to/my_util.h`? https://github.com/littlefs-project/littlefs/pull/491 would be an opportunity to change the lfs_util.h includes, but using angle-brackets seems uncommon in this situation.

2. Due to some name changes, the callbacks for the block devices are now `bd_read`/`bd_prog`/`bd_ctx`. I was thinking for LFS_THREADSAFE to use `mutex_lock`/`mutex_lock`. Do you think it would be useful to have a separate `mutex_ctx` for these operations?

geky:
Also I was curious why there is a 76 byte code-size increase, it looks like when the `*raw` functions have multiple callees, inlining doesn't happen, and GCC isn't smart enough to realize it doesn't need the extra function, perhaps due to function-address-uniqueness rules.

Oh well, it's probably not worth over-optimizing.

Issue #162, opened 2019-03-26
reduce build size for read-only support


mjr-eyelock:
Our team is developing a custom firmware update feature with a custom bootloader. We are not using mbed OS; however, we are investigating the use of littlefs, as a filesystem abstraction to manage storage on a SPI-connected flash chip. Our main application will use the full read and write features of littlefs, but we are trying to reduce the size of our separate bootloader app by building in only those portions of littlefs needed to _read_ from the filesystem. So far, after a bit of basic hacking, we've only been able to reduce our bootloader application size by a few hundred bytes.

If perhaps some build flags were put into place that would produce build of a subset of the library that would support only read operations, is it feasible to expect the size difference to be more than just a few hundred bytes? If so, might this be added as a feature?

geky:
Hi @matthewjrichey, from previous experiments I would guess building littlefs readonly would get you to about 1/3rd the code size. Most of the code cost comes from managing writes.

If you remove the writing if statements from [lfs_file_flush](https://github.com/ARMmbed/littlefs/blob/0907ba7813173901bc9a7689e8781434629a3696/lfs.c#L2508-L2573) and [lfs_file_sync](https://github.com/ARMmbed/littlefs/blob/0907ba7813173901bc9a7689e8781434629a3696/lfs.c#L2586-L2623), the compiler should be able to garbage collect the write functions and give you a more accurate prediction.

Unfortunately this is relatively low priority for me at the moment, but I'd be happy to take in a PR.

thrasher8390:
I've been wondering the same thing. I'll have a look to see if I can get the build size down. If anyone else has tried this, it would be great to hear from you.

geky:
Just an FYI, @KRekha3 was able to modify littlefs for readonly and the resulting size was ~5KiB: https://github.com/ARMmbed/littlefs/issues/183

PR #480, opened 2020-10-28
Add LFS_READONLY define, to allow smaller builds providing read-only mode

maximevince:
This PR tries to solve #162 , by providing an `LFS_READONLY` define.

LittleFS seems to be preferred over SPIFFS in many places and projects right now.
However, for my bootloader / firmware upgrade use-case, the code size matters a lot.
A bootloader could do with a read-only version of the file system, which should be a lot smaller in size.

With SPIFFS I am able to get a read-only version without caching compiled to around 3 KB (compiled for Cortex-M0),
while the read-write version with caching weighs in at around 13 KB, which is comparable to LittleFS.

By defining out a bunch of write-related functionality, I was able to reduce the read-only version of LittleFS to around 5 KB for that same Cortex-M0 target, which seems acceptable for bootloader use.

I have separately `#ifdef`'ed every function that is only needed for writing functionality.
However, I'm not sure if this is the preferred way, of if you'd rather have a single `#ifdef` around the whole lump of code.

Let me know if I can do anything to improve this PR.

Cheers and thanks for the awesome project! :+1:

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -136,6 +136,7 @@ static int lfs_bd_cmp(lfs_t *lfs,
     return LFS_CMP_EQ;
 }
 
+#ifndef LFS_READONLY
 static int lfs_bd_flush(lfs_t *lfs,
         lfs_cache_t *pcache, lfs_cache_t *rcache, bool validate) {
     if (pcache->block != LFS_BLOCK_NULL && pcache->block != LFS_BLOCK_INLINE) {
@@ -168,7 +169,9 @@ static int lfs_bd_flush(lfs_t *lfs,
 
     return 0;
 }
+#endif
 
+#ifndef LFS_READONLY
 static int lfs_bd_sync(lfs_t *lfs,
         lfs_cache_t *pcache, lfs_cache_t *rcache, bool validate) {
     lfs_cache_drop(lfs, rcache);
@@ -182,7 +185,9 @@ static int lfs_bd_sync(lfs_t *lfs,
     LFS_ASSERT(err <= 0);
     return err;
 }
+#endif
 
+#ifndef LFS_READONLY
 static int lfs_bd_prog(lfs_t *lfs,
         lfs_cache_t *pcache, lfs_cache_t *rcache, bool validate,
         lfs_block_t block, lfs_off_t off,
@@ -228,13 +233,16 @@ static int lfs_bd_prog(lfs_t *lfs,
 
     return 0;
 }
+#endif
 
+#ifndef LFS_READONLY
 static int lfs_bd_erase(lfs_t *lfs, lfs_block_t block) {
     LFS_ASSERT(block < lfs->cfg->block_count);
     int err = lfs->cfg->erase(lfs->cfg, block);
     LFS_ASSERT(err <= 0);
     return err;
 }
+#endif
 
 
 /// Small type-level utilities ///
@@ -388,10 +396,12 @@ static void lfs_ctz_fromle32(struct lfs_ctz *ctz) {
     ctz->size = lfs_fromle32(ctz->size);
 }
 
+#ifndef LFS_READONLY
 static void lfs_ctz_tole32(struct lfs_ctz *ctz) {
     ctz->head = lfs_tole32(ctz->head);
     ctz->size = lfs_tole32(ctz->size);
 }
+#endif
 
 static inline void lfs_superblock_fromle32(lfs_superblock_t *superblock) {
     superblock->version     = lfs_fromle32(superblock->version);
@@ -413,6 +423,7 @@ static inline void lfs_superblock_tole32(lfs_superblock_t *superblock) {
 
 
 /// Internal operations predeclared here ///
+#ifndef LFS_READONLY
 static int lfs_dir_commit(lfs_t *lfs, lfs_mdir_t *dir,
         const struct lfs_mattr *attrs, int attrcount);
 static int lfs_dir_compact(lfs_t *lfs,
@@ -429,10 +440,13 @@ static lfs_stag_t lfs_fs_parent(lfs_t *lfs, const lfs_block_t dir[2],
         lfs_mdir_t *parent);
 static int lfs_fs_relocate(lfs_t *lfs,
         const lfs_block_t oldpair[2], lfs_block_t newpair[2]);
+#endif
 int lfs_fs_traverseraw(lfs_t *lfs,
         int (*cb)(void *data, lfs_block_t block), void *data,
         bool includeorphans);
+#ifndef LFS_READONLY
 static int lfs_fs_forceconsistency(lfs_t *lfs);
+#endif
 static int lfs_deinit(lfs_t *lfs);
 #ifdef LFS_MIGRATE
 static int lfs1_traverse(lfs_t *lfs,
@@ -440,6 +454,7 @@ static int lfs1_traverse(lfs_t *lfs,
 #endif
 
 /// Block allocator ///
+#ifndef LFS_READONLY
 static int lfs_alloc_lookahead(void *p, lfs_block_t block) {
     lfs_t *lfs = (lfs_t*)p;
     lfs_block_t off = ((block - lfs->free.off)
@@ -451,6 +466,7 @@ static int lfs_alloc_lookahead(void *p, lfs_block_t block) {
 
     return 0;
 }
+#endif
 
 static void lfs_alloc_ack(lfs_t *lfs) {
     lfs->free.ack = lfs->cfg->block_count;
@@ -465,6 +481,7 @@ static void lfs_alloc_reset(lfs_t *lfs) {
     lfs_alloc_ack(lfs);
 }
 
+#ifndef LFS_READONLY
 static int lfs_alloc(lfs_t *lfs, lfs_block_t *block) {
     while (true) {
         while (lfs->free.i != lfs->free.size) {
@@ -510,6 +527,7 @@ static int lfs_alloc(lfs_t *lfs, lfs_block_t *block) {
         }
     }
 }
+#endif
 
 /// Metadata pair and directory operations ///
 static lfs_stag_t lfs_dir_getslice(lfs_t *lfs, const lfs_mdir_t *dir,
@@ -642,6 +660,7 @@ static int lfs_dir_getread(lfs_t *lfs, const lfs_mdir_t *dir,
     return 0;
 }
 
+#ifndef LFS_READONLY
 static int lfs_dir_traverse_filter(void *p,
         lfs_tag_t tag, const void *buffer) {
     lfs_tag_t *filtertag = p;
@@ -669,7 +688,9 @@ static int lfs_dir_traverse_filter(void *p,
 
     return false;
 }
+#endif
 
+#ifndef LFS_READONLY
 static int lfs_dir_traverse(lfs_t *lfs,
         const lfs_mdir_t *dir, lfs_off_t off, lfs_tag_t ptag,
         const struct lfs_mattr *attrs, int attrcount,
@@ -763,6 +784,7 @@ static int lfs_dir_traverse(lfs_t *lfs,
         }
     }
 }
+#endif
 
 static lfs_stag_t lfs_dir_fetchmatch(lfs_t *lfs,
         lfs_mdir_t *dir, const lfs_block_t pair[2],
@@ -1200,6 +1222,7 @@ struct lfs_commit {
     lfs_off_t end;
 };
 
+#ifndef LFS_READONLY
 static int lfs_dir_commitprog(lfs_t *lfs, struct lfs_commit *commit,
         const void *buffer, lfs_size_t size) {
     int err = lfs_bd_prog(lfs,
@@ -1214,7 +1237,9 @@ static int lfs_dir_commitprog(lfs_t *lfs, struct lfs_commit *commit,
     commit->off += size;
     return 0;
 }
+#endif
 
+#ifndef LFS_READONLY
 static int lfs_dir_commitattr(lfs_t *lfs, struct lfs_commit *commit,
         lfs_tag_t tag, const void *buffer) {
     // check if we fit
@@ -1259,7 +1284,9 @@ static int lfs_dir_commitattr(lfs_t *lfs, struct lfs_commit *commit,
     commit->ptag = tag & 0x7fffffff;
     return 0;
 }
+#endif
 
+#ifndef LFS_READONLY
 static int lfs_dir_commitcrc(lfs_t *lfs, struct lfs_commit *commit) {
     const lfs_off_t off1 = commit->off;
     const uint32_t crc1 = commit->crc;
@@ -1352,7 +1379,9 @@ static int lfs_dir_commitcrc(lfs_t *lfs, struct lfs_commit *commit) {
 
     return 0;
 }
+#endif
 
+#ifndef LFS_READONLY
 static int lfs_dir_alloc(lfs_t *lfs, lfs_mdir_t *dir) {
     // allocate pair of dir blocks (backwards, so we write block 1 first)
     for (int i = 0; i < 2; i++) {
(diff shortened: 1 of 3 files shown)


geky:
Hi @maximevince, thanks for creating a PR! Sorry being late in getting to this. This looks great.

> I have separately #ifdef'ed every function that is only needed for writing functionality.
However, I'm not sure if this is the preferred way, of if you'd rather have a single #ifdef around the whole lump of code.

I hadn't thought about this, it has the nice side-effect that it is easy to know which functions are read-only in the source code 👍

lfs.c
```diff
@@ -3436,11 +3530,15 @@ static int lfs_commitattr(lfs_t *lfs, const char *path,
...
             {LFS_MKTAG(LFS_TYPE_USERATTR + type, id, size), buffer}));
 }
+#endif
 
 int lfs_setattr(lfs_t *lfs, const char *path,
         uint8_t type, const void *buffer, lfs_size_t size) {
     LFS_TRACE("lfs_setattr(%p, \"%s\", %"PRIu8", %p, %"PRIu32")",
             (void*)lfs, path, type, buffer, size);
+#ifdef LFS_READONLY
+    return LFS_ERR_NOSYS;
```
geky, line 3540: It looks like you are using LFS_ERR_NOSYS for any functions exposed in lfs.h. I think it would be better to also conditionally omit the definitions in the header file.

Build errors are usually more helpful than runtime errors, and worst case the user can provide an additional stub function that returns LFS_ERR_NOSYS.

lfs.h
```diff
@@ -76,6 +76,7 @@ enum lfs_error {
     LFS_ERR_EXIST       = -17,  // Entry already exists
     LFS_ERR_NOTDIR      = -20,  // Entry is not a dir
     LFS_ERR_ISDIR       = -21,  // Entry is a dir
+    LFS_ERR_NOSYS       = -38,  // Function not implemened
```
geky, line 79: There is actually a specific error code for read-only filesystems. This would be a good time to use it:

``` c
LFS_ERR_ROFS        = -30,  // Read-only file system
```

lfs.c
maximevince: Right! I did that because it's the way SPIFFS implemented their read-only support. Not that we have to copy that behavior, of course, but that's where I'm migrating from.
I will rework this to simply omit the definitions as well.
I'll take into account the other remarks as well, and update the PR.

geky:
Also the CI was broken but should be fixed now. It should trigger on the next push to this branch.

dpgeorge:
It's good to see support for read-only mode in littlefs.  I did make a very similar patch to this (with the option called `LFS_ENABLE_WRITES`) but did not get time to make a PR out of it.  In my use of it in a bootloader, littlefs with write support takes about 11000 bytes.  In a read-only build that goes down to about 4200 bytes (on a Cortex-M7).  That's very similar to what's reported above.

Will be good to see this merged!

maximevince:
> It's good to see support for read-only mode in littlefs. I did make a very similar patch to this (with the option called `LFS_ENABLE_WRITES`) but did not get time to make a PR out of it. In my use of it in a bootloader, littlefs with write support takes about 11000 bytes. In a read-only build that goes down to about 4200 bytes (on a Cortex-M7). That's very similar to what's reported above.
> 
> Will be good to see this merged!

Alright, nice to hear from you @dpgeorge.
I believe we briefly conversed back when I was working on [picoTCP](https://github.com/tass-belgium/picotcp) and we made a port in the early days of [micropython](https://github.com/micropython/micropython).

If you see any differences between the parts of the code I have `#ifdef`-ed, or where it could still be made smaller, let me know!

maximevince:
Just updated the PR with a new commit:
* Removed write-related functions from the header altogether, and adjusted `lfs.c` accordingly
* Replaced `LFS_ERR_NOSYS` by `LFS_ERR_ROFS` (-30) 

I am not entirely sure what should happen to `int lfs_file_sync(lfs_t *lfs, lfs_file_t *file)`:
Is this function at all needed for LittleFS in read-only mode? Right now I only `#ifdef`ed the `LFS_F_DIRTY` portion out of it.
And I have the same question for `lfs_file_flush(lfs, file)`, too.

geky:
Oh that is a good question. I'm going to say no to both, I don't see how they could be used in read-only mode.

If there some strange use case we're missing, they are easy to add back into the read-only build. It would be harder to remove them later.

lfs.c
```diff
@@ -2390,11 +2438,15 @@ int lfs_file_opencfg(lfs_t *lfs, lfs_file_t *file,
 
     // deorphan if we haven't yet, needed at most once after poweron
     if ((flags & 3) != LFS_O_RDONLY) {
+#ifdef LFS_READONLY
+        return LFS_ERR_ROFS;
+#else
```
dpgeorge, line 2443: I would think that files can only ever be opened in `LFS_O_RDONLY` mode, so this entire if-block can just be excluded with the #if .

dpgeorge: There's also another similar block down below in this function which could also be #if'd out.

geky: Actually that's a good point, an `LFS_ASSERT((flags & 3) == LFS_O_RDONLY)` would work better here.

dpgeorge:
This is now looking almost exactly the same as the patch I made.

lfs.c
maximevince: I am not a big fan of "crashing" with an assert, instead of returning an error when a `lfs_file_opencfg()` is called with the wrong arguments, but I'll implement it that way, if you want.

Let's say an OS tries to write to a file mounted using the LittleFS filesystem, but LittleFS is compiled for read-only. Wouldn't it be more elegant to return an error indicating it's not possible to write, instead of asserting ?

geky: At the moment, it should be an assert for consistency (writing a file opened as rdonly currently asserts, for example).

Returning an error can be more graceful if there is an OS that understands that error, but for many littlefs applications this isn't the case. It's easy for the context of the error to get lost as "file open failed". Looking at POSIX errors, which littlefs follows, writing to an rdonly file only returns EINVAL. An assert takes you directly to the source of the problem, sort of like a poor-man's stack trace.

There's also a code cost for returning/handling errors, an assert can be dropped at compile time. Runtime checks can be added in the OS, but they can't be removed. This makes asserts preferred for conditions that should never happen unless there is a programmer mistake.

But you're not the first person to comment on this, maybe it would be worth adding a configuration option to change the use-error asserts to return errors. The next steps there would be to create a separate issue/PR.

maximevince: Alright, thanks for the detailed explanation. I can understand the reasoning indeed.

I have just pushed a new commit implementing the `LFS_ASSERT` for write attempts, as well as more code commented out, in case of `LFS_READONLY`.

dpgeorge: One idea would be to also #if out the write-based constants, like `LFS_O_WRONLY`.  Then it would be impossible to compile code that opened files for writing.

geky: Oh, that's an even better idea! @dpgeorge with the great ideas

maximevince: Okay, I just pushed another commit implementing this idea.
I had to make some extra `#ifdef` sections in other functions checking for these flags, so please carefully review that.
What I also did, is marking each `#endif` section with a comment indicating which `#ifdef` it is corresponding to.

lfs.c
```diff
@@ -2464,7 +2523,12 @@ int lfs_file_opencfg(lfs_t *lfs, lfs_file_t *file,
 
     // fetch attrs
     for (unsigned i = 0; i < file->cfg->attr_count; i++) {
-        if ((file->flags & 3) != LFS_O_WRONLY) {
+#ifdef LFS_READONLY
+        /* always LFS_O_RDONLY */
+#else
+        if ((file->flags & 3) != LFS_O_WRONLY)
+#endif /* LFS_READONLY */
+        {
```
geky, line 2531: Hmm, this dangling squiggly bracket is a risk.

It looks like this can be changed to:
```
if ((file->flags & LFS_O_RDONLY) == LFS_O_RDONLY) {
```

Which would avoid the syntax issue.

lfs.c
```diff
@@ -2476,6 +2540,7 @@ int lfs_file_opencfg(lfs_t *lfs, lfs_file_t *file,
             }
         }
 
+#ifndef LFS_READONLY
         if ((file->flags & 3) != LFS_O_RDONLY) {
```
geky, line 2544: Related to the above, this could be `(file->flags & LFS_O_WRONLY) == LFS_O_WRONLY` for consistency

lfs.h
```diff
@@ -123,12 +123,14 @@ enum lfs_type {
...
     LFS_O_WRONLY = 2,         // Open a file as write only
     LFS_O_RDWR   = 3,         // Open a file as read and write
     LFS_O_CREAT  = 0x0100,    // Create a file if it does not exist
     LFS_O_EXCL   = 0x0200,    // Fail if a file already exists
     LFS_O_TRUNC  = 0x0400,    // Truncate the existing file to zero size
     LFS_O_APPEND = 0x0800,    // Move to end of file on every write
+#endif
 
     // internally used flags
     LFS_F_DIRTY   = 0x010000, // File does not match storage
```
dpgeorge, line 136: Is it possible (and worth it) also #if'ing these constants like `LFS_F_DIRTY` which are not supposed to be used in read-only mode?

lfs.c
```diff
@@ -388,10 +396,12 @@ static void lfs_ctz_fromle32(struct lfs_ctz *ctz) {
     ctz->size = lfs_fromle32(ctz->size);
 }
 
+#ifndef LFS_READONLY
 static void lfs_ctz_tole32(struct lfs_ctz *ctz) {
     ctz->head = lfs_tole32(ctz->head);
     ctz->size = lfs_tole32(ctz->size);
 }
+#endif /* LFS_READONLY */
```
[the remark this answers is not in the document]
maximevince: fixed in 0360629

lfs.c
maximevince: ok, reworked, but will now result in slightly bigger/slower code for read-only builds,
since `if ((file->flags & LFS_O_RDONLY) == LFS_O_RDONLY) {` will still be checked each time.

maximevince: fixed in 83fd2e8

lfs.c
maximevince: fixed in 83fd2e8

lfs.h
maximevince: It was easy for `LFS_F_DIRTY` and `LFS_F_WRITING`.

However, I am not sure about `LFS_F_ERRED`.
It's documented as "An error occurred during write", but it has one non-obvious use at the end of `int lfs_file_opencfg()`.
It seems you could end up [there](https://github.com/maximevince/littlefs/blob/83fd2e868855bacd3e6e8ae95876af3c1895580a/lfs.c#L2594.) when e.g. `lfs_malloc()` [fails](https://github.com/maximevince/littlefs/blob/83fd2e868855bacd3e6e8ae95876af3c1895580a/lfs.c#L2559), or when opening a file for read hits an `LFS_ERR_NOENT` ([here](https://github.com/maximevince/littlefs/blob/83fd2e868855bacd3e6e8ae95876af3c1895580a/lfs.c#L2535)) (among other possibilities). 
Both of which cases do not seem to be strictly write-related.
So I conclude that this constant is needed in `LFS_READONLY` mode as well.

This is implemented in 35cc31e

lfs.c
geky: Hmm, that's a good point, but I don't know if the readability penalty is worth it

lfs.h
geky: Ah, it's not actually used in `lfs_file_close` but in the implicit `lfs_file_sync`. The purpose of the LFS_F_ERRED flag is to tell `lfs_file_close` to release resources without writing out to disk. `lfs_file_opencfg` immediately calls `lfs_file_close` if an error occurs.

You have the implicit `lfs_file_sync` removed in LFS_READONLY mode, so it should be safe to remove [this line](https://github.com/maximevince/littlefs/blob/83fd2e868855bacd3e6e8ae95876af3c1895580a/lfs.c#L2594) as well.

lfs.c
```diff
@@ -2787,17 +2863,23 @@ int lfs_file_sync(lfs_t *lfs, lfs_file_t *file) {
...
 }
+#endif
 
 lfs_ssize_t lfs_file_read(lfs_t *lfs, lfs_file_t *file,
         void *buffer, lfs_size_t size) {
     LFS_TRACE("lfs_file_read(%p, %p, %p, %"PRIu32")",
             (void*)lfs, (void*)file, buffer, size);
     LFS_ASSERT(file->flags & LFS_F_OPENED);
+#ifdef LFS_READONLY
+    /* always LFS_O_RDONLY */
```
geky, line 2874: If you're making changes, this should be a `//` comment for consistency, otherwise it's not a bit deal.

lfs.c
```diff
@@ -3122,11 +3211,15 @@ lfs_soff_t lfs_file_size(lfs_t *lfs, lfs_file_t *file) {
...
+
+#ifndef LFS_READONLY
     if (file->flags & LFS_F_WRITING) {
         LFS_TRACE("lfs_file_size -> %"PRId32,
                 lfs_max(file->pos, file->ctz.size));
         return lfs_max(file->pos, file->ctz.size);
-    } else {
+    } else
+#endif
+    {
```
geky, line 3222: nit: This would be slightly better:

``` c
#ifndef LFS_READONLY
    if (file->flags & LFS_F_WRITING) {
        LFS_TRACE("lfs_file_size -> %"PRId32,
                lfs_max(file->pos, file->ctz.size));
        return lfs_max(file->pos, file->ctz.size);
    } else {
#else
    {
#endif
        LFS_TRACE("lfs_file_size -> %"PRId32, file->ctz.size);
        return file->ctz.size;
    }
```

Or in this one specific case we can just use short-circuiting:
``` c
#ifndef LFS_READONLY
    if (file->flags & LFS_F_WRITING) {
        LFS_TRACE("lfs_file_size -> %"PRId32,
                lfs_max(file->pos, file->ctz.size));
        return lfs_max(file->pos, file->ctz.size);
    }
#endif

    LFS_TRACE("lfs_file_size -> %"PRId32, file->ctz.size);
    return file->ctz.size;
```

lfs.c
```diff
@@ -2418,6 +2470,10 @@ int lfs_file_opencfg(lfs_t *lfs, lfs_file_t *file,
     lfs->mlist = (struct lfs_mlist*)file;
 
     if (tag == LFS_ERR_NOENT) {
+#ifdef LFS_READONLY
+        err = LFS_ERR_NOENT;
+        goto cleanup;
+#else
```
geky, line 2476: I just have a minor nit here, since the `#ifdef` starts inside an if statement and ends outside. It's a complex location for LFS_READONLY, but maybe this would be better this way?

``` c
#ifdef LFS_READONLY
    if (tag == LFS_ERR_NOENT) {
        err = LFS_ERR_NOENT;
        goto cleanup;
#else
    if (tag == LFS_ERR_NOENT) {
        if (!(flags & LFS_O_CREAT)) {
            err = LFS_ERR_NOENT;
            goto cleanup;
        }
        ... snip ...
    } else if (flags & LFS_O_EXCL) {
        err = LFS_ERR_EXIST;
        goto cleanup;
#endif
    } else if (lfs_tag_type3(tag) != LFS_TYPE_REG) {
        err = LFS_ERR_ISDIR;
        goto cleanup;
        ... snip ...
    }
```

Other than that these changes are looking great.

maximevince: 94c9c4c should do the trick!

geky:
This looks great, thanks for this! Will bring this in on the next minor release.

I added a Travis job, so we should see a read-only build size in the CI results now.

geky (approved):

lfs.h
maximevince: I only noticed this comment now. I have `ifdef`ed `LFS_F_ERRED` out completely. (99c960b)

Issue #484, opened 2020-11-11
Dubious offset calculation in lfs_alloc_reset


gtaska:
The code for lfs_alloc_reset is:

```
// Invalidate the lookahead buffer. This is done during mounting and
// failed traversals
static void lfs_alloc_reset(lfs_t *lfs) {
    lfs->free.off = lfs->seed % lfs->cfg->block_size;
    lfs->free.size = 0;
    lfs->free.i = 0;
    lfs_alloc_ack(lfs);
}
```

In this code, lfs->free.off is initialized with a range-bound version of lfs->seed - however it is bounded based on the configured block_size. Everywhere else that uses/updates lfs->free.off bounds it based on block_count (which makes more sense, as you are trying to find an index such that 0 <= index < block_count).

I don't think that this is necessarily going to cause immediate issues, as everywhere else it is used it appears to be subject to another modulus operation, however in certain configurations where block_size is small compared to block_count this may bias the wear leveling.

geky:
Hi @gtaska, thanks for raising an issue! You're right this is innocuous but should be fixed.

PR #487, opened 2020-11-14
Fix several wear-leveling issues found in lfs_alloc_reset

geky:
- Fixed incorrect modulus in lfs_alloc_reset

   Modulus of the offset by block_size was clearly a typo, and should be block_count. Interesting to note that later moduluses during alloc calculations prevents this from breaking anything, but as gtaska notes it could skew the wear-leveling distribution.

- Removed unnecessary randomization of offsets in lfs_alloc_reset

  On first read, randomizing the allocators offset may seem appropriate for lfs_alloc_reset. However, it ends up using the filesystem-fed pseudorandom seed in situations it wasn't designed for.

  As noted by gtaska, the combination of using xors for feeding the seed and multiple traverses of the same CRCs can cause the seed to flip to zeros with concerning frequency.

  Removed the randomization from lfs_alloc_reset, leaving it in only lfs_mount.

- Switched to CRC as seed collection function instead of xor

  As noted by gtaska, we are sitting on a better hash-combining function than xor: CRC. Previous issues with xor were solvable, but relying on xor for this isn't really worth the risk when we already have a CRC function readily available.

  To quote a study found by gtaska:
  https://michiel.buddingh.eu/distribution-of-hash-values
  > CRC32 seems to score really well, but its graph is skewed by the results of Dataset 5 (binary numbers), which may or may not be too synthetic to be considered a fair benchmark. But even if you substract the results from that test, it does not fare significantly worse than other, cryptographic hash functions.

Should fix https://github.com/littlefs-project/littlefs/issues/484
Related discussion https://github.com/littlefs-project/littlefs/issues/488
Found by @guiserle and @gtaska

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -459,7 +459,7 @@ static void lfs_alloc_ack(lfs_t *lfs) {
 // Invalidate the lookahead buffer. This is done during mounting and
 // failed traversals
 static void lfs_alloc_reset(lfs_t *lfs) {
-    lfs->free.off = lfs->seed % lfs->cfg->block_size;
+    lfs->free.off = lfs->seed % lfs->cfg->block_count;
     lfs->free.size = 0;
     lfs->free.i = 0;
     lfs_alloc_ack(lfs);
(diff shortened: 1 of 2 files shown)


geky:
The new allocation pattern from changing to CRC as a hash-combination function exposed an interesting bad-block-erase-is-noop bug if we happen to write metadata to a block that is bad, stuck, and contains the exact data we were about to write. Added a fix in https://github.com/littlefs-project/littlefs/pull/487/commits/0aba71d0d6d57813db2306670a9dd9d25efdfae5.

Issue #481, opened 2020-11-03
build failed when using the tasking compiler!


snikeguo:
IDE:AURIX DS
compiler:CCTC ( tasking company)

```
ctc W562: ["..\littlefs-2.2.1\lfs_util.h" 141/26] unary minus applied to unsigned value
ctc W560: ["../littlefs-2.2.1/lfs.c" 361/24] possible truncation at implicit conversion to type "unsigned char"
ctc W560: ["../littlefs-2.2.1/lfs.c" 738/29] possible truncation at implicit conversion to type "unsigned short int"
ctc W560: ["../littlefs-2.2.1/lfs.c" 980/21] possible truncation at implicit conversion to type "unsigned short int"
ctc W560: ["../littlefs-2.2.1/lfs.c" 1043/16] possible truncation at implicit conversion to type "unsigned char"
ctc W515: ["../littlefs-2.2.1/lfs.c" 1405/200] side effects of 'sizeof' operand are ignored
ctc W515: ["../littlefs-2.2.1/lfs.c" 1959/37] side effects of 'sizeof' operand are ignored
ctc W515: ["../littlefs-2.2.1/lfs.c" 1981/42] side effects of 'sizeof' operand are ignored
ctc W515: ["../littlefs-2.2.1/lfs.c" 1996/39] side effects of 'sizeof' operand are ignored
ctc W515: ["../littlefs-2.2.1/lfs.c" 2438/45] side effects of 'sizeof' operand are ignored
ctc W515: ["../littlefs-2.2.1/lfs.c" 2778/45] side effects of 'sizeof' operand are ignored
ctc W515: ["../littlefs-2.2.1/lfs.c" 3206/37] side effects of 'sizeof' operand are ignored
ctc W515: ["../littlefs-2.2.1/lfs.c" 3331/40] side effects of 'sizeof' operand are ignored
ctc W515: ["../littlefs-2.2.1/lfs.c" 3350/44] side effects of 'sizeof' operand are ignored
ctc W515: ["../littlefs-2.2.1/lfs.c" 3440/38] side effects of 'sizeof' operand are ignored
ctc W515: ["../littlefs-2.2.1/lfs.c" 3642/42] side effects of 'sizeof' operand are ignored
ctc W515: ["../littlefs-2.2.1/lfs.c" 4055/48] side effects of 'sizeof' operand are ignored
[...]

PR #494, opened 2020-12-01
Fix compilation errors when using tasking compiler(cctc)

snikeguo:
Fix compilation errors when using tasking compiler

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -358,7 +358,7 @@ static inline bool lfs_gstate_hasorphans(const lfs_gstate_t *a) {
 }
 
 static inline uint8_t lfs_gstate_getorphans(const lfs_gstate_t *a) {
-    return lfs_tag_size(a->tag);
+    return (uint8_t)lfs_tag_size(a->tag);
 }
 
 static inline bool lfs_gstate_hasmove(const lfs_gstate_t *a) {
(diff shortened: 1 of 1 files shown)


lfs.c
```diff
@@ -1400,8 +1400,17 @@ static int lfs_dir_drop(lfs_t *lfs, lfs_mdir_t *dir, lfs_mdir_t *tail) {
...
+    {
+        struct lfs_mattr __lfs_mattr__[]=
+        {
+            {
+                (((lfs_tag_t)(LFS_TYPE_TAIL + tail->split) << 20) | ((lfs_tag_t)(0x3ff) << 10) | (lfs_tag_t)(8)), 
+                tail->tail
+            },
+        };
+        int __attrcount__=sizeof(__lfs_mattr__)/sizeof(struct lfs_mattr);
+        err = lfs_dir_commit(lfs, dir, __lfs_mattr__,__attrcount__);
```
geky, line 1412: Sorry, this change adds too much noise and defeats the API of the internal lfs_dir_commit function. Is there an alternative?

geky (changes requested):
Hi @snikeguo, thanks for making a PR (related https://github.com/littlefs-project/littlefs/issues/481).

Happy to take in the casts, thanks for those.

However, the expanded LFS_MKATTRS macro adds to much noise. Is there an alternative way to make the tasking compiler happy?

lfs.c
snikeguo: no .the tasking compiler is a professional compiler in the field of automotive electronics. This compiler does not support high C language standard(such as : c14)

geky: What warning does cctc output for this line? LFS_MKATTRS just passes a point to a stack allocated array of static size, which shouldn't be too strange.

littlefs is should be strictly c99, otherwise useful features such as inline, arbitrary variable declarations, uint32_t, wouldn't be available.

PR #495, opened 2020-12-04
Minor release: v2.3

geky:
You may or may not have noticed a slight change to this repo's branches. There is now a [devel](https://github.com/littlefs-project/littlefs/tree/devel) branch, which for the near future, will act as a staging area for release before they are merged to master.

Is this an indirect method of having master/release branches? Maybe! But I'm still learning and this should help how fast I can merge PRs without committing to releases at the same time.

Anyways!

This PR brings in:
- #395 lfs_bd_cmp() compares more bytes at one time
- #412 Added littlefs-python to the related projects section
- #443 Assert that the file isnt open in lfs_file_opencfg
- #470 Add thread safe wrappers
- #480 Add LFS_READONLY define, to allow smaller builds providing read-only mode
- #486 Fix assert
- #487 Fix several wear-leveling issues found in lfs_alloc_reset
- #490 Fix allocation-eviction issue when erase state is multiple of block_cycles+1

A draft of the release notes follows:

### What's new?

Thanks to @maximevince, @renesas-billgesner, and @roykuper13, we now have two impactful compile-time options for littlefs!

- LFS_READONLY

  Define `LFS_READONLY` to enable read-only mode. This will limit littlefs to functions that do not modify the disk, but will reduce the code cost to ~1/3. This is a great addition for bootloaders or other applications where a small, read-only storage blob is useful.

- LFS_THREADSAFE

   Define `LFS_THREADSAFE` to enable thread-safe mode. Note this needs the `lock` and `unlock` callbacks in the lfs_config struct.

   When defined, littlefs will call the related `lock` and `unlock` callbacks when necessary to synchronize the filesystem. Allowing easy integration into an RTOS without needing an OS or wrapper layer.
  
In addition, this release brings in a number of improvements:
- More strict user asserts on common file/dir open mistakes thanks to @NoahGorny (https://github.com/littlefs-project/littlefs/pull/443)
- More efficient, 8-byte lfs_bd_cmp thanks to @gmpy (see https://github.com/littlefs-project/littlefs/pull/395 for measurements)
- A number of wear-leveling fixes thanks to @guiserle and @gtaska (https://github.com/littlefs-project/littlefs/pull/487)
- Fixes for allocation-eviction math thanks to @tim-nordell-nimbelink (https://github.com/littlefs-project/littlefs/pull/490)

And a number of other fixes.

Will update this with the related code sizes when CI finishes, it will be interesting to the see the different code costs.

Issue #245, opened 2019-07-25
lfs assert from lfs_ctz_find  (head >= 2 && head <= lfs->cfg->block_count) w/ 1.7.1


asteriskSF:
We're using LFS 1.7.1 in our system and recently had an issue on a single device where an LFS_ASSERT was triggered when lfs_file_read was called.   We're wondering if this is a flaw in 1.7.1 or possibly related to something we might be doing incorrectly in our use of the lfs API?

The LFS_ASSERT occurred in lfs_ctz_find in lfs.c line 1144.    

The specific assertion says:         LFS_ASSERT(head >= 2 && head <= lfs->cfg->block_count);

This is the relevant code:

static int lfs_ctz_find(lfs_t *lfs,
        lfs_cache_t *rcache, const lfs_cache_t *pcache,
        lfs_block_t head, lfs_size_t size,
        lfs_size_t pos, lfs_block_t *block, lfs_off_t *off) {
    if (size == 0) {
        *block = 0xffffffff;
        *off = 0;
        return 0;
    }

    lfs_off_t current = lfs_ctz_index(lfs, &(lfs_off_t){size-1});
    lfs_off_t target = lfs_ctz_index(lfs, &pos);

    while (current > target) {
        lfs_size_t skip = lfs_min(
                lfs_npw2(current-target+1) - 1,
                lfs_ctz(current));

        int err = lfs_cache_read(lfs, rcache, pcache, head, 4*skip, &head, 4);
        head = lfs_fromle32(head);
        if (err) {
            return err;
        }

        LFS_ASSERT(head >= 2 && head <= lfs->cfg->block_count);
        current -= 1 << skip;
    }

    *block = head;
    *off = pos;
    return 0;
}

The git hash we are using: 

SHA-1: d3a2cf48d449f259bc15a3a1058323132f3eeef7

Our configuration is:

// Statically allocated read buffer. Must be read sized.
static uint8_t read_buffer[256];

// Statically allocated program buffer. Must be program sized.
static uint8_t prog_buffer[256];

// Statically allocated lookahead buffer. Must be 1 bit per lookahead block.
static uint8_t lookahead_buffer[128];

// configuration of the filesystem is provided by this struct
static struct lfs_config m_lfsCfg = {
    // block device operations
    .read  = lfs_read,
    .prog  = lfs_prog,
    .erase = lfs_erase,
    .sync  = lfs_sync,

    // block device configuration
[...]

PR #497, opened 2020-12-07
Forward-looking erase-state CRCs

geky:
Original issue here: https://github.com/littlefs-project/littlefs/issues/245#issuecomment-606050523

This is a proposal to fix the out-of-order writes found by @pjsg's fuzzing work.

The problem is that it is possible for (non-NOR) block devices to write pages in any order, or to even write random data in the case of a power-loss. This breaks littlefs's use of the first bit in a page to indicate the erase-state.

@pjsg notes this behavior is documented in the W25Q here:
https://community.cypress.com/docs/DOC-10507

---

The basic idea here is to CRC the next page, and use this "erase-state CRC" to check if the next page is erased and ready to accept programs.

```
.------------------. \   commit
|     metadata     | |
|                  | +---.
|                  | |   |
|------------------| |   |
| erase-state CRC -----. |
|------------------| | | |
|   commit CRC    ---|-|-'
|------------------| / |
|     padding      |   | padding (doesn't need CRC)
|                  |   |
|------------------| \ | next prog
|     erased?      | +-'
|        |         | |
|        v         | /
|                  |
|                  |
'------------------'
```

This is made a bit annoying since littlefs doesn't actually store the page (prog_size) in the superblock, since it doesn't need to know the size for any other operation. We can work around this by storing both the CRC and size of the next page when necessary.

Another interesting note is that we don't need to any bit tweaking information, since we read the next page every time we would need to know how to clobber the erase-state CRC. And since we only read prog_size, this works really well with our caching, since the caches must be a multiple of prog_size.

This also brings back the internal lfs_bd_crc function, in which we can use some optimizations added to lfs_bd_cmp in https://github.com/littlefs-project/littlefs/pull/395.

TODO:

- [x] Fix failing tests cases
- [x] Add tests for backwards compatibility
- [x] Implement proper handling of on-disk minor version bump
- [x] Documentation update for DESIGN/SPEC

Depends on https://github.com/littlefs-project/littlefs/pull/495
Related to https://github.com/littlefs-project/littlefs/issues/245

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -473,6 +473,42 @@ jobs:
           path: status
           retention-days: 1
 
+  # run compatibility tests using the current master as the previous version
+  test-compat:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v2
+        if: ${{github.event_name == 'pull_request'}}
+      # checkout the current pr target into lfsp
+      - uses: actions/checkout@v2
+        if: ${{github.event_name == 'pull_request'}}
+        with:
+          ref: ${{github.event.pull_request.base.ref}}
+          path: lfsp
+      - name: install
+        if: ${{github.event_name == 'pull_request'}}
+        run: |
+          # need a few things
+          sudo apt-get update -qq
+          sudo apt-get install -qq gcc python3 python3-pip
+          pip3 install toml
+          gcc --version
+          python3 --version
+      # adjust prefix of lfsp
+      - name: changeprefix
+        if: ${{github.event_name == 'pull_request'}}
+        run: |
+          ./scripts/changeprefix.py lfs lfsp lfsp/*.h lfsp/*.c
+      - name: test-compat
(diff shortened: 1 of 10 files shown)


pjsg:
This looks exciting!

geky:
Just added documentation about this change to the SPEC.md:
https://github.com/littlefs-project/littlefs/pull/497/files?short_path=7426c9e#diff-7426c9e3a694ca6015df5f98637912975f2edea23270203ee89a8bdeed246ee0

Things look like they are working so this should be ready to merge soon.

Sorry again about the delay, this is the first change that involves bumping the on-disk minor version. This sort of irreversible change can lead to unpleasant situations if a mistake is made, so I wanted to make sure forward/backward compatibility is decently tested first.

116332d3f7e33e2d4ada3b4896024b1d44567593 and 4c9360020ed8b6db5d943fd61155a09553e8f30f should hopefully provide decent confidence that the on-disk minor version is respected. This will also need a good explanation in the release notes...

Issue #203, opened 2019-06-07
LFSv2: inline files performance drop


pasi-ojala-tridonic:
Running on a device with very little flash available, decided to give the inline files of v2 a go. Not sure if I got everything configured ok, tried to follow the instructions.

Creating a couple of files holding an integer, increment the value in a loop, things basically seem to work. I see lots of reads towards my flash, I assume it's the inline files that cause the need to read a lot. I could leave with that, the data seems valid, the file contents keep changing as expected.

However, after some loops, starting from line 1246 and lasting till 1446 in the log, the reads just explode when updating a value in file 8, the update taking about 25 seconds. There's an ERASE happening in the middle, but that's not taking much time compared to all the reads happening.
Two minutes later (search for ERASE), this happens with file 6, then later for file 4, then 2 and then back to 8.

I was looking around a bit in debugger: when these huge read blocks appear, the recursion in lfs_dir_traverse (https://github.com/ARMmbed/littlefs/blob/master/lfs.c#L700) seems to kick in, although I never saw the recursion go deeper than level 1.

In the test snippet, mountLittleFS() does the mount and printing out of directory contents.

[lfstest.txt](https://github.com/ARMmbed/littlefs/files/3265181/lfstest.txt)
[lfsv2-8files.txt](https://github.com/ARMmbed/littlefs/files/3265183/lfsv2-8files.txt)

geky:
Hi @pasi-ojala-tridonic, thanks for creating an issue.

This is a part of how littlefs works. At its core, littlefs is built out of a bunch of small 2-block logs (metadata pairs, similar to journals). These logs are appended with new information as the filesystem is written to. Eventually they get full and need to be garbage collected to remove outdated entries. Unfortunately this garbage collection is expensive (_O(erase size^2)_).

The main approach littlefs takes to prevent this from becoming unusable is to keep the logs small and limited to 2-erase blocks, building more scalable data structures out of the 2-block logs. But there's still a risk that this is no scalable as the erase size gets too large.

Inline files are files that are small enough we can fit them directly in their parent directory's 2-block log, instead of just storing a pointer to the file's data structure. This can lead to more data being stored in the 2-block log. It's possible limiting the size of inline files could increase the performance of garbage collection, but there's still the risk of long-running spikes as the 2-block log will still have to garbage collect at some point.

More info here:
https://github.com/ARMmbed/littlefs/blob/master/DESIGN.md#metadata-pairs

[...]

PR #502, opened 2020-12-15
Add metadata_max config to help performance on devices with large blocks

mon:
Alright, hopefully I've done this correctly. Inspired by our conversation on #203.
Not sure of the best way to validate performance "properly", if there's a test bench already setup I should be using, but I am still seeing a 10x increase in directory traversal/file deletion and a 2x-4x speedup in file reads with this code and a `metadata_max` of 4096 and `inline_file_max` of -1.

Commit message:
We have seen poor read performance on NAND flashes with 128kB blocks.
The root cause is inline files having to traverse many sets of metadata
pairs inside the current block before being fully reconstructed. Simply
disabling inline files is not enough, as the metadata will still fill up
the block and eventually need to be compacted.

By allowing configuration of how much size metadata takes up, along with
limiting (or disabling) inline file size, we achieve read performance
improvements on an order of magnitude.

### DIFF
--- a/lfs.h
+++ b/lfs.h
@@ -207,7 +207,7 @@ struct lfs_config {
     // Number of erasable blocks on the device.
     lfs_size_t block_count;
 
-    // Number of erase cycles before littlefs evicts metadata logs and moves 
+    // Number of erase cycles before littlefs evicts metadata logs and moves
     // the metadata to another block. Suggested values are in the
     // range 100-1000, with large values having better performance at the cost
     // of less consistent wear distribution.
@@ -256,6 +256,20 @@ struct lfs_config {
     // larger attributes size but must be <= LFS_ATTR_MAX. Defaults to
     // LFS_ATTR_MAX when zero.
     lfs_size_t attr_max;
+
+    // Optional upper limit on total space given to metadata pairs in bytes. On
+    // devices with large blocks (e.g. 128kB) setting this to a low size (2-8kB)
+    // can help bound the metadata compaction time. Must be <= block_size.
+    // Defaults to block_size when zero.
+    lfs_size_t metadata_max;
+
+    // Optional upper limit on inline files in bytes. On devices with large
+    // blocks (e.g. 128kB) setting this to a low size or disabling inline files
+    // can help bound file read overhead. Must be <= LFS_FILE_MAX. Defaults to
+    // block_size/8 when zero.
+    //
+    // Set to -1 to disable inline files.
+    lfs_ssize_t inline_file_max;
 };
 
 // File info structure
@@ -406,6 +420,8 @@ typedef struct lfs {
     lfs_size_t name_max;
     lfs_size_t file_max;
     lfs_size_t attr_max;
+    lfs_size_t metadata_max;
+    lfs_ssize_t inline_file_max;
 
 #ifdef LFS_MIGRATE
     struct lfs1 *lfs1;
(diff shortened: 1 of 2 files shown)


mon:
Also, as far as backwards compat/superblock inclusion goes, I've been able to move backwards and forwards between custom/default values with no data loss, errors or asserts. I don't know the codebase well enough to be sure, but I believe since these new limits only come into play when deciding to compact, adding new blocks, or rewriting files, they will be silently applied only when required, and old blocks can be traversed fine.

geky:
This looks great!

The only question I have is do we need `inline_file_max` if we have `metadata_max`? It is currently based on `block_size/8`, but could instead be changed to `metadata_max/8`. I'm not sure a separate `inline_file_max` is useful.

lfs.h
```diff
@@ -406,6 +420,8 @@ typedef struct lfs {
     lfs_size_t name_max;
     lfs_size_t file_max;
     lfs_size_t attr_max;
+    lfs_size_t metadata_max;
```
geky, line 423: Ah, if `metadata_max` is not dependent on the superblock, it's better to not keep it in the `lfs_t` struct, but instead calculate its value on every use (`lfs->metadata_max ? lfs->metadata_max : lfs->cfg->block_size`).

Something similar is done for `block_cycles`. It's unfortunately a bit noisy, but it reduces RAM at the cost of code size, and has the potential to be optimized out in https://github.com/littlefs-project/littlefs/pull/491.

geky:
> Not sure of the best way to validate performance "properly", if there's a test bench already setup I should be using

Unfortunately not at the moment. It's on my TODO list to put together a set of benchmarks similar to the tests, and would be the first step in addressing the current performance issues.

Performance hasn't been a high priority for littlefs, and the only real basis for performance during development was theory, so the current performance issues probably shouldn't have been a surprise. I believe "you are what you test" is how the saying goes...

lfs.h
mon: Ah, I was a little worried at how sparse `lfs_t` was and whether this was the right choice. I'll modify this to calculate-at-use when I sort the inline files config.

geky:
> all our files are strictly larger than the inline file size, so I want to be able to disable them entirely

Ah, if you create a file larger than the inline file size, it should never get written out to disk as an inline file (unless there's a bug, or you call `lfs_file_sync` while the file is still inline size).

Due to other conditions with multiple open files, open inline files are stored entirely in the file cache. So in theory, when you write a file larger than cache_size, by the time littlefs needs to flush the data to disk it will have already decided the file should not be inlined.

mon:
Aha, and that explains why we had the perf issues at 128k blocks and 4k cache - the inline file limit is 16k, so it will always be written out before un-inlining.

Fix commits incoming shortly!

geky:
Looks perfect 👍 Thanks for following all the unwritten style rules.

Will bring this in on the next minor release.

geky (approved):

geky:
Assuming it passes CI ;)

But it did pass the first time

geky:
> perf issues at 128k blocks and 4k cache - the inline file limit is 16k

Hmmm, the actual check is against `min(0x3fe, cache_size, block_size/8)` (0x3fe is due to a 10-bit size field). So in theory it still should not have been written out as inlined:
https://github.com/littlefs-project/littlefs/blob/1a59954ec64ca168828a15242cc6de94ac75f9d1/lfs.c#L2966-L2969

mon:
Hmm, are directories/dir entries stored as files, and thus inline-able? That's the only other possibility if they were indeed being inlined as that 0x3fe shows.

geky:
Nope, littlefs doesn't do directory inlining. I wonder if it might have been a bug.

geky:
Thanks for the PR!

PR #511, opened 2020-12-24
Skip flushing file if lfs_file_rawseek() doesn't change position

### LINKED ISSUE #436: Seeking to the current file offset is inefficient
If you call `lfs_file_seek` with `LFS_SEEK_SET` and the current offset of the file, it still [unconditionally flushes](https://github.com/ARMmbed/littlefs/blob/4c9146ea539f72749d6cc3ea076372a81b12cb11/lfs.c#L3009) the file.

In my case, I'm using littlefs with a custom VFS wrapper, so I reset littlefs' file offset each time I write to the file. This "feature" resulted in significant performance loss.

Currently, I've worked around this by calling `lfs_file_tell` before I set the new offset. I'd be happy to send a pull request that:

* moves the flushing to after the new file position is calculated, and
* adds a check for whether the new position will be the same

embeddedt:
This fixes #436.

@geky It appears that your concerns were valid, as one of the truncation test cases is now failing with this change:

```
====== results ======
tests/test_truncate.toml:96:failure: test_truncate#3#1 failed with -6
tests/test_truncate.toml:127:assert: assert failed with 48, expected eq 0
    memcmp(rb, wb, trunc) => 0;
```

Do you have any ideas about how to go about resolving this?

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -3048,14 +3048,6 @@ relocate:
 
 static lfs_soff_t lfs_file_rawseek(lfs_t *lfs, lfs_file_t *file,
         lfs_soff_t off, int whence) {
-#ifndef LFS_READONLY
-    // write out everything beforehand, may be noop if rdonly
-    int err = lfs_file_flush(lfs, file);
-    if (err) {
-        return err;
-    }
-#endif
-
     // find new pos
     lfs_off_t npos = file->pos;
     if (whence == LFS_SEEK_SET) {
@@ -3070,6 +3062,19 @@ static lfs_soff_t lfs_file_rawseek(lfs_t *lfs, lfs_file_t *file,
         // file position out of range
         return LFS_ERR_INVAL;
     }
+
+    if (file->pos == npos) {
+        // noop - position has not changed
+        return npos;
+    }
+
+#ifndef LFS_READONLY
+    // write out everything beforehand, may be noop if rdonly
+    int err = lfs_file_flush(lfs, file);
+    if (err) {
+        return err;
+    }
+#endif
 
     // update pos
     file->pos = npos;
--- a/lfs.c
+++ b/lfs.c
@@ -3106,6 +3106,9 @@ static int lfs_file_rawtruncate(lfs_t *lfs, lfs_file_t *file, lfs_off_t size) {
             return err;
         }
 
+        // need to set pos/block/off consistently so seeking back to
+        // the old position does not get confused
+        file->pos = size;
         file->ctz.head = file->block;
         file->ctz.size = size;
         file->flags |= LFS_F_DIRTY | LFS_F_READING;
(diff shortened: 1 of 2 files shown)


geky:
The good news is that only a single test case is failing, that's quite rare.

You can see the full test case here:
[https://github.com/littlefs-project/littlefs/blob/1a59954ec64ca168828a15242cc6de94ac75f9d1/tests/test_truncate.toml#L98-L145 ](https://github.com/littlefs-project/littlefs/blob/1a59954ec64ca168828a15242cc6de94ac75f9d1/tests/test_truncate.toml#L98-L145)

Hmm, it's interesting that `lfs_file_truncate` internally calls `lfs_file_seek`. It's also interesting that this test contains a write and then a seek, and then a truncate after seeking. Since there's no longer an implicit flush I wonder if data is getting lost between these calls.

But `lfs_file_truncate` does contain an explicit flush in the shrinking case (which is what is being tested), so in theory that shouldn't happen:
[https://github.com/littlefs-project/littlefs/blob/1a59954ec64ca168828a15242cc6de94ac75f9d1/lfs.c#L3090 ](https://github.com/littlefs-project/littlefs/blob/1a59954ec64ca168828a15242cc6de94ac75f9d1/lfs.c#L3090)

Will need to investigate more what's going on

geky:
Ah! I think the issue is with LFS_SEEK_END. It's calculated based off `file->ctz.size`. 
https://github.com/littlefs-project/littlefs/blob/6bb404315405ba6f5217c061c0e3fb2a0d19099b/lfs.c#L3057-L3059

But `file->ctz.size` is only updated during flush. Previously the implicit flush updated `file->ctz.size` before the seek offset is calculated.

Compare to lfs_file_rawsize, which finds the size on unflushed files:
https://github.com/littlefs-project/littlefs/blob/6bb404315405ba6f5217c061c0e3fb2a0d19099b/lfs.c#L3154-L3164

The easiest solution may be just to call lfs_file_rawsize in lfs_file_rawseek, but I'll leave that up to you.

---

Also it's interesting to note that we no longer need this check in lfs_file_truncate, since lfs_file_rawseek detects noops. Previously it was to avoid the extra flush:
https://github.com/littlefs-project/littlefs/blob/1a59954ec64ca168828a15242cc6de94ac75f9d1/lfs.c#L3109-L3114

embeddedt:
Hmm... I've tried adding that fix but the test still seems to be failing, unless the CI isn't picking up on my latest changes.

geky:
Had to pull up the debugger for this one. It turned out to be a complicated interaction between lfs_file_truncate and lfs_file_seek where lfs_file_truncate wasn't setting `file->pos` correctly, causing the seek-noop to happen while the file metadata was in a bad state.

lfs_file_truncate expected lfs_file_seek to unconditionally set file->pos, which isn't true any more.

I just put what I found in the commit message (745d98c), but let me know it leaves you with any questions.

Off topic, but I've also ported the CI to Github Actions (https://github.com/littlefs-project/littlefs/pull/516), which hopefully means less of a wait for CI in the future.

embeddedt:
Your fix makes sense; thanks!

geky:
Thanks for the PR!

embeddedt:
Not a problem. Thank you for continuing to develop littlefs! :+1:

Issue #503, opened 2020-12-15
Thoughts: allowing LFS_ASSERT to fail without abort


mon:
Had an idea for a PR, but I won't go to the trouble if you think it's not merge-able.

Currently there are two ways to use `LFS_ASSERT`: as a no-op with `LFS_NO_ASSERT`, or as a "kill the entire program" with `assert`. I propose a third option: return `LFS_CORRUPT` and log the error when it occurs, instead of taking down the entire system.

Since LFS_ASSERT is used in functions returning both int and void, I propose adding `...` to LFS_ASSERT, so I can, for example:
`#define LFS_ASSERT(test, ...) if(!(test)) { log_err("%s", #test); return __VA_ARGS__; }`

Alternatively, perhaps splitting the assert into one that returns a value and one that doesn't could work.

geky:
My main thought is that it's preferable to find corrupted disk structures at mount/fetch/read time. If, say, `lfs_bd_prog` is called with an out of range block address, that means some logic called `lfs_bd_prog` without checking the address bounds, and that logic should be fixed.

That being said, until we ensure littlefs catches all invalid metadata, this proposal is better than the current state of things.

So I'd be happy to bring in this sort of change, but will also work to replace it in some unknown time frame.

> Alternatively, perhaps splitting the assert into one that returns a value and one that doesn't could work.

I think a second return assert is preferable, it would be better for the control flow to be explicit in case a later rewrite of the function needs to handle cleanup resources.

> log the error when it occurs

There's also the option of logging the assert failure and then trying to continue, though this may just cause more problems than it is worth.

davidefer:
Hi
I find the proposal of mon very useful. We have the same problem, an assert in lfs would bring the system down.
We would like very much to perform a sort of graceful degradation in this case. And the error logging is also an issue for us.

mon:
With the merging of #514, sorted!

PR #514, opened 2021-01-04
lfs_fs_preporphans: return int to alllow graceful LFS_ASSERT

mon:
From the discussion in #503, I started to implement the split `LFS_ASSERT`/`LFS_ASSERT_RETURN` functionality, but realised that all but one assert already exists in a function returning `int` or `lfs_ssize_t`.

Thus, by changing this function to return `int`, I can continue to use my custom LFS_ASSERT impl:
`#define LFS_ASSERT(test) do {if(!(test)) {log_error("lfs ASSERT %s:%d ", __FILE__, __LINE__); return LFS_ERR_ASSERT;}} while(0)`

Though there aren't any checks in place to prevent newly written LFS_ASSERT-using functions from returning `void`, I added `LFS_ERR_ASSERT` to the `lfs_error` enum to hopefully encourage consistency. Happy to revert that part of the commit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -224,6 +224,22 @@ jobs:
     # report-size will compile littlefs and report the size
     script: [*report-size]
 
+  # test compilation with asserts that return -1
+  - stage: test
+    env:
+      - NAME=littlefs-assert-return
+      - CC="arm-linux-gnueabi-gcc --static -mthumb"
+      - CFLAGS="-Werror -D'LFS_ASSERT(test)=do {if(!(test)) {return -1;}} while(0)'"
+    if: branch !~ -prefix$
+    install:
+      - *install-common
+      - sudo apt-get install
+            gcc-arm-linux-gnueabi
+            libc6-dev-armel-cross
+      - arm-linux-gnueabi-gcc --version
+    # report-size will compile littlefs and report the size
+    script: [*report-size]
+
   # test compilation in thread-safe mode
   - stage: test
     env:
--- a/lfs.c
+++ b/lfs.c
@@ -465,7 +465,7 @@ static int lfs_file_rawsync(lfs_t *lfs, lfs_file_t *file);
 static int lfs_file_outline(lfs_t *lfs, lfs_file_t *file);
 static int lfs_file_flush(lfs_t *lfs, lfs_file_t *file);
 
-static void lfs_fs_preporphans(lfs_t *lfs, int8_t orphans);
+static int lfs_fs_preporphans(lfs_t *lfs, int8_t orphans);
 static void lfs_fs_prepmove(lfs_t *lfs,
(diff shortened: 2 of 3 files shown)


geky:
Hi @mon, thanks for creating a PR.

Lets go with this. I've rejected previous proposals like this in the past, so hopefully no one else reading this is frustrated by the change in direction, but from the feedback this change has too much practical value.

So from this point on LFS_ASSERT will only be used in littlefs in functions that can return an int if users override the macro. Users are still free to use the macro in non-int functions, but they lose this ability. Though I guess they can probably just use their own assert macro in most cases.

> Though there aren't any checks in place to prevent newly written LFS_ASSERT-using functions from returning `void`

Ah! The way to handle this is a CI job that tests compilation with your error-returning assert. You could probably copy the readonly-compilation CI job and change the `-DLFS_READONLY` to `-D'LFS_ASSERT(test) do {if(!(test)) {log_error("lfs ASSERT %s:%d ", __FILE__, __LINE__); return LFS_ERR_ASSERT;}} while(0)'`. It's not the prettiest but apparently works ([according to this](https://stackoverflow.com/questions/31857559/gcc-define-function-like-macros-using-d-argument)):
https://github.com/littlefs-project/littlefs/blob/1a59954ec64ca168828a15242cc6de94ac75f9d1/.travis.yml#L211-L225

> I added `LFS_ERR_ASSERT` to the `lfs_error` enum to hopefully encourage consistency. Happy to revert that part of the commit.

Hmm, I've been been sticking to Linux errno codes since they offer a well-tested set of standard error codes that cover most system operations ([unix_system_error](https://www-numi.fnal.gov/offline_software/srt_public_context/WebDocs/Errors/unix_system_errors.html)). Unfortunately there isn't one for "ASSERT". I like using -1, which I've previously repurposed to a general "unknown" error code, though I've also seen EINVAL and EFAULT returned from assert failures.

I think it would be better to leave it out, since it does run the risk of confusing users on the default behavior of asserts. LFS_ERR_INVAL specifically may be preferred by users since littlefs also uses asserts to catch bad file configurations.

mon:
Alright! I've added that test, not using readonly since more assert-using code is built with the normal compile.

Have also reverted the new constant - I took EDOOFUS (Programming error) from BSD, but I think letting the user choose -1, EINVAL or similar works better.

geky:
>  took EDOOFUS (Programming error) from BSD

Oh that is a fun one! Unfortunately -88 is the less exciting ENOTSOCK (footwear is shoe) in Linux. For better or worse it also looks like some of the error codes are already Linux specific (ENOATTR looks different in BSD). It would have been a beautiful world if POSIX defined the actual integer values, but oh well.

---

Assuming CI passes, this looks perfect 👍. Thanks again for the PR and I will merge this into the next minor release.

geky (approved):

PR #529, opened 2021-02-22
scripts/test.py: Fix infinite busy loops on macOS

yamt:
I confirmed that the same number of tests are run
with "make test" on:

    * Ubuntu with and without this change
    * macOS with this change

>   ====== results ======
>   tests passed 817/817 (100.00%)
>   tests failed 0/817 (0.00%)


pjdennis:
Please merge this pull request. I just spent an hour troubleshooting this same issue on Mac and reached the exact same code change as a fix... before discovering this pull request already exists.

geky (approved):

PR #550, opened 2021-04-28
A better algorithm for ctz

shscp:
x*2^k equivalent to x<<k . Construct a number 0x0xFB5CA62, it have a features: when it left shift 0, left shift 1... left shift 31, it‘s most 5 significant bit just take all numbers from 0 to 31.


geky:
Hmmm, the next step would be to run the CI benchmarks to see if this is worth taking in, but it seems CI doesn't want to run on this PR for some reason... debugging...

PR #567, opened 2021-06-12
Patch release: v2.4.1

geky:
Small patch release here thanks to @yamt and @tniessen. No changes to the littlefs core.

I probably should've just bypassed the devel branch, but still fiddling with the workflow.


geky:
Note to self: devel is still still a good intermediary stage, but creating a PR is not necessary (extra step and forces an extra merge commit). Just push devel to master for patch release.

PR #569, opened 2021-06-14
Fix compiler warnings when LFS_READONLY defined

tniessen:
Without this patch, when defining `LFS_READONLY`:

```
littlefs/lfs.c:14:20: warning: unused function 'lfs_cache_drop' [-Wunused-function]
static inline void lfs_cache_drop(lfs_t *lfs, lfs_cache_t *rcache) {
                   ^
littlefs/lfs.c:271:20: warning: unused function 'lfs_pair_sync' [-Wunused-function]
static inline bool lfs_pair_sync(
                   ^
littlefs/lfs.c:283:20: warning: unused function 'lfs_pair_tole32' [-Wunused-function]
static inline void lfs_pair_tole32(lfs_block_t pair[2]) {
                   ^
littlefs/lfs.c:368:20: warning: unused function 'lfs_gstate_hasorphans' [-Wunused-function]
static inline bool lfs_gstate_hasorphans(const lfs_gstate_t *a) {
                   ^
littlefs/lfs.c:372:23: warning: unused function 'lfs_gstate_getorphans' [-Wunused-function]
static inline uint8_t lfs_gstate_getorphans(const lfs_gstate_t *a) {
                      ^
littlefs/lfs.c:376:20: warning: unused function 'lfs_gstate_hasmove' [-Wunused-function]
static inline bool lfs_gstate_hasmove(const lfs_gstate_t *a) {
                   ^
littlefs/lfs.c:391:20: warning: unused function 'lfs_gstate_tole32' [-Wunused-function]
static inline void lfs_gstate_tole32(lfs_gstate_t *a) {
                   ^
littlefs/lfs.c:419:20: warning: unused function 'lfs_superblock_tole32' [-Wunused-function]
static inline void lfs_superblock_tole32(lfs_superblock_t *superblock) {
                   ^
8 warnings generated.
```

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -11,12 +11,15 @@
 #define LFS_BLOCK_INLINE ((lfs_block_t)-2)
 
 /// Caching block device operations ///
+
+#ifndef LFS_READONLY
 static inline void lfs_cache_drop(lfs_t *lfs, lfs_cache_t *rcache) {
     // do not zero, cheaper if cache is readonly or only going to be
     // written with identical data (during relocates)
     (void)lfs;
     rcache->block = LFS_BLOCK_NULL;
 }
+#endif
 
 static inline void lfs_cache_zero(lfs_t *lfs, lfs_cache_t *pcache) {
     // zero to avoid information leak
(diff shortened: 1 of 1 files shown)


geky:
Thanks! Would be good to have this in. Will bring this in next patch release (may be a bit longer).

Out of curiosity, which version of which compiler are you using? I believe both gcc and clang treat `static inline` functions specially since they are common in header files (otherwise CI should have caught this).

geky (approved):

tniessen:
> which version of which compiler are you using?

I am using `clang+llvm-10.0.0-x86_64-linux-gnu-ubuntu-18.04`, so Clang 10.

> treat `static inline` functions specially since they are common in header files (otherwise CI should have caught this).

I'd imagine that clang warns because the definitions are in a `.c` file, which means that they are definitely dead code. If these were in the header file, they'd still be dead code, but Clang probably wouldn't warn.

geky:
Ah interesting, it looks like gcc doesn't have the same check. But nonetheless this is a good change to have.

geky:
Thanks for the PR!

PR #580, opened 2021-07-13
Add attribute operation, use file parameter

DS-LK:
I found that in the process of using littlefs, sometimes the save path is not obtained or obtained, and the custom attribute cannot be used at this time. I think it is better to add a file parameter, it allows us to apply more scenarios, and can also save memory usage in some cases, and this function is also helpful for nuttx's vfs

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -5265,6 +5265,58 @@ lfs_soff_t lfs_file_size(lfs_t *lfs, lfs_file_t *file) {
     return res;
 }
 
+lfs_ssize_t lfs_file_getattr(lfs_t *lfs, lfs_file_t *file,
+        uint8_t type, void *buffer, lfs_size_t size)
+{
+    int err = LFS_LOCK(lfs->cfg);
+    if (err) {
+        return err;
+    }
+    LFS_TRACE("lfs_file_setattr(%p, %p)", (void*)lfs, (void*)file);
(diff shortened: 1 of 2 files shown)


dpkristensen:
Would love for this feature to be merged.

geky:
Hi @DS-LK, thanks for creating a PR, sorry about the extremely late response. I should have commented on this earlier.

This functionality is a great idea, but unfortunately I don't think it fits into how littlefs currently models open files.

Consider:

``` c
// attr in RAM
struct lfs_attr example_attrs[1] = {
    {.type=1, .buffer=&(uint8_t[1]){}, .size=1}
};
struct lfs_file_config file_config = {.attrs=example_attrs, .attr_count=1};

lfs_file_t file;
lfs_file_opencfg(lfs, &file, "example_file", LFS_O_RDWR, &file_config);

// write RAM attr
memset(example_attrs[0].buffer, "a", 1);

// write file attr
lfs_file_setattr(lfs, &file, "b", 1);

// close file 
lfs_file_close(lfs, &file);
```

With this implementation the file attr will be set to "a".

A bit less clear is that this implementation also immediately writes the attrs to disk, where as other `lfs_file_*` functions stage changes to be written until the next `lfs_file_sync`.

There may be ways of making this work where `lfs_file_setattr` only updates RAM-backed attrs or errors, but I think a better solution would be to address the core problem @dpkristensen highlighted in #759. i.e. make it so RAM-backed attrs are always fetched during `lfs_file_open`.

Issue #349, opened 2019-12-11
Automatic volume recognition


davidefer:
Hi everybody
I was wondering whether lfs could automatically detect the volume setup when mounting, without specifying the volume configuration in the lfs_config structure.
Would it be possible to store somewhere in the volume this information while formatting? Or at least all the information needed to detect the volume topology and let it me mounted. 
The only information to be hardcoded in the fw  would be the start address of lfs in the flash.
Thanks in advance.

geky:
Hi @davidefer, this has been on my mind for a while.

littlefs does have a "superblock entry" that contains static info about the filesystem (can you can add your own with `lfs_setattr("/")`, though I realize this doesn't help _before_ you mount the filesystem).

Sidenote: Most of the configuration isn't needed for portability.
- cache_size - Does nothing with disk.
- read_size/prog_size - As long as these are compatible with the underlying disk, they are only used for cache-alignment and can change between systems.
- block_cycles - Only used to determine when to evict blocks, can change.

littlefs only needs to know two things: The block_size and block_count.

The block_count is easy, we can read that from the superblock entry.

The block_size, however, present a chicken-and-egg problem I haven't found a good answer for yet.

The root/superblock in littlefs is a small two-block log in blocks 0 and 1. This means that at any given time one of the two blocks may be in the middle of an erase operation. For most flash, this means temporarily setting the flash to 1s, losing any data on the block. Handling is a hard requirement for power-resilience.

So we have a superblock in one of the two blocks at the start of storage. If we can get to one of the two blocks, we can fetch the configuration. But this is where we run into a problem. How do we load the second block if we don't know our block size? We only need the offset (not the size) to fetch a block, but without knowing the size, we don't know where the second block could be.

I'm very interested in solving this problem. But have no idea how to.

---

Some possible solutions:
1. Provide configuration-detection as a "best-effort" operation, allowing for a possible failure if the device was caught erasing the superblock. I really don't like this "solution" as it could create rare problems for unaware-users and can't be relied on for any sort of real product
2. Reserve one permanently-static block to store the superblock. This is a problem for small devices, such as internal flash, that only have two blocks in total. It could be an optional feature, though that doesn't mesh well with a consistent user-interface for storage. It's also costly.
[...]

Issue #279, opened 2019-08-31
question: is it possible to "grow" the filesystem?


mdevan:
If lfs is unmounted, and then the underlying block device size increases, is it possible to remount such that lfs can make use of the entire space? Like resize2fs or xfs_growfs, basically.

lsilvaalmeida:
#238

geky:
Short answer is: lfs_fs_resize is a good idea.

littlefs uses a garbage collection scheme, so if you increase the block device everything will work fine. But, as @ksstms pointed out in #238, there is a pesky block_size field in the superblock that becomes out of date.

The field is also unused, so by pure coincidence, resizing the filesystem will work in the current version. But there will be problems if a future version of littlefs starts relying on this field.

I think adding lfs_fs_resize is a good idea, similar to resize2fs or xfs_growfs as you mentioned. This would also make the filesystem forward compatible if we decide to remove the garbage collection (https://github.com/ARMmbed/littlefs/issues/75).

mdevan:
Something like this, then?

```C
int lfs_fs_resize(lfs_t *lfs, lfs_size_t new_block_count) {
	int err = 0;

	// read existing superblock from disk
	lfs_superblock_t superblock;
	err = lfs_dir_get(lfs, &dir, LFS_MKTAG(0x7ff, 0x3ff, 0),
			LFS_MKTAG(LFS_TYPE_INLINESTRUCT, 0, sizeof(superblock)),
			&superblock);
	if (err < 0) {
		goto cleanup;
	}
	lfs_superblock_fromle32(&superblock);

	// size can only be increased
	if (new_block_count <= superblock.block_count) {
		LFS_ERROR("Invalid new block count %d, not > %d", new_block_count,
			superblock.block_count);
		err = LFS_ERR_INVAL;
		goto cleanup;
	}

	// update superblock block_count
	superblock.block_count = new_block_count;

	// write back the updated superblock to disk
	lfs_superblock_tole32(&superblock);
	err = lfs_dir_commit(lfs, lfs->root, LFS_MKATTRS(
			{LFS_MKTAG(LFS_TYPE_SUPERBLOCK, 0, 8), "littlefs"},
			{LFS_MKTAG(LFS_TYPE_INLINESTRUCT, 0, sizeof(superblock)),
				&superblock}));
	if (err) {
		goto cleanup;
	}

	// succeeded, use new block_count from hereon
	lfs->cfg->block_count = new_block_count;

cleanup:
	return err;
[...]

PR #584, opened 2021-07-21
Fail mount when the block size changes

colin-foster-in-advantage:
When the on-disk block size doesn't match the config block size, it is
possible to get file corruption. For instance, if the num blocks was
0x200 and we re-mount with 0x100 files could be corrupt.

If we re-mount with a larger number of blocks things should be safer,
but could be handled with a resize option or perhaps a mount flag to
ignore this parameter.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -3762,10 +3762,17 @@ static int lfs_rawmount(lfs_t *lfs, const struct lfs_config *cfg) {
                 lfs->attr_max = superblock.attr_max;
             }
 
-            if ((superblock.block_count != lfs->cfg->block_count) ||
-                (superblock.block_size != lfs->cfg->block_size))
-            {
-                err = LFS_ERR_CORRUPT;
+            if (superblock.block_count != lfs->cfg->block_count) {
+                LFS_ERROR("Invalid block count (%"PRIu32" != %"PRIu32")",
+                        superblock.block_count, lfs->cfg->block_count);
+                err = LFS_ERR_INVAL;
+                goto cleanup;
+            }
+
+            if (superblock.block_size != lfs->cfg->block_size) {
+                LFS_ERROR("Invalid block size (%"PRIu32" != %"PRIu32")",
+                        superblock.block_count, lfs->cfg->block_count);
+                err = LFS_ERR_INVAL;
                 goto cleanup;
             }
         }
--- a/lfs.c
+++ b/lfs.c
@@ -3761,6 +3761,12 @@ static int lfs_rawmount(lfs_t *lfs, const struct lfs_config *cfg) {
 
                 lfs->attr_max = superblock.attr_max;
             }
+
+            if (superblock.block_count != lfs->cfg->block_count)
+            {
+                err = LFS_ERR_CORRUPT;
+                goto cleanup;
+            }
         }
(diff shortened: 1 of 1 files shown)


preetpalkang:
The change looks good, but we should also check the `block_size` in addition to `block_count`

We should merge this and get a new release :)

colin-foster-in-advantage:
> The change looks good, but we should also check the `block_size` in addition to `block_count`

Done. In hindsight I wasn't clear about size vs count. Count is the one that is probably more likely to change, whereas size is most likely dependent on the hardware.

preetpalkang (approved):
Sorry for small scope creep as the #584 was only meant to handle changes to `block_count`.

Regardless, I think it is a good idea because we should have sanity checks to ensure that the `lfs_config` differs from the `superblock` and the read/write/erase functions are not going to work, hence we want to force the user to reformat.

This is a great change until something like #279 is implemented. I'd recommend merging to the master.

lfs.c
```diff
@@ -3761,6 +3761,13 @@ static int lfs_rawmount(lfs_t *lfs, const struct lfs_config *cfg) {
 
                 lfs->attr_max = superblock.attr_max;
             }
+
+            if ((superblock.block_count != lfs->cfg->block_count) ||
+                (superblock.block_size != lfs->cfg->block_size))
+            {
```
preetpalkang, line 3767: nit picking: brace should be on previous line for consistency of code base.

colin-foster-in-advantage: Agreed. Fixed below. I'm not sure how well github works with squashing commits. Let me know if there's anything I should do for that.

preetpalkang: looks good!

@maintainer let's approve and merge this :)

geky:
This looks reasonable to me, thanks for creating this PR. Sorry about the delay on my end, doing a bit of catch-up.

lfs.c
```diff
@@ -3761,6 +3761,12 @@ static int lfs_rawmount(lfs_t *lfs, const struct lfs_config *cfg) {
 
                 lfs->attr_max = superblock.attr_max;
             }
+
+            if ((superblock.block_count != lfs->cfg->block_count) ||
+                (superblock.block_size != lfs->cfg->block_size)) {
+                err = LFS_ERR_CORRUPT;
```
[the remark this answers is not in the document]
colin-foster-in-advantage: Thanks for the feedback. I made these changes and pushed them to my branch. I can squash the commits if you prefer that.

geky:
Thanks for updating this based on the feedback, looks good and will bring this in on the next minor release.

I agree with @dpgeorge and @preetpalkang, this is a good protection to have, at least until https://github.com/littlefs-project/littlefs/issues/279 is resolved (though a shrinking filesystem will probably always be a hard error).

---

> whereas size is most likely dependent on the hardware.

Quick side note, but the `block_size` can actually be a multiple of the physical erase size, which you might do for performance reasons. So this should be treated roughly the same as the `block_count`, but determining both at mount time has some problems  (https://github.com/littlefs-project/littlefs/issues/349)

geky (approved):

geky:
Thanks for putting together this PR!

PR #610, opened 2021-11-08
New API function lfs_find_free_blocks. 

opilat:
It performs a free block search which later will not slow down a write operation.

It helps to have predictable write times. In scenarios like storing data from a microphone. Write taken too long when littlefs does the free block search during it.

I am open to other solutions.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -662,7 +662,9 @@ static int lfs_alloc(lfs_t *lfs, lfs_block_t *block) {
 }
 
 int lfs_find_free_blocks(lfs_t *lfs){
-    lfs->free.off = (lfs->free.off + lfs->free.size)
+    // Move free offset at the first unused block (lfs->free.i)
+    // lfs->free.i is equal lfs->free.size when all blocks are used
+    lfs->free.off = (lfs->free.off + lfs->free.i)
         % lfs->cfg->block_count;
     lfs->free.size = lfs_min(8*lfs->cfg->lookahead_size, lfs->free.ack);
     lfs->free.i = 0;
--- a/lfs.c
+++ b/lfs.c
@@ -654,19 +654,26 @@ static int lfs_alloc(lfs_t *lfs, lfs_block_t *block) {
             return LFS_ERR_NOSPC;
         }
 
-        lfs->free.off = (lfs->free.off + lfs->free.size)
-                % lfs->cfg->block_count;
-        lfs->free.size = lfs_min(8*lfs->cfg->lookahead_size, lfs->free.ack);
-        lfs->free.i = 0;
-
-        // find mask of free blocks from tree
-        memset(lfs->free.buffer, 0, lfs->cfg->lookahead_size);
-        int err = lfs_fs_rawtraverse(lfs, lfs_alloc_lookahead, lfs, true);
-        if (err) {
-            lfs_alloc_drop(lfs);
+        int err = lfs_find_free_blocks(lfs);
+        if(err) {
             return err;
         }
     }
+}
+
+int lfs_find_free_blocks(lfs_t *lfs){
+    lfs->free.off = (lfs->free.off + lfs->free.size)
+        % lfs->cfg->block_count;
+    lfs->free.size = lfs_min(8*lfs->cfg->lookahead_size, lfs->free.ack);
+    lfs->free.i = 0;
+
+    // find mask of free blocks from tree
+    memset(lfs->free.buffer, 0, lfs->cfg->lookahead_size);
+    int const err = lfs_fs_rawtraverse(lfs, lfs_alloc_lookahead, lfs, true);
+    if (err) {
+        lfs_alloc_drop(lfs);
+    }
+    return err;
 }
 #endif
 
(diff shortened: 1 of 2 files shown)


geky:
Sorry about leaving this open for so long. It arrived at an inconvenient time and ended up flying under my radar.

I've been look at other ways of approaching this problem, but that's not an excuse to not merge this when it's likely immediately useful to users.

Thoughts on renaming this function to `lfs_fs_gc` with the intention of adding more garbage-collection functionality (persistence, metadata compaction, etc) in the future?

The only concern I can think of is that it may be preferable to have an incremental/interruptable garbage collector, but that would be much more involved, and we can always introduce a new API for this in the future.

---

I realize this is way past the statute of limitations for open PRs, so I can go ahead and take it over. This needs to be updated for consistent code style (4-space indentions, 80-column width) and needs at least minimal tests.

geky:
After more thought, I don't think this will make it in to this minor release. It's a good idea, but needs a bit of work:
- I don't think lfs_find_free_blocks/lfs_fs_gc should error with LFS_ERR_NOSPC. It's possible blocks will be freed before the next allocation, and this error is likely to be raised when a user hasn't tested for it.
- lfs_find_free_blocks/lfs_fs_gc should only shift the lookahead buffer by the first clear bit so we don't skip any found free blocks in the next lookahead scan. The current code shifts the lookahead buffer by a whole lookahead_size because it is only called when the lookahead buffer is exhausted.
- The code should be reorganized so `lfs_alloc` calls some sort of `lfs_alloc_scan` function that lfs_find_free_blocks/lfs_fs_gc also calls so that big of logic is reused in both cases.

And the current minor release has already been delayed enough.

With a number of ideas around garbage-collection in https://github.com/littlefs-project/littlefs/issues/791, I think it would make more sense to look at this as a piece of a larger set of garbage-collection work.

geky:
Sorry, I did write that confusingly.

Consider a new filesystem with say 32 blocks and a lookahead of 8 bits:

```
in-use
| .- free
v v
11000000_00000000_00000000_00000000
  ^- free.i
'---+---'
    '- free.off .. free.size
```

In the current implementation, calling lfs_find_free_blocks moves `free.i` here:

```
11000000_00000000_00000000_00000000
         ^- free.i
         '---+---'
             '- free.off..free.size
```

But it should move here:

```
11000000_00000000_00000000_00000000
  ^- free.i
  '----+---'
       '- free.off..free.size
```

Otherwise we miss free blocks. The implementation as-is isn't broken in the sense that it gives incorrect blocks, but:

1. It will mess with wear-leveling a bit.

2. We probably want `lfs_find_Free_blocks` to be a noop if called multiple times. Since it won't really make "progress" on repeated calls.

I think changing it to something like this would work:

``` c
lfs->free.off = (lfs->free.off + lfs->free.i) % lfs->cfg->block_count;
```

opilat:
What if we allow users to pick a strategy? Your proposal is consistent in repeated calls and makes let's say necessary shift. 

It may work very badly in specific cases like this (I am not sure whether and how frequently it may happen):
```
in-use
| .- free
v v
11011111_11100000_00000000_00000000
  ^- free.i
'---+---'
    '- free.off .. free.size
```
If this case is low probable or never may exist I am going to update the equation to your proposal. 

Why doesn't the code use the _free.i_ directly it should be equal to _free.size_ when all block are used?

geky:
That is good example of where the lookahead buffer performs poorly.

But I think that's more a flaw the  approach of the lookahead buffer as a whole.

This example works poorly in both strategies:

```
in-use
| .- free
v v
11011111_11011111_11000000_00000000
  ^- free.i
'---+---'
    '- free.off .. free.size
```

> Why doesn't the code use the free.i directly it should be equal to free.size when all block are used?

I don't think there was a reason. The previous code just wasn't written to handle `free.i != free.size` there.

opilat:
The new function is updated to move the lookahead at the first free block. Would you like something else to be done?

geky:
Hi @opilat, sorry, I somehow missed your last comment's question.

TODO list for this PR, from my perspective:
1. Add some testing over this function.
2. Rename to `lfs_fs_gc` to lay groundwork for future improvements, unless there is a good reason not to.
3. Non-functional nitpicks to keep the codebase consistent:
   1. Adopt consistent style (mainly 80-column limit I think).
   2. Move behind LFS_READONLY ifdef, I don't think there's a reason to make this function available in readonly mode.
   3. Add API boilerplate (LFS_TRACE, LFS_LOCK, there's a common pattern at the bottom of the lfs.c)

I went ahead and made these changes [here](https://github.com/ozobot/littlefs/compare/New-function-lfs_find_free_blocks...littlefs-project:littlefs:fs-gc), but forgot I don't have push access to this PR:
https://github.com/ozobot/littlefs/compare/New-function-lfs_find_free_blocks...littlefs-project:littlefs:fs-gc

Are you able to either grant push access or merge the commits in the above?

If you're ok with these changes than this looks good to me for the next minor release.

geky:
Hi @opilat, I've gone ahead and created #875 so this makes it into the next minor release.

Let me know if you see any issues with it, and thanks for the original PR.

opilat:
Thank you for the updates. [#875](https://github.com/littlefs-project/littlefs/pull/875) looks good to me. Should I close this PR?

geky:
Sure thing, #875 should be coming in with https://github.com/littlefs-project/littlefs/pull/877 soon. Thanks again for the PR!

PR #613, opened 2021-11-17
fix a missing cache buffer in lfs_filerawopencfg()

nnayo:


lfs.c
```diff
@@ -2584,6 +2585,12 @@ static int lfs_file_rawopencfg(lfs_t *lfs, lfs_file_t *file,
 
     // zero to avoid information leak
     lfs_cache_zero(lfs, &file->cache);
+#else
+   // local buffer to be shared among all calls
+   static uint8_t no_malloc_buffer[1024];
```
tniessen, line 2590: I believe that the buffer should have a size of `lfs->cfg->cache_size` bytes, which is probably much less than 1024.

Also, as far as I can tell, this would mean that all files (and worse, all file systems) use the same buffer, which is inherently unsafe in multi-tasking (or multi-threaded) environments.

tniessen (commented):
What is the benefit of a static buffer over setting `file->cfg->buffer` to point to memory allocated by the user?

lfs.c
nnayo: sorry, you're right
I've pushed this commit in a rush which is bad
you should through it away
I'll provide a better one

geky:
Is this superseded by https://github.com/littlefs-project/littlefs/pull/614?

nnayo:
yes
reject this PR

PR #614, opened 2021-11-17
don't use lfs_file_open() when LFS_NO_MALLOC is set

nnayo:
fixing my previous fix

just prevent the use of lfs_file_open() when LFS_NO_MALLOC is set.
because it fails with the return code LFS_ERR_NOMEM due to lfs_malloc() returning NULL in lfs_file_rawopencfg (lfs.c:2578).

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -5082,6 +5082,7 @@ int lfs_removeattr(lfs_t *lfs, const char *path, uint8_t type) {
 }
 #endif
 
+#ifndef LFS_NO_MALLOC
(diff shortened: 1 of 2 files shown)


geky:
Hi @nnayo, thanks for this. Removing `lfs_file_open` from the LFS_NO_MALLOC APIs is a good idea.

Now just to figure out why CI is not running on this PR...

geky (approved):

geky:
It looks like I had to rebase the branch for whatever reason, GitHub Actions is a bit flaky sometimes. Sorry if this caused any inconvenience.

geky:
Thanks for putting together this PR!

PR #620, opened 2021-11-30
fix bug:lfs_alloc will alloc one block repeatedly in multiple split

XinStellaris:
BUG CASE:Assume there are 6 blocks in littlefs, block 0,1,2,3 already allocated. 0 has a tail pair of {2, 3}. Now we try to write more into 0.
When writing to block 0, we will split(FIRST SPLIT), thus allocate block 4 and 5. Up to now , everything is as expected.
Then we will try to commit in block 4, during which split(SECOND SPLIT) is triggered again(In our case, some files are large, some are small, one split may not be enough).  Still as expected now.
BUG happens when we try to alloc a new block pair for the second split:
As lookahead buffer reaches the end , a new lookahead buffer will be generated from flash content, and block 4, 5 are unused blocks in the new lookahead buffer because they are not programed yet. HOWEVER, block 4,5 should be occupied in the first split!!!!!  The result is block 4,5 are allocated again(This is where things are getting wrong).

commit ce2c01f results in this bug. In the commit, a lfs_alloc_ack is inserted in lfs_dir_split, which will cause split to reset lfs->free.ack to block count.
In summary, this problem exists after 2.1.3.

Solution: don't call lfs_alloc_ack in lfs_dir_split.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -1509,7 +1509,6 @@ static int lfs_dir_split(lfs_t *lfs,
         lfs_mdir_t *dir, const struct lfs_mattr *attrs, int attrcount,
         lfs_mdir_t *source, uint16_t split, uint16_t end) {
     // create tail directory
-    lfs_alloc_ack(lfs);
     lfs_mdir_t tail;
     int err = lfs_dir_alloc(lfs, &tail);
     if (err) {


geky:
Hi @XinStellaris, thanks for the PR and tracking down this bug. I'm really curious if this passes CI or if a different solution is needed. If it passes CI I'm happy to take it in.

Unfortunately due to some GitHub flakiness I had to touch your commit to trigger CI. Sorry if that causes any problems.

---

I'm also planning to improve how we test different block device geometries, we should have better testing of small block devices but it's difficult in the current test framework since all tests need to share a common block device.

geky:
Looks like CI is happy with it, so this should be good to go. Thanks for the PR!

geky (approved):

XinStellaris:
Thanks for the help of triggering CI for me. I did forget to trigger it.
I am glad to hear that my bug fix has passed.



---Original---
From: "Christopher ***@***.***&gt;
Date: Mon, Mar 21, 2022 12:05 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [littlefs-project/littlefs] fix bug:lfs_alloc will alloc oneblock repeatedly in multiple split (PR #620)




 
Looks like CI is happy with it, so this should be good to go. Thanks for the PR!
 
—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

geky:
It should have triggered automatically, but GitHub disabled it for a time due to site-wide crypto-mining issues. CI should run automatically on all new PRs now again as long as the account is a certain amount of time old.

PR #630, opened 2022-01-11
add the limit, the cursor cannot be set to a negative number

Johnxjj:


lfs.c
```diff
@@ -3064,11 +3064,26 @@ static lfs_soff_t lfs_file_rawseek(lfs_t *lfs, lfs_file_t *file,
     // find new pos
     lfs_off_t npos = file->pos;
     if (whence == LFS_SEEK_SET) {
-        npos = off;
+        if (off < 0) {
+            return LFS_ERR_INVAL;
+        }
+        else {
```
geky, line 3070: These squiggly brackets should be on the same line for consistency
```
} else {
```

lfs.c
```diff
@@ -3064,11 +3064,26 @@ static lfs_soff_t lfs_file_rawseek(lfs_t *lfs, lfs_file_t *file,
...
-        npos = file->pos + off;
+        if ((lfs_soff_t)file->pos + off < 0) {
+            return LFS_ERR_INVAL;
+        }
+        else {
+            npos = file->pos + off;
+        }
     } else if (whence == LFS_SEEK_END) {
-        npos = lfs_file_rawsize(lfs, file) + off;
+        if (lfs_file_rawsize(lfs, file) + off < 0) {
```
geky, line 3081: We should avoid calling `lfs_file_rawsize` multiple times, since it is actually a bit complex.

lfs.c
```diff
@@ -3064,11 +3064,26 @@ static lfs_soff_t lfs_file_rawseek(lfs_t *lfs, lfs_file_t *file,
     // find new pos
     lfs_off_t npos = file->pos;
     if (whence == LFS_SEEK_SET) {
```
geky, line 3066: LFS_SEEK_SET is a bit of a special case, and should probably be documented better.

One problem with using a signed (`lfs_soff_t`) integer to represent file offsets is that this doesn't capture the full range of files stored with an unsigned file size.

Unsigned file sizes are NOT the default (`LFS_FILE_MAX` defaults to `2147483647`), and if enabled (`LFS_FILE_MAX = 4294967294`) it causes problems, but it's still possible to use with a limited set of littlefs APIs.

To help this, `lfs_file_seek` with `LFS_SEEK_SET` treats its offset as unsigned, even though it is a passed as a signed integer. So `off < 0` in this case isn't an error, it should set the file to an extremely large value.

This does risk some confusing cases if `lfs_file_seek` is called with a negative value, but most `LFS_SEEK_SET` operations are usually calculated using absolute, unsigned values anyways.

geky (commented):
Hi @Johnxjj, thanks for the PR

This math gets very tricky because we are dealing with both signed and unsigned values.

These checks are a good idea, but I've left some comments that need to be addressed. Sorry that there are quite a few comments.

Johnxjj:
@geky I don't know why it fails(test/coverage), can you help me take a look?

geky:
Hey @Johnxjj, sorry about the late response. I think GitHub is just being flaky, it has had some issues this past month. Nothing looks wrong in your PR.

I think because the CI technically "runs in your fork", I can't restart it, but you might be able to if you see a "re-run all jobs" button in the PR's "checks" tab?

If not, touching the last commit should re-run the CI:
```
git commit --amend --no-edit
git push origin dev-johnxjj --force
```

geky:
I went ahead and touched the commit to restart CI to see if that fixes it, sorry if this causes any issues.

geky:
I also went ahead and removed the check when using LFS_SEEK_SET, due to https://github.com/littlefs-project/littlefs/pull/630#discussion_r810581243

geky:
Thanks for the PR!

Issue #631, opened 2022-01-16
Performance improvement for lfs_file_rawseek


robekras:
A while ago, there has been introduced a performance improvement in lfs_file_rawseek.
That was for the case if the new seek position is the current position ([see here](https://github.com/littlefs-project/littlefs/commit/6bb404315405ba6f5217c061c0e3fb2a0d19099b#diff-df639af987b6c5baffcccf1f2e1e63e1175837fc826981494bdea1d9476a3f01))
That means that the cache doesn't need to be invalidated as the needed data (at the seek position) is already available within the block cache.
But that's not only true for that single position (new seek pos is current file pos), but for all seek positions that belongs to the cached block data.
It's only a simple calculation to check whether the new seek position is a position which matches the cached block.
At least if the new seek position is after the current file position. For positions before the current file position it's not really clear, because there might be some data at the beginning of the block cache, which do not belong to the file data.

For example, when reading a binary font file with about 4000 glyphs (16 bits high) in lvgl, The reading tooks about 20 seconds. With the enhanced lfs_file_rawseek function the reading tooks only 2 seconds.

geky:
Ah, this is quite a good optimization, thanks for creating the issue+PR. I hadn't really considered seeks < cache_size, but it makes sense that you'd want to take advantage of cached data in this situation.

PR #632, opened 2022-01-16
Fix lfs_file_rawseek performance issue

robekras:
This should fix the performance issue if a new seek position belongs to currently cached data.
This avoids unnecessary rereads of file data.

See [issue #631](https://github.com/littlefs-project/littlefs/issues/631)

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -3091,17 +3091,23 @@ static lfs_soff_t lfs_file_rawseek(lfs_t *lfs, lfs_file_t *file,
         return npos;
     }
 
-    // Get the difference between new and current file position
-    // If new position is after the current file position
-    // If new position belongs to the currently cached data
-    // Set new file position,
-    // and update also the block related position
-    int offset = npos - file->pos;
-    if (offset > 0) {
-        if ((file->off + offset) < lfs->cfg->block_size) {
-            file->pos  = npos;
-            file->off += offset;
-
+    // if we're only reading and our new offset is still in the file's cache
+    // we can avoid flushing and needing to reread the data
+    if (
+#ifndef LFS_READONLY
+        !(file->flags & LFS_F_WRITING)
+#else
+        true
+#endif
+            ) {
+        int oindex = lfs_ctz_index(lfs, &(lfs_off_t){file->pos});
+        lfs_off_t noff = npos;
+        int nindex = lfs_ctz_index(lfs, &noff);
+        if (oindex == nindex
+                && noff >= file->cache.off
+                && noff < file->cache.off + file->cache.size) {
+            file->pos = npos;
+            file->off = noff;
             return npos;
         }
     }
--- a/lfs.c
+++ b/lfs.c
@@ -3091,6 +3091,21 @@ static lfs_soff_t lfs_file_rawseek(lfs_t *lfs, lfs_file_t *file,
         return npos;
     }
 
+    // Get the difference between new and current file position
+    // If new position is after the current file position
+    // If new position belongs to the currently cached data
+    // Set new file position,
+    // and update also the block related position
+    int offset = npos - file->pos;
+    if (offset > 0) {
+        if ((file->off + offset) < lfs->cfg->block_size) {
+            file->pos  = npos;
+            file->off += offset;
+
+            return npos;
+        }
+    }
+
     // write out everything beforehand, may be noop if rdonly
     int err = lfs_file_flush(lfs, file);
     if (err) {


geky:
Hi @robekras, thanks for putting this optimization together, will bring this in next minor release.

lfs.c
```diff
@@ -3081,6 +3081,16 @@ static lfs_soff_t lfs_file_rawseek(lfs_t *lfs, lfs_file_t *file,
         return npos;
     }
 
+    int offset = npos - file->pos;                          // Get the difference between new and current file position
+    if (offset > 0) {                                       // If new position is after the current file position
```
geky, line 3085: Small style nit: Could we reduce the column width to <=80 chars? Possibly just by putting the comments in front of this condition

robekras: Hi geky,

as it is your project you could do what ever you want ;-).
Or does it mean, I should do the changes?
I don't want to go into a 'style' discussion. Just a little question: what is the reason for limiting the line length to 80?

Back to performance. As I wrote in issue #631: This change will only fix the case where the new position is after the old/previous position. It does not fix the other case, if the new position is before the old/previous position.
The problem is, that in this case, we do not know whether the data within the cache start from offset 0, or some bytes later. I think the first bytes of the cached block can contain some other date (block link data?).
This can only be fixed, if we would have an additional variable within the administrative structure which hold the start offset of the real file data?

BTW, I have an idea of an additional performance improvement, but I think it's better to open a new issue!?

geky:
> Back to performance. As I wrote in issue #631: This change will only fix the case where the new position is after the old/previous position.

Ah sorry, I missed this and taking a second look it looks like there's some issues.

I had assumed CI had passed but it looks like GitHub Actions broke again :/

I'm a bit nervous that this solution make not work in the general case, since we are skipping that file_flush, but CI should quickly determine if that's a problem. I've gone ahead and touched your commit to retrigger CI, sorry if that causes any issues.

---

> I think the first bytes of the cached block can contain some other date (block link data?).
> This can only be fixed, if we would have an additional variable within the administrative structure which hold the start offset of the real file data?

Hmm, this would be a better way to enable this. In theory you can compute the current block's offset from the position in the file with `lfs_ctz_index`:
``` c
lfs_off_t off = npos;
int index = lfs_ctz_index(lfs, &off);
// off is now the offset in the current block

// index is the block number in the list, but that's not important here
```

geky:
Ok apparently GitHub Actions don't run if the PR is not up to date with master. That's... new and annoying...

robekras:
I think I'm a little been lost now with all the messages here ...

I'm not an expert in working with github, and it would be no problem for me, if I'm not mentioned as contributor.

BTW, the mentioned performance issue is here: #667. I hope it is understandable

geky:
Hi @robekras, I went ahead and added a commit that implements the full idea outlined in https://github.com/littlefs-project/littlefs/issues/631#issue-1105160382 in https://github.com/littlefs-project/littlefs/pull/632/commits/4484165ff604a04fd5beb84abf27567b74b619b4, that is, `lfs_file_seek` should now avoid flushes when reading and the offset is in the cache.

This mostly involved `lfs_ctz_index` and some undocumented subtleties of the file structure/cache in order to pass the tests (though I haven't ran the full CI on it yet).

It's worth noting that this optimization only works when reading. When writing a file we always need to flush the cache since we need to make sure we writing out any pending data to disk. Any "holes" created by not-flushing the cache would corrupt the underlying data that we should be merging with our writes.

---

Would you be able to check that this provides the same performance benefit in your application with lvgl?

I know this is functionally correct, thanks to testing, but don't have good performance measurements yet.

geky:
Thanks for the PR!

lfs.c
colin-foster-in-advantage: > > Just a little question: what is the reason for limiting the line length to 80?
> 
> At some point style is all just personal preference. I like keeping the line length to 80 so that I can fit several open files next to each other at once.

Depending on the project - a .editorconfig or a .clang-format can be of value. The former makes it easier to jump from project to project. The latter is more aggressive (performs automatic formatting for you) but can make format checks in a CI workflow trivial. I'm not necessarily suggesting either, but figured this information might be beneficial to some.

geky: I feel like the diversity of editors (which is a good thing) makes .editorconfig have limited value, but .clang-format is interesting to learn about. Though I'm hesitant to adopt automatic formatting as I think overzealous rule following does more harm than good, and automatic formatters tend to be really overzealous...

 Still, we really should have some simple CI rules for things like the 80 col rule, squiggly brackets, paren spacing, etc.

PR #636, opened 2022-01-29
Support attribute enumeration for files and directories

mikee47:
This PR provides some additional attribute enumeration features.

- Add `lfs_statcfg`, `lfs_dir_readcfg` functions

    Extends regular `lfs_stat` by querying set of attribute at the same time.
    Improves performance of directory listings.

- Add `lfs_file_getattr`, `lfs_file_setattr`, `lfs_file_removeattr`

    Supports attribute manipulation on open files.
    Important for updating things like security descriptors, timestamps, etc.

    Note: Attempting to remove registered attributes fails.
    This could be handled by marking registered attribute as dirty.

- Add `lfs_file_enumattr()` function

    Allows file attributes to be efficiently enumerated,
    required for filesystem backup, etc.

- Support opening handle to directory

    Allows attributes to be accessed more efficiently, and enumerated


mikee47:
Note: These changes are part of the [Sming LittleFS](https://github.com/mikee47/Sming-LittleFS) library, used by [Sming](https://github.com/SmingHub/Sming).

xiaoxiang781216:
look like some feature is also implemented in: https://github.com/littlefs-project/littlefs/pull/636

geky:
Hi, yes, sorry this PR did not go in this release, due to time I was only looking at PRs with smaller scopes this release. I'm slow to bring in new APIs since they need to be supported for quite a while for backwards compatibility reasons.

This looks like a very interesting PR, thank @mikee47 for putting it together.

In particular I will need to see how it interacts with https://github.com/littlefs-project/littlefs/pull/580 and https://github.com/littlefs-project/littlefs/pull/513 (though https://github.com/littlefs-project/littlefs/pull/513 may come in much later due to being an API breaking change)

geky:
Hi @mikee47, sorry about the late response and lack of feedback. This is some really interesting work that deserves attention.

Unfortunately there is also work going on in the background to make attrs behave more reasonably in littlefs's CoW model (as well as a bunch of unrelated file work). My current thinking is the best route forward will be to hold off on bringing these in until other changes land, though this may create challenges for maintaining this PR...

Other requested new-API features are also in a holding pattern at the moment for similar reasons.

---

That being said, some high-level thoughts:

> Add `lfs_statcfg`, `lfs_dir_readcfg` functions

These are a clever. These would also probably make integration layers quite a bit easier when populating time/permission/etc attrs.

The only though was do you think these need a full `cfg` object? Maybe we can get away with `lfs_statattrs/lfs_stata/lfs_statx/?`, etc, that take an attr list directly?

``` c
int lfs_stat(lfs_t *lfs, const char *name, struct lfs_info *info);
int lfs_stata(lfs_t *lfs, const char *name, struct lfs_info *info,
        struct lfs_attr *attrs, lfs_size_t attr_count);
```

I'm not sure what other configuration would be useful here.

> Add `lfs_file_getattr`, `lfs_file_setattr`, `lfs_file_removeattr`

I realize these are appealing, but they are problematic with littlefs's CoW.

In littlefs, writes are not committed to mdirs until `lfs_file_sync` or `lfs_file_close`. This has the effect of making file writes an atomic "either-or" operation, which is helpful for implementing power-loss safe systems. But since attrs are stored directly in the mdir, we can't really save them anywhere unless they are backed by `lfs_attr` structs in `lfs_file_config`.

`lfs_file_getattr` would at least work.

Maybe `lfs_file_setattr` returning LFS_ERR_NOMEM if there is no backing `lfs_attr` struct associated with the specific attr would work? I'm not sure how useful this would be...

> Add `lfs_file_enumattr()` function

This is a reasonable solution. It's at least better than Linux's `listxattr` returning one big string.

Though some of the background work includes moving `lfs_fs_traverse` away from callbacks and towards a `lfs_dir_read`-esque state machine.

If applied the same transformation here, this might look like:

``` c
int lfs_attrs_open(lfs_t *lfs, lfs_attrs_t *attrs, const char *path);
int lfs_attrs_read(lfs_t *lfs, lfs_attrs_t *attrs, struct lfs_attr *attr);
int lfs_attrs_close(lfs_t *lfs, lfs_attrs_t *attrs);
int lfs_attrs_rewind(lfs_t *lfs, lfs_attrs_t *attrs);
```

Thoughts?

This might lend itself towards other ways of opening the `lfs_attrs_t` object, such as `lfs_attrs_openat`, `lfs_attrs_openmid`, `lfs_attrs_openfile`, etc.

> Support opening handle to directory

We don't really want to do this, since it loses the type info that makes it impossible to "write" to a dir object at compile time.

I'm curious if the hypothetical dir-relative openat APIs would solve most of the cases where path-based `lfs_setattr/lfs_getattr` are unwieldy.

mikee47:
@geky Thank you for your comprehensive review of this PR.

I agree with all your notes, though I would like to comment on attribute enumeration.

> > Add `lfs_file_enumattr()` function
> 
> This is a reasonable solution. It's at least better than Linux's `listxattr` returning one big string.
> 
> Though some of the background work includes moving `lfs_fs_traverse` away from callbacks and towards a `lfs_dir_read`-esque state machine.
> 
> If applied the same transformation here, this might look like:
> 
> ```c
> int lfs_attrs_open(lfs_t *lfs, lfs_attrs_t *attrs, const char *path);
> int lfs_attrs_read(lfs_t *lfs, lfs_attrs_t *attrs, struct lfs_attr *attr);
> int lfs_attrs_close(lfs_t *lfs, lfs_attrs_t *attrs);
> int lfs_attrs_rewind(lfs_t *lfs, lfs_attrs_t *attrs);
> ```
> 
> Thoughts?

I would argue here that the callback approach has several advantages over the state-based one:

- Only one function required instead of three (four including rewind)
- No overhead of repeatedly calling `read`
- Simpler to implement, no state required
- Less opportunity for internal state to change between calls. Callback function would be invoked whilst holding the LFS lock, so care required there

This meets my two main use cases for attribute enumeration:

1. Filesystem backup, there is no existing way to do this
2. Reading a set of attributes more efficiently

mikee47:
I've updated this PR based on littlefs v2.11.2:

**Revert `lfs_file_xxx` functions**

Use the regular path-based API instead.

**Rename `lfs_statcfg`, etc.**

Use `lfs_stata`, `lfs_dir_reada` and drop the `stat_cfg` structure.

**Use path for attribute enumeration**

Rename `lfs_file_enumattr` -> `lfs_enumattr` and use path instead of file handle.
Supports both files and directories without resorting to 'open directory handle' kludge.

**Revert 'open handle to directory' thing**

The need for this mostly goes away with path-based attribute enumeration.
I can manage it pretty well within Sming's integration without requiring a new API.

geky:
Hi @mikee47, this is looking really nice. I fear I will be unable to give it the attention it deserves until littlefs3 work completes (https://github.com/littlefs-project/littlefs/pull/1111, unstable). littlefs3 is the snowball of the at-the-time-experimental "unrelated file work" I mentioned above.

To complicate things littlefs3 also includes its own attr rework, though mainly focused on how attrs are broadcasted between file handles during sync, and orthogonal to `stata`/`reada`. I think `stata`/`reada` would be good to add eventually, but it may take some time to get to.

---

RE: `lfs_enumattr()`

> I would argue here that the callback approach has several advantages over the state-based one:
> 
>     * Only one function required instead of three (four including rewind)
> 
>     * No overhead of repeatedly calling `read`
> 
>     * Simpler to implement, no state required
> 
>     * Less opportunity for internal state to change between calls. Callback function would be invoked whilst holding the LFS lock, so care required there
> 
> 
> This meets my two main use cases for attribute enumeration:
> 
>     1. Filesystem backup, there is no existing way to do this
> 
>     2. Reading a set of attributes more efficiently

I agree with your points here. Coming back to this, `lfs_attrs_open` and friends would be fairly clunky.

It's interesting to note that attr iteration ended up a pretty important API in littlefs3, with the core metadata data-structure being a tree of attrs (rattrs internally). We managed to get away with a single function for both lookup and iteration by returning the closest rattr that is >= the requested rattr:

https://github.com/littlefs-project/littlefs/blob/b10efaa00654a1f65a0b7ba9d1edd47505c5d68b/lfs3.c#L3217-L3220

A single function for both get/iteration might be a bit much for the public API, but I wonder if the same idea could be applied in a sort of `lfs_nextattr` function?

``` c
// Iterate over custom attributes
//
// Finds the next attribute where type > input type
//
// Returns the number of bytes read, LFS_ERR_NOENT if no more attrs
// exist, or a negative error code on failure.
lfs_ssize_t lfs_nextattr(lfs_t *lfs, const char *path,
        int16_t *type, void *buffer, lfs_size_t size);
```

Then iteration could be done like so:
``` c
int16_t type = -1;
uint8_t buffer[SIZE];
while (true) {
    lfs_ssize_t size = lfs_nextattr(&lfs, "path.txt", &type, buffer, SIZE);
    if (size < 0) {
        if (size == LFS_ERR_NOENT) {
            break;
        }
        return size;
    }
    
    // do stuff with type + buffer + size
}
```

This would have I think 3/4 of the benefits you mentioned:

- ✓ Only one function required instead of three (four including rewind)
- ✗ No overhead of repeatedly calling read
- ✓ Simpler to implement, no state required
- ✓ Less opportunity for internal state to change between calls. Callback function would be invoked whilst holding the LFS lock, so care required there

The benefits of avoiding callbacks:

- Easier static analysis (stack measurements can't cross function pointers)
- More control flow flexibility (break, continue, goto, etc)
- Compilers seem to be a bit more cautious around callbacks, resulting in less efficient code

Mostly thinking out loud. This may also be easier to implement in littlefs3. The use of trees internally makes this sort of nearest-neighbor lookup easy.

PR #637, opened 2022-01-31
Fix lfs_file_seek doesn't update cache properties correctly.

opilat:
lfs_file_seek doesn't update cache properties in read-only mode. Invalidate cache to fix it. 

How to reproduce issue:
- Create a read-only lfs FS with a read, write, cache parameters set to 256 and erase to 4096.
- Open file bigger than 4096 and read 78B from it.
- Seek the file to position 0.
- Read from the file. This read will continue from offset 78 instead of 0.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -2730,14 +2730,18 @@ static int lfs_file_outline(lfs_t *lfs, lfs_file_t *file) {
 }
 #endif
 
-#ifndef LFS_READONLY
-static int lfs_file_flush(lfs_t *lfs, lfs_file_t *file) {
+static void lfs_file_invalidate_reading_flag(lfs_t *lfs, lfs_file_t *file) {
     if (file->flags & LFS_F_READING) {
         if (!(file->flags & LFS_F_INLINE)) {
             lfs_cache_drop(lfs, &file->cache);
         }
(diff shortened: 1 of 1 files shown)


geky:
Hi @opilat, thanks for the PR and fix! Will merge this in shortly, though I may tweak it slightly for bikeshedding reasons.

Unfortunately we don't test read-only mode at the moment, it's a bit tricky to since you need write functions to test read functions.

geky (approved):

lfs.c
```diff
@@ -2730,14 +2730,18 @@ static int lfs_file_outline(lfs_t *lfs, lfs_file_t *file) {
...
         if (!(file->flags & LFS_F_INLINE)) {
             lfs_cache_drop(lfs, &file->cache);
         }
         file->flags &= ~LFS_F_READING;
     }
+}
+
+#ifndef LFS_READONLY
+static int lfs_file_flush(lfs_t *lfs, lfs_file_t *file) {
+    lfs_file_invalidate_reading_flag(lfs, file);
```
geky, line 2744: Just a note, but I think this would be better structured by moving the `#ifndef LFS_READONLY` into `lfs_file_flush`. I will go ahead and do this after merging since I am hoping to make a patch release today.

PR #643, opened 2022-02-13
Fixes to use lfs_filebd on windows platforms

m8ddin:
There are two issues, when using the file-based block device emulation
on Windows Platforms:
1. There is no fsync implementation available. This needs to be mapped
   to a Windows-specific FlushFileBuffers system call.
2. The block device file needs to be opened as binary file (O_BINARY)
	   The corresponding flag is not required for Linux.

### DIFF
--- a/bd/lfs_filebd.c
+++ b/bd/lfs_filebd.c
@@ -10,6 +10,15 @@
 #include <unistd.h>
 #include <errno.h>
 
+#ifdef _WIN32
+// Map fsync to FlushFileBuffers on Windows
+#include <windows.h>
+#define fsync(fd) (FlushFileBuffers ((HANDLE) _get_osfhandle(fd)) ? 0 : -1)
+#else
+// O_BINARY is required and exists for Windows platforms.
+// For Linux it is not needed thus 0
+#define O_BINARY 0
+#endif
 int lfs_filebd_createcfg(const struct lfs_config *cfg, const char *path,
         const struct lfs_filebd_config *bdcfg) {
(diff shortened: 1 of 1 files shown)


geky:
Hi @m8ddin, thanks for thePR.

A request, could these be changed to explicit ifdef statements?
```
#ifdef _WIN32
bd->fd = open(path, O_RDWR | O_CREAT | O_BINARY, 0666);
#else
bd->fd = open(path, O_RDWR | O_CREAT, 0666);
#endif
```

This would just help future work to know what exact idiosyncrasies are required for each platform.

---

I also want to highlight that we can't easily test these platform specific changes in CI, so future regressions are possible. This unfortunately may break in the future without warning.

m8ddin:
Hi @geky !
Thanks for reviewing the PR so quickly! Of course, I can add explicit ifdefs. Will add another commit today.

geky:
Hi @m8ddin, thanks for the update.

Sorry, I meant that each platform specific function should be an ifdef, both the `O_BINARY` and `FlushFileBuffers`, instead of aliasing fsync. Consider what would happen if Windows adds `fsync` in the future (or more likely `fsync` is provided by mingw or some other toolchain on Windows).

``` c
#ifdef _WIN32
int err = FlushFileBuffers((HANDLE)_get_osfhandle(bd->fd)) ? 0 : -1;
#else
int err = fsync(bd->fd);
#endif
```

m8ddin:
Makes sense :+1:

geky:
Thanks for this! I will bring this in next minor release, which shouldn't be too far away.

geky (approved):

PR #647, opened 2022-02-22
Allow lfs_crc to be explicitly enabled even with a custom config

SebastianBoe:
Retain the behaviour of ifdef'ing out lfs_crc when a user has supplied
a custom config. But allow the user to explicitly enable the
implementation with the define LFS_SOFTWARE_CRC_WITH_SMALL_LOOKUP

This allows users with custom config's to re-use this lfs_crc
implementation.

Signed-off-by: Sebastian Bøe <sebastian.boe@nordicsemi.no>

### DIFF
--- a/lfs_util.c
+++ b/lfs_util.c
@@ -6,9 +6,14 @@
  */
 #include "lfs_util.h"
 
-// Only compile if user does not provide custom config
+// Only compile if the user does not provide a custom config or if the
+// user has explicitly requested the implementation with the define
+// LFS_SOFTWARE_CRC_WITH_SMALL_LOOKUP
 #ifndef LFS_CONFIG
+#define LFS_SOFTWARE_CRC_WITH_SMALL_LOOKUP
+#endif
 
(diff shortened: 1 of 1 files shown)


tejlmand (changes requested):
I understand the reasoning, but this goes against the reason of https://github.com/zephyrproject-rtos/zephyr/pull/41729 which is to remove glue code and avoid custom patches to LittleFS fork.

Only having `zephyr/module.yml` make maintenance and upmerging much easier.

Please see: https://github.com/zephyrproject-rtos/zephyr/pull/41729#discussion_r817665077

tejlmand (changes requested):
I understand the reasoning, but this goes against the reason of https://github.com/zephyrproject-rtos/zephyr/pull/41729 which is to remove glue code and avoid custom patches to LittleFS fork.

Only having zephyr/module.yml make maintenance and upmerging much easier.

Please see: https://github.com/zephyrproject-rtos/zephyr/pull/41729#discussion_r817665077

PR #657, opened 2022-03-21
Update copyright notice

geky:
Update the copyright notice to note the owner as the littlefs project moving forward.

This should have no impact on the current license (littlefs is still available under BSD 3-Clause), but just documents an ownership change.

littlefs was at one point funded by Arm under [Mbed OS](https://github.com/ARMmbed/mbed-os), but it hasn't been this way for a while. With the permission from the head of OSO at Arm, this has been moved into the category of personal project, with myself taking ownership. This is a mutual change as it clarifies that the direction of littlefs has zero reflection on Arm.

For full transparency, I'm still employed at Arm, but my work has little to do with littlefs at the moment. Anything I write here is my own. Employment aside, I'm personally still very interested in developing/maintaining littlefs and seeing where it goes.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,3 +1,4 @@
+Copyright (c) 2022, The littlefs authors.  
 Copyright (c) 2017, Arm Limited. All rights reserved.
 
 Redistribution and use in source and binary forms, with or without modification,
--- a/bd/lfs_filebd.c
+++ b/bd/lfs_filebd.c
@@ -1,6 +1,7 @@
 /*
  * Block device emulated in a file
  *
+ * Copyright (c) 2022, The littlefs authors.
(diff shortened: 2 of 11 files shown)


LICENSE.md
```diff
@@ -1,3 +1,4 @@
+Copyright (c) 2022, The littlefs project.
```
dpgeorge, line 1: Maybe `The littlefs project authors`?  Then it's copyright by the authors of the project, rather than the project itself.

geky: That is a better word choice, though I may shorten it to just `The littlefs authors`

geky: Should be updated

dpgeorge: Looks good!

PR #673, opened 2022-04-21
Fix readme Mbed link

monowii:


geky:
Hi @monowii, thanks for the PR. Do you think this should actually be 
```
https://os.mbed.com/docs/mbed-os/latest/apis/littlefilesystem.html
```

to avoid needing to keep updating the link?

geky:
Hi @monowii, I went ahead and changed this to "latest" to get it in on this patch release, but let me know if this is a mistake.

PR #675, opened 2022-04-25
Fix unused function warning with LFS_NO_MALLOC

kevinior:
If you compile lfs.c with LFS_NO_MALLOC defined you get an unused function warning for `lfs_file_rawopen`.

If you're compiling with `-Werror` that becomes an error:
```
third_party/littlefs/lfs.c:3001:12: error: 'lfs_file_rawopen' defined but not used [-Werror=unused-function]
 3001 | static int lfs_file_rawopen(lfs_t *lfs, lfs_file_t *file,
      |            ^~~~~~~~~~~~~~~~
```

This pull request fixes the warning.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -2998,12 +2998,14 @@ cleanup:
     return err;
 }
 
+#ifndef LFS_NO_MALLOC
 static int lfs_file_rawopen(lfs_t *lfs, lfs_file_t *file,
         const char *path, int flags) {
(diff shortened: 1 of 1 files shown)


dpgeorge:
+1, we are also seeing this warning.

geky:
@kevinior thanks for the PR! I will bring this in on the next patch release

geky (approved):

PR #679, opened 2022-05-16
support slash-ended mkdir, such as "mkdir dir/".

XinStellaris:
"mkdir dir/" will not work in littlefs, littlefs will return LFS_ERR_NOENT in such case. However, linux supports such commands. This patch adds support for such behavior.

Signed-off-by: 田昕 <tianxin7@xiaomi.com>


XinStellaris:
Oh no, this patch causes other problems. If we input following commands:
1. echo 1 > testfile
2. echo 2 > testfile/
 the result is testfile has content of 2. However, step2 should fail because testfile/ is a dir.

I am confused now, how to support slash-ended mkdir command? Or, is this really necessary?
@geky

geky:
Hi @XinStellaris sorry about the late response. Thanks for creating the PR, it would certainly be good to match Linux/POSIX here.

Hmm, this does look tricky to add. The simplest thing would be include a flag indicating that `lfs_dir_find` should ignore the trailing slash, this should only be set for `lfs_mkdir`. Sometimes special cases need to be special cases.

Then the test for the last name could be written:

``` c
const char *tail = name + strcspn(name, "/");
if (ignore_trailing_slashes) {
    tail += strspn(tail, "/");
}
// are we last name?
(*tail == '\0' ? id : NULL)
```

geky:
I've gone ahead and reimplemented this in: https://github.com/littlefs-project/littlefs/pull/1046

The solution was a bit involved, I decided to solve the trailing slash issue in the calling functions by checking the modified path. It turns out after `lfs_dir_find`, since the modified path is no longer NULL-terminated (slash-terminated?), we can determine whether or not the path points to a directory or not based on what follows the last file name.

Let me know if you see any issues with https://github.com/littlefs-project/littlefs/pull/1046, and thanks for the original PR!

PR #691, opened 2022-05-31
added LFS_NO_ASSERT guards around uses of lfs_mlist_isopen

embeddedpenguin:
I was getting some compilation failures when using `#define LFS_NO_ASSERT`

```
/usr/libexec/gcc/arm-none-eabi/ld: ./Shared/littlefs/lfs.o: in function `lfs_file_opencfg':
../Shared/littlefs/lfs.c:5501: undefined reference to `lfs_mlist_isopen'
/usr/libexec/gcc/arm-none-eabi/ld: ./Shared/littlefs/lfs.o: in function `lfs_file_close':
../Shared/littlefs/lfs.c:5516: undefined reference to `lfs_mlist_isopen'
/usr/libexec/gcc/arm-none-eabi/ld: ./Shared/littlefs/lfs.o: in function `lfs_file_read':
../Shared/littlefs/lfs.c:5550: undefined reference to `lfs_mlist_isopen'
/usr/libexec/gcc/arm-none-eabi/ld: ./Shared/littlefs/lfs.o: in function `lfs_file_write':
../Shared/littlefs/lfs.c:5568: undefined reference to `lfs_mlist_isopen'
/usr/libexec/gcc/arm-none-eabi/ld: ./Shared/littlefs/lfs.o: in function `lfs_file_seek':
../Shared/littlefs/lfs.c:5586: undefined reference to `lfs_mlist_isopen'
collect2: error: ld returned 1 exit status
```

I think this is because LFS_ASSERT will still be called even with `#define LFS_NO_ASSERT`. It just makes LFS_NO_ASSERT do nothing. So on 
https://github.com/littlefs-project/littlefs/blob/master/lfs.c#L5501
https://github.com/littlefs-project/littlefs/blob/master/lfs.c#L5516
and so on...

The assumption is LFS_ASSERT() will void out, but since lfs_mlist_isopen is still being called and isn't defined when using LFS_NO_ASSERT, it doesn't really get that far.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -5477,7 +5477,9 @@ int lfs_file_open(lfs_t *lfs, lfs_file_t *file, const char *path, int flags) {
     }
     LFS_TRACE("lfs_file_open(%p, %p, \"%s\", %x)",
             (void*)lfs, (void*)file, path, flags);
+#ifndef LFS_NO_ASSERT
     LFS_ASSERT(!lfs_mlist_isopen(lfs->mlist, (struct lfs_mlist*)file));
+#endif
 
     err = lfs_file_rawopen(lfs, file, path, flags);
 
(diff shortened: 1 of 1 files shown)


embeddedpenguin:
I also don't really get why lfs_mlist_isopen needs to be wrapped in a LFS_NO_ASSERT guard so I may just be missing something.

geky:
I can see it's confusing. Are you defining a custom LFS_ASSERT? If you do that you shouldn't define LFS_NO_ASSERT, lfs_util.h includes an ifdef for LFS_ASSERT to avoid conflicts ([line 94](https://github.com/littlefs-project/littlefs/blob/40dba4a556e0d81dfbe64301a6aa4e18ceca896c/lfs_util.h#L94)).

LFS_NO_ASSERT is intended to disable asserts completely. Rather than putting ifdefs in every function, lfs_util.h defines LFS_ASSERT to be a noop.

Rather annoyingly, lfs_mlist_isopen is only used in an assert. This led to unused function warnings with LFS_NO_ASSERT, which is why it gets an ifdef https://github.com/littlefs-project/littlefs/pull/499.

PR #715, opened 2022-07-27
lfs_filebd_sync: fix compilation on Windows

Mixaill:


geky:
Hi @Mixaill, thanks for the PR, looks good to me. Not sure how that was introduced but I suppose that's the frailty of not testing an OS.

geky (approved):

PR #718, opened 2022-08-02
Add "chamelon" to the related projects section.

yomimono:
"chamelon" implements a subset of littlefs (no global move state or
singly-linked list threaded through the directory tree) for use in the
MirageOS library operating system project. It is written entirely in
OCaml and is interoperable (with the above caveats) with the reference
implementation via FUSE.


yomimono:
OK! Thanks for your response. I'll close this and let future contributors include `chamelon`, or choose not to do so, as is appropriate with the context at the time.

Issue #684, opened 2022-05-23
Compiler warnings in `lfs_dir_traverse` about `dir = dir` etc


dpgeorge:
Removal of recursion in `lfs_dir_traverse` in 8109f282668d0396c1a8df05d38ed215b0b8feac added 4 lines which give warnings on some compilers (eg clang):
```
               dir = dir;
               off = off;
               ptag = ptag;
               attrs = attrs;
```
I suggest to remove these 4 lines.

dpgeorge:
Example of warning:
```
../../lib/littlefs/lfs2.c:868:27: error: explicitly assigning value of variable of type 'int' to itself [-Werror,-Wself-assign]
                attrcount = attrcount;
                ~~~~~~~~~ ^ ~~~~~~~~~
```

geky:
Thanks for creating an issue, it sounds like we should add clang to CI with -Werror like we have GCC.

I'll look at these for the next patch release.

dpgeorge:
> it sounds like we should add clang to CI with -Werror like we have GCC

Yes I think that would be a good idea, to prevent other warnings from creeping in.

So, is it the correct fix to just remove these 5 lines?  (We need to apply that patch to our copy of littlefs.)

geky:
> So, is it the correct fix to just remove these 5 lines? (We need to apply that patch to our copy of littlefs.)

Yes those lines have no effect. I think I left them as a documentation tool since those variables are the "stack frame" of the lfs_dir_traverse function.

geky:
Sorry for about the timing. I have a PR for this up here now: https://github.com/littlefs-project/littlefs/pull/724

PR #724, opened 2022-09-07
Fix self-assign warnings discovered by clang, remove some warning flags

geky:
This was a false positive, but having clean warnings in clang would be good to have. See https://github.com/littlefs-project/littlefs/issues/684 for more info.

Also added clang compilation to CI.

---

Note that I've also removed some of the extra warning flags applied to littlefs up until this point. Some are not available in clang, and I've been meaning to remove these for a while as I've not found them more annoying than useful.

Found by @dpgeorge

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -292,6 +292,27 @@ jobs:
       - name: test-valgrind
         run: make test TESTFLAGS+="-k --valgrind"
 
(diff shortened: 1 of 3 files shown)


dpgeorge:
Looks good, thanks!

yamt:
lgtm

PR #736, opened 2022-10-06
Fix "unused" compiler warning when using static allocation

fwolter:


fwolter:
Just saw there is already a PR pending for this issue.

PR #740, opened 2022-10-10
Fix invalid block size reporting.

cbiffle:
This boilerplate got copied from the stanza just above and incompletely edited.


geky:
Hi @cbiffle, good catch. Thanks for creating a PR, will merge this on the next patch release.

PR #742, opened 2022-10-27
lfs_util: Fix endiannes conversion when LFS_NO_INTRINSICS is set

carlescufi:
The logic for endiannes conversion was wrong when LFS_NO_INTRINSICS was set, since on endinanes match a check of that macro would prevent the unchanged value from being returned.

Signed-off-by: Carles Cufi <carles.cufi@nordicsemi.no>


gmarull (approved):

de-nordic (approved):
Looks OK, Thanks.

geky:
Hi @carlescufi, this looks great, thanks for this. Will merge into the next patch release.

geky (approved):

PR #752, opened 2022-12-02
Add test/bench runners, benchmarks, additional scripts

geky:
This PR brings in a number of changes to how littlefs is tested and measured.

Originally, the motivation was to add a method for benchmarking the filesystem, to lay the groundwork for future performance improvements, but the scope ended up growing to include a number of fixes/improvements to general littlefs testing.

1. **Reworked test framework, no. 3**

   The test framework gets a rework again, taking what worked well in the current test framework and throwing out what doesn't.

   The main goals behind this rework were to 1. simplify the framework, even if it means more boilerplate, as this should make it easier to extend with new features, and 2. run the tests as fast as possible.
   
   Previously I've disregarded test performance, worried a focus on test performance risks complexity and difficulty in understanding the system that is being debugged, but my perspective is changing as faster tests => more tests => more confidence => ~the dark side~ a safer filesystem. If you've told me previously to parallelize the tests, etc, this is the part where you can say you told me so.

   - **Tests incrementally compile, and we don't rebuild lfs.c for every suite**

     Previously the test's build system and runner was all self-contained in test.py. On one hand this meant you only needed test.py to build/run the tests, but on the other hand this design was confusing, limiting, and just all around problematic. One big issue was that, being outside of the build system, tests couldn't be built incrementally and every test suite needed a custom built version of lfs.c. This led to a slow debugging experience as each change to lfs.c needed at least 16 recompilations.

     Now the test framework is integrated into the Makefile with separate build steps for applying prettyasserts.py and other scripts, all of which can be built incrementally, significantly reducing the time spent waiting for tests to recompile.

   - **runners/test_runner is now its own standalone application**

     Previously any extra features/configuration had to be built into the test binaries during compilation. Now there is an explicit test_runner.c which can contain high-level test features that can be engaged at runtime through additional flags.

     This makes it easier to add new test features, but also makes it easier to debug the test_runner itself, as it's no longer hidden inside test.py.

     The actual tests are provided at link-time using a custom linker section, and are still generated by `./scripts/test.py -c`

     ``` bash
     $ make test-runner
     ...
     $ ./runners/test_runner -l
     suite                      flags   cases       perms
     test_alloc                     -      12       62/70
     test_attrs                     -       4       20/20
     ...
     $ ./runners/test_runner
     running test_alloc_parallel:1g12gg2f3g1ghsj5
     ...
     ```

   - **Tests now avoid spawning processes as much as possible**

     When you find a bug in C, it often leads to undefined behavior, memory corruption, etc, making the current test process no longer sound. But you also often want to keep running tests to see if there is any trends among the test failures. To accomplish this, the previous test framework ran each test in its own process.

     Unfortunately, process spawning is not really a cheap operation. And with _most_ tests not failing (hopefully), this ends up wasting a significant amount of time just spawning processes.

     Now, with a more powerful test_runner, the test framework tries to run as many tests in a single process. Only spawning a new process when a test fails. This is all handled by scripts/test.py, which interacts with runners/test_runner, telling it which tests to run via the low-lever `--step` flag.

   - **Powerloss is now simulated with setjmp/longjmp**

     As a part of reducing process spawning, powerloss is directly simulated in the test_runner using setjmp/longjmp. Previously powerloss was simulated by killing and restarting the process. Which is a simple, heavy-handed solution that works. Slowly.

     Since there can be thousands of powerlosses in a single test, this needed to be moved into the test_runner, especially since powerloss testing is arguably the most important feature of littlefs's test framework.

     As an added plus, the simulated block-device no longer needs to be persisted in the host's filesystem when powerloss testing, and can stay comfortably in the test_runner's RAM. The cost of persisting the block-device could be mitigated by using a RAM-backed tmpfs disk, but this still incurred a cost as all block-device operations would need to go through the OS.

     Using setjmp/longjmp can lead to memory leaks when reentrant tests call malloc, but since littlefs uses malloc in only a handful of convenience functions (littlefs's whole goal is minimal RAM after all), this doesn't seem to been a problem so far.

   - **Tests now run in parallel**

     Perhaps the lowest-hanging fruit, tests now run in parallel.

     The exact implementation here is a bit naive/suboptimal, giving each process n/m tests to run for n tests and m cores, but this keeps the process/thread management in the high-level test.py python layer, simplifying thread management and avoiding a multi-threaded test_runner.

     ``` bash
     $ ./scripts/test.py ./runners/test_runner -j -v
     using runner: ./runners/test_runner
     ./runners/test_runner --list-cases
     ./runners/test_runner --list-case-paths
     found 17 suites, 130 cases, 4202/4315 permutations

     ./runners/test_runner --list-cases
     ./runners/test_runner --list-case-paths
     ./runners/test_runner -s0,,12
     ./runners/test_runner -s1,,12
     ./runners/test_runner -s2,,12
     ./runners/test_runner -s3,,12
     ./runners/test_runner -s4,,12
     ./runners/test_runner -s5,,12
     ./runners/test_runner -s6,,12
     ./runners/test_runner -s7,,12
     ./runners/test_runner -s8,,12
     ./runners/test_runner -s9,,12
     ./runners/test_runner -s10,,12
     ./runners/test_runner -s11,,12
     ...
     ```

   The combination of the above improvements allows us to run the tests a lot faster, and/or cram in a lot more tests:

   |        | Test&nbsp;permutations | Runtime (single&nbsp;core) | Runtime (6&nbsp;cores/12&nbsp;threads) | Tests&nbsp;per&nbsp;second (single&nbsp;core) | Tests&nbsp;per&nbsp;second (6&nbsp;cores/12&nbsp;threads) |
   |:-------|-----------------------:|---------------------------:|---------------------------------------:|----------------------------------------------:|----------------------------------------------------------:|
   | Before |                    897 |               51.67&nbsp;s |                           31.73&nbsp;s |                                17.36&nbsp;t/s |                                            28.27&nbsp;t/s |
   | After  |                   4202 | 1&nbsp;m&nbsp;22.26&nbsp;s |                           26.19&nbsp;s |                51.08&nbsp;t/s&nbsp;(+194.23%) |                           160.44&nbsp;t/s&nbsp;(+467.53%) |

   (Most of the new permutations are from moving the different test geometries out of CI and into the test_runner. Note the previous test framework _does_ parallelize builds, which are included.)

2. **Exhaustive powerloss testing**

   In addition to the heuristic-based powerloss testing, the new test_runner can also exhaustively search all possible powerloss scenarios for a given reentrant test.

   To speed this up, the test_runner uses a simulated, copy-on-write block-device (reintroducing emubd), such that all possible code-paths in all possible powerloss scenarios are executed at most once. And, because most of the block-device's state can be shared via copy-on-write operations, each powerloss branch needs at most one additional block of memory in RAM.

   The runtime still grows exponentially, and we each have a finite lifetime, so it will be more useful to exhaustively search a bounded number of powerlosses. Here's a run of all possible 5-deep powerlosses in the test_move test suite:

   ``` bash
   $ ./scripts/test.py ./runners/test_runner test_move -P5 -j
   using runner: ./runners/test_runner -P5
   found 1 suites, 17 cases, 10/10 permutations

   running tests: 1/1 suites, 17/17 cases, 10/10 perms, 7981852pls!

   done: 10/10 passed, 0/10 failed, 7981852pls!, in 951.48s
   ```

   Since it can be a bit annoying to wait 15 minutes to reproduce a test failure, each powerloss scenario is encoded in a leb16 suffix appended to the current test identifier. This, combined with a leb16-encoding of the test's configuration and the test's name, can uniquely identify and reproduce any test run in the test_runner:

   ```
   test_dirs_many_reentrant:2gg2cb:k4o6
   ^                        ^  ^^^ ^ ^
   '------------------------|--|||-|-|-- test_dirs_many_reentrant
                            '--|||-|-|--   2 =   0x2 = BLOCK_SIZE
                               '||-|-|-- gg2 = 0x200 = 512
                                '|-|-|--   c =   0xc = N
                                 '-|-|--   b =   0xb = 11
                                   '-|--  k4 =  0x44 = powerloss after 68 writes
                                     '--  o6 =  0x68 = powerloss after 104 writes
   ```

   So once a failing test scenario is found, the exact state of the failure can be quickly reproduced for debugging:

   ``` bash
   $ ./scripts/test.py ./runners/test_runner -P2 -b -j
   using runner: ./runners/test_runner -P2
   found 17 suites, 130 cases, 390/400 permutations

   running test_alloc: 12/12 cases, 0/0 perms
   running test_attrs: 4/4 cases, 0/0 perms
   running test_badblocks: 4/4 cases, 0/0 perms
   running test_bd: 5/5 cases, 0/0 perms
   running test_dirs: 12/14 cases, 12/18 perms, 39477pls!, 1/18 failures

   done: 12/18 passed, 1/18 failed, 39477pls!, in 23.46s

   tests/test_dirs.toml:180:failure: test_dirs_many_reentrant:2gg2cb:k4o6 (BLOCK_SIZE=512, N=11) failed
   powerloss test_dirs_many_reentrant:2gg2cb:k4k6
   powerloss test_dirs_many_reentrant:2gg2cb:k4l6
   powerloss test_dirs_many_reentrant:2gg2cb:k4m6
   powerloss test_dirs_many_reentrant:2gg2cb:k4n6
   powerloss test_dirs_many_reentrant:2gg2cb:k4o6
   tests/test_dirs.toml:196:assert: assert failed with false, expected eq true
           assert(err == 0 || err == LFS_ERR_EXIST);

   $ ./scripts/test.py ./runners/test_runner test_dirs_many_reentrant:2gg2cb:k4o6 --gdb
   ...
   196             assert(err == 0 || err == LFS_ERR_EXIST);
   (gdb)
   ```

   Unfortunately, the current tests are not the most well designed for exhaustive powerloss testing. Some of them, test_files and test_interspersed specifically, write large files a byte at a time. Under exhaustive powerloss testing, these result in, well, a lot of powerlosses, but outside of the writes with data-structure changes, don't reveal anything interesting. This is something that can probably be improved over time.

   Exhaustively testing all powerlosses at a depth of 1 takes 12.79 minutes with 84,484 total powerlosses.

   Exhaustively testing all powerlosses at a depth of 2 takes at least 4 days, and is still running... I'll let you know when it finishes...

3. **scripts/bench.py and runners/bench_runner**

   This PR introduces scripts/bench.py and runners/bench_runner, siblings to scripts/test.py and runners/test_runner, for measuring the performance of littlefs. Instead of reporting pass/fail, the bench_runner reports the total number of bytes read, programmed, and erased during a bench case. This can be useful for comparing different littlefs implementations, as these numbers map directly to hardware-dependent performance in IO-bound applications.

   One feature that makes this useful, added to both the bench_runner and test_runner, is a flexible configuration system evaluated at runtime. This has the downside of limiting configurable bench/test defines to `uintmax_t` integers, but makes it easy to quickly test/compare/reproduce different configurations:

   ``` bash
   $ ./scripts/bench.py ./runners/bench_runner bench_file_write -Gnor --list-defines
   READ_SIZE=1
   PROG_SIZE=1
   BLOCK_SIZE=4096
   BLOCK_COUNT=256
   CACHE_SIZE=64
   LOOKAHEAD_SIZE=16
   BLOCK_CYCLES=-1
   ERASE_VALUE=255
   ERASE_CYCLES=0
   BADBLOCK_BEHAVIOR=0
   POWERLOSS_BEHAVIOR=0
   CHUNK_SIZE=64
   ORDER=0,1,2
   SIZE=131072
   $ ./scripts/bench.py ./runners/bench_runner bench_file_write -Gnor -DORDER=0 -DSIZE="range(0,24576,64)"
   using runner: ./runners/bench_runner -Gnor -DORDER=0 '-DSIZE=range(0,24576,64)'
   found 1 suites, 1 cases, 384/384 permutations

   running benches: 1/1 suites, 1/1 cases, 384/384 perms

   done: 4801420 readed, 4729056 proged, 5492736 erased, in 0.08s
   ```

   At the moment I've only added a handful of benchmarks, though the number may increase in the future. The goal isn't to maintain a fully cohesive benchmark suite, as much as it is to have a set of tools for analyzing specific performance bottlenecks.

4. **Reworked scripts/summary.py and other scripts to be a bit more flexible**

   This mainly means scripts/summary.py is no longer hard-wired to work with the compile-time measurements, allowing it to be used with other results such as benchmarks, though this comes with the cost of a large number of flags for controlling the output.

   Each measurement script also now comes with a `*-diff` Makefile rule for quick comparisons.

   ``` bash
   $ make summary
   ./scripts/code.py lfs.o lfs_util.o -q  -o lfs.code.csv
   ./scripts/data.py lfs.o lfs_util.o -q  -o lfs.data.csv
   ./scripts/stack.py lfs.ci lfs_util.ci -q  -o lfs.stack.csv
   ./scripts/structs.py lfs.o lfs_util.o -q  -o lfs.structs.csv
   ./scripts/summary.py lfs.code.csv lfs.data.csv lfs.stack.csv lfs.structs.csv -fcode=code_size -fdata=data_size -fstack=stack_limit --max=stack -fstructs=struct_size -Y
                               code    data   stack structs
   TOTAL                      25614       -    2176     908
   $ make summary-diff
   ./scripts/summary.py <(./scripts/code.py ./lfs.o ./lfs_util.o -q -o-) <(./scripts/data.py ./lfs.o ./lfs_util.o -q -o-) <(./scripts/stack.py ./lfs.ci ./lfs_util.ci -q -o-) <(./scripts/structs.py ./lfs.o ./lfs_util.o -q -o-) -fcode=code_size -fdata=data_size -fstack=stack_limit --max=stack -fstructs=struct_size -Y -d <(./scripts/summary.py ./lfs.code.csv ./lfs.data.csv ./lfs.stack.csv ./lfs.structs.csv -q -o-)
                              ocode   odata  ostack    ostructs   ncode   ndata  nstack    nstructs   dcode   ddata  dstack    dstructs
   TOTAL                      25614       -    2176         908   25614       -    2176         908      +0      +0      +0          +0
   ```

5. **Reworked scripts/cov.py to take advantage of the --json-format introduced in GCC 9**

   It's a bit concerning that this was a breaking change in gcov's API, albeit on a major version, but the new --json-format is much easier to work with.

   It's also worth noting this PR includes a change in ideology around coverage measurement. Instead of collecting coverage from as many sources as possible in CI, coverage is only collected from the central `make test` run. This will result in lower coverage numbers than previously, but these are the coverage numbers we actually care about: test coverage via easy to reproduce and isolate tests.

   This also simplifies coverage collection in CI, which is a plus.

6. **scripts/perf.py and scripts/perfbd.py**

   perf.py was added as an experiment with Linux's perf tool, which uses an interest method of sampling performance counters to build an understanding of the performance of a system. Unfortunately this isn't the most useful measurement for littlefs, as we should expect littlefs's performance to be dominated by IO overhead. But it may still be useful for tracking down CPU bottlenecks

   perfbd.py takes the ideas in Linux's perf tool and applies them to the bench_runner. Instead of sampling performance counters, we can sample littlefs's trace output to find low-level block-device operations. Combining this with stack traces provided by the [backtrace](https://www.gnu.org/software/libc/manual/html_node/Backtraces.html) function, we can propagate IO cost to their callers, building a useful map of the source of IO operations in a given benchmark run:

   ``` bash
   $ ./scripts/bench.py ./runners/bench_runner bench_file_write -Gnor -DORDER=0 -DSIZE="range(0,24576,64)" -t lfs.bench.trace --trace-backtrace --trace-freq=10000
   using runner: ./runners/bench_runner -Gnor -tlfs.bench.trace --trace-backtrace --trace-freq=10000 -DORDER=0 '-DSIZE=range(0,24576,64)'
   found 1 suites, 1 cases, 384/384 permutations

   running benches: 1/1 suites, 1/1 cases, 384/384 perms

   done: 4801420 readed, 4729056 proged, 5492736 erased, in 0.10s
   $ ./scripts/perfbd.py ./runners/bench_runner lfs.bench.trace -Flfs.c -s
   function                      readed  proged  erased                                           
   lfs_bd_erase                       0       0   36864                                           
   lfs_format                       372      52   28672                                           
   lfs_dir_orphaningcommit           64      52   28672                                           
   lfs_dir_relocatingcommit          64      52   28672                                           
   lfs_dir_compact                   28      52   28672                                           
   lfs_fs_deorphan                    0       0   28672                                           
   lfs_file_rawwrite               3268    3200    8192                                           
   lfs_file_write                  3268    3200    8192                                           
   lfs_file_flushedwrite           3268    3200    4096                                           
   lfs_file_relocate                516     512    4096                                           
   lfs_bd_flush                    2752    3252       0                                           
   lfs_bd_prog                     2752    3200       0                                           
   lfs_dir_commit                    64      52       0                                           
   lfs_dir_commitcrc                 64      52       0                                           
   lfs_bd_read                     3688       0       0                                           
   lfs_bd_cmp.constprop.0          2752       0       0                                           
   lfs_dir_fetchmatch               868       0       0                                           
   lfs_dir_fetch                    856       0       0                                           
   lfs_alloc                        516       0       0                                           
   lfs_fs_rawtraverse               516       0       0                                           
   lfs_file_open                     36       0       0                                           
   lfs_file_rawopencfg               36       0       0                                           
   lfs_mount                          8       0       0                                           
   lfs_dir_find                       4       0       0                                           
   lfs_dir_get                        4       0       0                                           
   lfs_dir_getslice                   4       0       0                                           
   lfs_file_close                     4       0       0                                           
   lfs_file_rawclose                  4       0       0                                           
   lfs_file_rawsync                   4       0       0                                           
   TOTAL                          25780   16876  204800                                           
   ```

   It's worth noting that these numbers are samples. They are a subset and don't add up to the total IO cost of the benchmark. But they are still useful as a metric for understand benchmark performance.

   You could parse the entire trace output without sampling, but this would be quite slow and not really show you any more info.

7. **scripts/plot.py and scripts/plotmpl.py**

   Added plot.py and plotmpl.py for quick plotting of littlefs measurements in the terminal and with Matplotlib. I think these will mostly be useful for looking for growth rates in benchmark results. And also future documentation.

   ``` bash
   $ ./scripts/bench.py ./runners/bench_runner bench_file_write -Gnor -DORDER=0 -DSIZE="range(0,24576,64)" -o lfs.bench.csv
   using runner: ./runners/bench_runner -Gnor -DORDER=0 '-DSIZE=range(0,24576,64)'
   found 1 suites, 1 cases, 384/384 permutations

   running benches: 1/1 suites, 1/1 cases, 384/384 perms

   done: 4801420 readed, 4729056 proged, 5492736 erased, in 0.25s
   $ ./scripts/plotmpl.py lfs.bench.csv -o lfs.bench.svg -tbench_file_write -l -xSIZE -ybench_readed -ybench_proged -ybench_erased --x2 --xunits=B --y2 --yunits=B --github
   updated lfs.bench.svg, 3 datasets, 1152 points
   ```

   example-plot-gh
   example-plot-gh-dark

8. **scripts/tracebd.py, scripts/tailpipe.py, scripts/teepipe.py**

   These are some extra scripts for interacting with/viewing littlefs's trace output.

   tailpipe.py and teepipe.py behave similarly to Unix's tail and tee programs, but work a bit better with Unix pipes, with resumability and fast paging.

   The most interesting script is tracebd.py, which parses littlefs's trace output for block-device operations and renders it as ascii art. I've used this sort of block-device operation rendering previously for a quick demo and it can be surprisingly useful for understanding how filesystem operations interact with the block-device.

   ``` bash
   $ mkfifo trace
   $ ./scripts/bench.py ./runners/bench_runner bench_file_write -Gnor -DORDER=0 -DSIZE="range(0,24576,64)" -t trace
   ...
   ```
   ``` bash
   $ ./scripts/tracebd.py trace -c10000 -z                 
   e.........................................e.....................................
   e.........................................e.....................................
   e.........................................ee....................................
   e.........................................ee....................................
   e.........................................ee....................................
   e.........................................ee....................................
   e.........................................ee....................................
   e.........................................ee....................................
   e.........................................ee....................................
   e.........................................ee....................................
   e.........................................ee....................................
   e.........................................eee...................................
   e.........................................eee...................................
   e.........................................eee...................................
   e.........................................eee...................................
   e.........................................eee...................................
   ```

9. **Changed lfs.a -> liblfs.a in default build rule**

   The `lib*` prefix is usually required by the linker, so I suspect this won't break anything. But it's worth mentioning this change in case someone relies on the current build target.

10. **Added a `make help` rule**

    I think I first saw this [here](https://marmelab.com/blog/2016/02/29/auto-documented-makefile.html), this self-documenting Makefile rule gives some of the useful Makefile rules a bit more discoverability.

11. **Adopted script changes in CI, added a bot comment on PRs**

    Thanks to GitHub Actions, we have a lot of info about builds in CI. Unfortunately, statuses on GitHub have been becoming harder to find each UI change. To help keep this info discoverable I've added an automatically generated comment that @geky-bot should post after a succesful CI run. Hopefully this will contribute to PRs without being too annoying.

    You can see some example comments on the PR I created on my test fork:
    https://github.com/geky/littlefs-test-repo/pull/4

---

The increased testing did find a couple bugs: eba55533140e28fd838b5ec1154b892e08bea838 and 0b11ce03b7ded49a878f155d49bb3ca9aed2f0a4. Their commit messages have more details on the bugs and their fixes. And with the new test identifiers I can tell you the exact state that will trigger the failures:

- `test_relocations_reentrant_renames:112gg261dk1e3f3:123456789abcdefg1h1i1j1k1l1m1n1o1p1q1r1s1t1u1v1g2h2i2j2k2l2m2n2o2p2q2r2s2t2` - eba55533140e28fd838b5ec1154b892e08bea838 - found with linear heuristic powerlosses
- `test_dirs_many_reentrant:2gg2cb:k4o6` - 0b11ce03b7ded49a878f155d49bb3ca9aed2f0a4 - found with 2-deep exhaustive powerlosses


geky:
After only 4 days, 20 hours, with 144,437,889 powerlosses, the exhaustive powerloss testing with all 2-deep powerlosses finished successfully:

```
$ ./scripts/test.py runners/test_runner -b -j -P2
using runner: runners/test_runner -P2
found 17 suites, 130 cases, 390/400 permutations

running test_alloc: 12/12 cases, 0/0 perms
running test_attrs: 4/4 cases, 0/0 perms
running test_badblocks: 4/4 cases, 0/0 perms
running test_bd: 5/5 cases, 0/0 perms
running test_dirs: 14/14 cases, 18/18 perms, 87700pls!
running test_entries: 8/8 cases, 0/0 perms
running test_evil: 8/8 cases, 0/0 perms
running test_exhaustion: 5/5 cases, 0/0 perms
running test_files: 10/10 cases, 245/245 perms, 7559438pls!
running test_interspersed: 4/4 cases, 30/30 perms, 123954557pls!
running test_move: 17/17 cases, 10/10 perms, 6120pls!
running test_orphans: 2/2 cases, 13/13 perms, 73319pls!
running test_paths: 13/13 cases, 0/0 perms
running test_relocations: 4/4 cases, 24/24 perms, 142460pls!
running test_seek: 6/6 cases, 15/15 perms, 12501750pls!
running test_superblocks: 14/14 cases, 15/15 perms, 64018pls!
running test_truncate: 7/7 cases, 20/20 perms, 48527pls!

done: 390/390 passed, 0/390 failed, 144437889pls!, in 418702.72s
```

PR #766, opened 2023-01-26
Add test scenario for truncating to a block size

colin-foster-in-advantage:
When truncation is done on a file to the block size, there seems to be an error where it points to an incorrect block. Perform a write / truncate / readback operation to verify this issue.

Signed-off-by: Colin Foster <colin.foster@in-advantage.com>

### DIFF
--- a/tests/test_truncate.toml
+++ b/tests/test_truncate.toml
@@ -437,3 +437,37 @@ code = '''
     lfs_file_close(&lfs, &file) => 0;
     lfs_unmount(&lfs) => 0;
 '''
+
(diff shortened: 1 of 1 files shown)


geky:
Hi @colin-foster-in-advantage, thanks for creating a test case. This made it very easy to reproduce. Sorry about the delay in addressing this.

I ended up generalizing the existing truncation tests so this case and others should be covered in this PR: https://github.com/littlefs-project/littlefs/pull/800

Let me know if you see any issues with this.

Issue #764, opened 2023-01-25
unsafe free-ing of memory


evpopov:
All calls to lfs_free() suffer from 2 issues:
- The if() clause that they are in does NOT ensure that the pointer that is about to be free-ed is non-null
- After the call to lfs_free(), the pointer that was just free-ed is not set to null

Reproducible by 2 consecutive calls to lfs_unmount() causing an attempt to free the cache pointers twice. I don't know about other heap implementations, but mine cannot guard against double free attempts and corrupts the heap in an instant.

Here's my solution for lfs_deinit() that covers 3 of the 4 calls to lfs_free. I did the same in lfs_file_rawclose() which is the only other place that lfs_free is called.
```
static int lfs_deinit(lfs_t *lfs) {
    // free allocated memory
    if (!lfs->cfg->read_buffer && lfs->rcache.buffer) {
        lfs_free(lfs->rcache.buffer);
		lfs->rcache.buffer = NULL;
    }

    if (!lfs->cfg->prog_buffer && lfs->pcache.buffer) {
        lfs_free(lfs->pcache.buffer);
		lfs->pcache.buffer = NULL;
    }

    if (!lfs->cfg->lookahead_buffer && lfs->free.buffer) {
        lfs_free(lfs->free.buffer);
		lfs->free.buffer = NULL;
    }

    return 0;
}
```

BenBE:
`lfs_free` calls `free` which has the NULL check integrated. Thus no check has to be done for `NULL` in advance.

BUT, and I agree with the suggestion here, after `lfs_free` has been called, the pointer should be cleared explicitly.

evpopov:
@BenBE
Yes you are absolutely correct and my heap implementation is just fine when it receives a null.
I guess I'm just one of those programmers that feels safer with the non-null check.... my bad.

BenBE:
There's nothing wrong with the additional check. It's just not necessary.

evpopov:
I created #790 to address the issue. Closing this to keep things clean.

PR #776, opened 2023-02-08
Do not call lfs_free if the buffer to free is NULL

cdesjardins:
This can happen if lfs_malloc fails, or if the file doesn't exist, all "goto cleanup" clauses before and including the the lfs_malloc will cause this.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -3014,7 +3014,7 @@ static int lfs_file_rawclose(lfs_t *lfs, lfs_file_t *file) {
     lfs_mlist_remove(lfs, (struct lfs_mlist*)file);
 
     // clean up memory
-    if (!file->cfg->buffer) {
+    if ((!file->cfg->buffer) && (file->cache.buffer)) {
         lfs_free(file->cache.buffer);
     }
 


vonj:
> This can happen if lfs_malloc fails, or if the file doesn't exist, all "goto cleanup" clauses before and including the the lfs_malloc will cause this.

Can this be triggered in a unit test, or would a malloc fail somehow have to be provoked?

BenBE:
Please note [my comment in this discussion](https://github.com/littlefs-project/littlefs/issues/764#issuecomment-1403724362): `free` with `NULL` is a no-op. Only thing that littlefs should guarantee is to properly initialize all its memory so unset pointers are `NULL` instead of random values..

geky:
It's worth noting @cdesjardins's case is a bit different from #790, since the case of the "goto cleanup" in lfs_file_open can be reached in normal operation.

That being said, as @BenBE points out, free with NULL is usually a no-op, and is guaranteed to be so in a standard-conforming stdlib.

And, if free does not support NULL, the `lfs_free` function is under user control if they create their own lfs_util.h file. I believe this would be the correct place to add such a NULL check.

Issue #785, opened 2023-03-13
Using lfs_dir_tell at the end of a directory succeds, but then lfs_dir_seek fails


sosthene-nitrokey:
Behaviour
========

When iterating over a directory with `lfs_dir_open` and `lfs_dir_read` in a loop, if after calling `lfs_dir_read` on the last file, one calls `lfs_dir_tell`, the call success. However, calling `lfs_dir_seek` with the given value fails with `-22` (invalid).

Expected behaviour
===============

I'd expect this to succeed and only further calls to `lfs_dir_read` to fail. It would also be reasonable to have `lfs_dir_tell` fail too with a `NOENT` error..

Reproducer
=========

The `test_dirs` doesn't iterate over all the files created in the directory, so this bug doesn't happen in the test suite. See #786  for an updated test suite that triggers the bug.

geky:
Hi @sosthene-nitrokey, thanks for creating an issue. I agree with your evaluation.

Out of curiosity I tested telldir/seekdir locally on Linux and it returns a NULL dirent without error. It makes sense as this matches the behavior of file tell/seek when at the end of a file.

I created https://github.com/littlefs-project/littlefs/pull/805 to fix this, let me know if you see any issues with it.

PR #786, opened 2023-03-13
Add missing iterations in tests

sosthene-nitrokey:
The tests don't actually test every file.
The iteration skips the two `.` and `..` directories, which means that the last two files were not read. This PR fixes that. This seems to reach an edge case reported in #785.

As can be seen, the tests fail at a call to `lfs_dir_seek`, when they shouldn't.

### DIFF
--- a/tests/test_dirs.toml
+++ b/tests/test_dirs.toml
@@ -727,7 +727,7 @@ code = '''
     }
     lfs_unmount(&lfs) => 0;
 
-    for (int j = 2; j < COUNT; j++) {
+    for (int j = 2; j < COUNT + 2; j++) {
         lfs_mount(&lfs, &cfg) => 0;
         lfs_dir_open(&lfs, &dir, "hello") => 0;
         lfs_dir_read(&lfs, &dir, &info) => 1;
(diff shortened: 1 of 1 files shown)


geky:
Hi @sosthene-nitrokey, thanks for creating an issue.

I'm trying to dig into the bug, but I'm not sure this test was actually missing files?

It's a bit convoluted, but the lfs_dir_reads at the beginning of the loop should be skipping the `.` and `..` entries before the inner loop iterates over directory entries from `0` to `j`.

https://github.com/littlefs-project/littlefs/blob/6a53d76e90af33f0656333c1db09bd337fa75d23/tests/test_dirs.toml#L790-L814

Correct me if I'm wrong. I'll look into the error it's throwing.

geky:
Ah, the issue is it should be `i <= j`, so this does show the issue.

geky:
Well, no, the `i < j` iterates the all previous entries, the final `lfs_dir_tell` should hold the correct `pos`.

The `j = 2` is incorrect in the loop though, so we are missing the _first_ two entries. Though this doesn't reveal the underlying bug.

I think this needs its own test.

geky:
I went ahead and created #805, thanks for creating a PR!

sosthene-nitrokey:
Thank you for the investigation and the fix!

Feel free to close this PR as your PR also adds tests that cover this. This PR was meant mostly as an example reproducer, and not really with the goal to be merged.

PR #801, opened 2023-04-18
Add an assert for truthy-preserving bool conversions

geky:
This has caught enough people that an explicit assert is warranted.

How littlefs, a c99 project, should be integrated with c89 projects is still an open question, but no one deserves to debug this sort of undetected casting issue.

Found by @johnernberg and @XinStellaris. This an alternative to https://github.com/littlefs-project/littlefs/pull/772 and https://github.com/littlefs-project/littlefs/pull/677.


johnernberg (approved):
Thanks for doing this.

Issue #264, opened 2019-08-09
Minimal block_size


jrast:
What is the minimal block size? With 32 byte formatting fails (Error -28, No space left on device), with 64 byte formatting and read/write files worked in my short testcase. Is 64 Byte the minimum?

geky:
Short answer: 128 bytes

Long answer: The data structure for files can use at most this many bytes:

![image](https://camo.githubusercontent.com/0473bdb02825320b5f1e4a6876d6ae4575edb79e/68747470733a2f2f6c617465782e636f6465636f67732e636f6d2f7376672e6c617465783f422532302533442532302535436672616325374277253744253742382537442535436c6566742535436c6365696c2535436c6f675f322535436c656674253238253543667261632537423225354577253744253742422d3225354366726163253742772537442537423825374425374425354372696768742532392535437269676874253543726365696c)

For w = 32-bits, our minimum block size is 104 bytes. Rounding to 128 bytes gives us a bit of fudge room if we need to expand data structures in the future.

Though this check should be asserting if your block size is too small. Are you compiling with asserts disabled?
https://github.com/ARMmbed/littlefs/blob/494dd6673d008c03e3199b24af7bff0fe82017f1/lfs.c#L3378-L3380

Maybe I should change the assert to be a bit more explicit and just check against 128...

Out of curiosity, what is the need to make the blocks so small? Smaller blocks will improve metadata access, but unfortunately will also increase the allocator cost.

jrast:
> Short answer: 128 bytes

Maybe the commet for the `block_size` in the header file should be extended to include this information.

Are the asserts disabled by default? Because I did not change / disable anything. But actually I'm not working on an embedded system but on a python wrapper for littlefs (more on this later).

I used the small block size to see if littlefs would fit our needs on the embedded system. We only have a 8KB external EEPROM and therefore a small block size would be prefered. I once tried littlefs and it was easy to integrate, thats why I'm currently fiddeling around with the project to see how it would behave in our case.

e107steved:
[...]

PR #802, opened 2023-04-18
Add explicit assert for minimum block size of 128 bytes

geky:
See https://github.com/littlefs-project/littlefs/issues/264 for more info.

There was already an assert for this, but because it included the underlying equation for the requirement it was too confusing for users that had no prior knowledge for why the assert could trigger.

The math works out such that 128 bytes is a reasonable minimum requirement, so I've added that number as an explicit assert. Hopefully this makes this sort of situation easier to debug.

Note that this requirement would need to be increased to 512 bytes if block addresses are ever increased to 64-bits. [DESIGN.md#ctz-skip-lists](https://github.com/littlefs-project/littlefs/blob/master/DESIGN.md#ctz-skip-lists) goes into more detail why this is.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -3917,7 +3917,10 @@ static int lfs_init(lfs_t *lfs, const struct lfs_config *cfg) {
     LFS_ASSERT(lfs->cfg->cache_size % lfs->cfg->prog_size == 0);
     LFS_ASSERT(lfs->cfg->block_size % lfs->cfg->cache_size == 0);
 
-    // check that the block size is large enough to fit ctz pointers
+    // check that the block size is large enough to fit all ctz pointers
+    LFS_ASSERT(lfs->cfg->block_size >= 128);
+    // this is the exact calculation for all ctz pointers, if this fails
+    // and the simpler assert above does not, math must be broken
     LFS_ASSERT(4*lfs_npw2(0xffffffff / (lfs->cfg->block_size-2*4))
             <= lfs->cfg->block_size);
 


e107steved:
Seems a good idea to make the reason for the assert obvious (although would an explanatory comment suffice?).

However, a concern for the future. If you should ever make block addresses 64 bit, I hope this would be an option. The jump from 128 byte blocks to 512 byte blocks could be quite painful for those using small memories (I have just 128K bytes of FRAM, and I think other users have less!). The low overhead of littleFS is currently a key advantage, especially over FAT.
There would also be the issue of upgrading existing file systems.

geky:
> Seems a good idea to make the reason for the assert obvious (although would an explanatory comment suffice?).

Well this also makes the rule-of-thumb 128 byte limit a hard requirement, which may be valuable in future changes (though unlikely).

It also asserts the math is correct, which is an intimidating piece of code someone might think is broken if they hit this assert. It might reduce the mental load when debugging if you're also calculating your block size a complicated way.

Though that may be a bit of a stretch, it could probably go either way. At least if code size is an issue these asserts can always be turned off.

> However, a concern for the future. If you should ever make block addresses 64 bit, I hope this would be an option. The jump from 128 byte blocks to 512 byte blocks could be quite painful for those using small memories (I have just 128K bytes of FRAM, and I think other users have less!). The low overhead of littleFS is currently a key advantage, especially over FAT.
There would also be the issue of upgrading existing file systems.

It's just a hypothetical. As is, a 64-bit version. of littlefs would be fundamentally incompatible with the current 32-bit version. And the devices that would need 64-bits addresses would probably be very unlikely to have issues with 512 byte blocks.

Another hypothetical would be a 16-bit version of littlefs, which could in theory push the CTZ skip-list requirement down to 32-bytes (24 exactly). But I think at that point you'd really start having issues fitting file metadata into the block.

PR #805, opened 2023-04-18
Fix issue where seeking to end-of-directory return LFS_ERR_INVAL

geky:
This was just an oversight. Seeking to the end of the directory should not error, but instead restore the directory to the state where the next read returns 0.

Note this matches the behavior of lfs_file_tell/lfs_file_seek.

Found by @sosthene-nitrokey, see #785 for more info.


sosthene-nitrokey:
I can't speak for the fix itself but the tests appear to cover the case I describe in #785 and expect the same behaviour I would.

Issue #604, opened 2021-10-17
lfs_fs_forceconsistency()->lfs_fs_deorphan() doesn't commit updated gstate


kasper0:
Hi,
while debugging the lfs, I've noticed that lfs_fs_forceconsistency()->lfs_fs_deorphan() doesn't commit updated gstate after fixing orphans. The responsibility for saving new gstate lies with a function calling lfs_fs_forceconsistency(), but it fails if the function returns before making a commit (e.g. mkdir() when the directory already exists).

Unfortunately, on systems with limited resources and dozen of files, it takes a long time to perform the consistency check. As a result, if there weren't any operations making commits after the check, another consistency check will occur after a reboot or power loss. This time without fixing orphans, but traversing the files will take the same time (in my case 1600 files/ 30minutes).

Moreover, it would be great to be able to perform a consistency check directly from API without using the side effect of mkdir/open/remove/rename.

Ldd309:
Hi,
I have the same problem as you, do you have any solution?

geky:
Hi @kasper0, thanks for creating an issue, this is good feedback.

The motivation for delaying the gstate commit is that we effectively write the gstate for free in the common case of being followed by a commit. This can be significant on storage with large prog_sizes (SD, NAND, mostly).

If a function fails before it can commit, the pending gstate should at least still be saved in `lfs_t`, and written out on the next operation that commits. So you should only pay for `lfs_fs_forceconsistency` at most once per power-cycle. Let me know if you're seeing different behavior.

Though it makes sense to me to add a public `lfs_fs_forceconsistency` that can be called to 1. check consistency without other side effects and 2. commit the gstate eagerly.

Ldd309:
> Hi @kasper0, thanks for creating an issue, this is good feedback.
> 
> The motivation for delaying the gstate commit is that we effectively write the gstate for free in the common case of being followed by a commit. This can be significant on storage with large prog_sizes (SD, NAND, mostly).
> 
> If a function fails before it can commit, the pending gstate should at least still be saved in `lfs_t`, and written out on the next operation that commits. So you should only pay for `lfs_fs_forceconsistency` at most once per power-cycle. Let me know if you're seeing different behavior.
> 
> Though it makes sense to me to add a public `lfs_fs_forceconsistency` that can be called to 1. check consistency without other side effects and 2. commit the gstate eagerly.

[...]

PR #811, opened 2023-04-27
Fix issue where lfs_fs_deorphan may run more than needed

geky:
The underlying issue is that lfs_fs_deorphan did not updating gstate correctly. The way it determined if there are any orphans remaining in the filesystem was by subtracting the number of found orphans from an internal counter.

This internal counter is a leftover from a previous implementation that allowed leaving the lfs_fs_deorphan loop early if we know the number of expected orphans. This can happen during recursive mdir relocations, but with only a single bit in the gstate, can't happen during mount. If we detect orphans during mount, we set this internal counter to 1, assuming we will find at least one orphan.

But this presents a problem, what if we find _no_ orphans? If this happens we never decrement the internal counter of orphans, so we would never clear the bit in the gstate. This leads to a running lfs_fs_deorphan on more-or-less every mutable operation in the filesystem, resulting in an extreme performance hit.

The solution here is to not subtract the number of found orphans, but assume that when our lfs_fs_deorphan loop finishes, we will have no orphans, because that's the whole point of lfs_fs_deorphan.

Note that the early termination of lfs_fs_deorphan was dropped because it would not actually change the runtime complexity of lfs_fs_deorphan, adds code cost, and risks fragile corner cases such as this one.

---

Also added tests to assert we run lfs_fs_deorphan at most once.

Found by @kasper0 and @Ldd309
See https://github.com/littlefs-project/littlefs/issues/604 for more info

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -4741,8 +4741,6 @@ static int lfs_fs_deorphan(lfs_t *lfs, bool powerloss) {
         return 0;
     }
 
-    int8_t found = 0;
-
     // Check for orphans in two separate passes:
     // - 1 for half-orphans (relocations)
     // - 2 for full-orphans (removes/renames)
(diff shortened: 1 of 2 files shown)


geky:
Going ahead and bringing this in so it makes it into the next minor release

PR #812, opened 2023-04-27
Add lfs_fs_mkconsistent

geky:
`lfs_fs_mkconsistent` allows running the internal consistency operations (desuperblock/deorphan/demove) on demand and without any other filesystem changes.

This can be useful for front-loading and persisting consistency operations when you don't want to pay for this cost on the first write to the filesystem.

https://github.com/littlefs-project/littlefs/blob/259535ee73a792556bb16eebfbd076f467c5bdfa/lfs.h#L680-L688

This also offers a way to force the on-disk minor version to bump, if that is wanted behavior. An on-disk minor version bump is being added in https://github.com/littlefs-project/littlefs/pull/497, so this is convenient timing.

---

Not entirely sure about the name, open to suggestions though I think this is an improvement over `lfs_fs_forceconsistency` which was really meant to be internal only and may imply abnormal behavior.

Idea from @kasper0
See https://github.com/littlefs-project/littlefs/issues/604 for more info

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -4887,6 +4887,36 @@ static int lfs_fs_forceconsistency(lfs_t *lfs) {
 }
 #endif
 
+#ifndef LFS_READONLY
+int lfs_fs_rawmkconsistent(lfs_t *lfs) {
+    // lfs_fs_forceconsistency does most of the work here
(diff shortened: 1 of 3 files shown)


geky:
Going ahead and bringing this in so it makes it into the next minor release

PR #819, opened 2023-05-05
Fix build for AVR

benpicco:
This fixes the overflowing left shift on 8 bit platforms.

    littlefs2/lfs.c: In function ‘lfs_dir_commitcrc’:
    littlefs2/lfs.c:1654:51: error: left shift count >= width of type [-Werror=shift-count-overflow]
             commit->ptag = ntag ^ ((0x80 & ~eperturb) << 24);

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -1651,7 +1651,7 @@ static int lfs_dir_commitcrc(lfs_t *lfs, struct lfs_commit *commit) {
 
         commit->off = noff;
         // perturb valid bit?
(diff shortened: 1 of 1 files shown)


geky:
Hi @benpicco, thanks for creating a PR. Will merge this into the next patch release.

I think 8/16-bit platforms will continue to be a bit fragile like this until we have a build in CI, but I see no reason to not bring this in.

geky (approved):

geky:
Thanks again for the PR!

PR #830, opened 2023-05-16
Fix compilation with Clang

shacron:
Avoid passing GCC-specific flags to non-GCC $CC
Build with Clang on Darwin targets.
Fix a print format warning when an int is passed to format PRIu16


shacron:
Ugh, broken if `$(CC) = cc`. Fix forthcoming.

PR #855, opened 2023-07-11
initlize struct lfs_diskoff disk = {0}

mdahamshi:
so we don't use it uninitlized in first run


geky:
Hi @mdahamshi, thanks for creating a PR.

It's worth noting that this struct does not get used unless the top-bit of the tag is set, which only happens when we initialize the struct's contents.

But we can bring this in to at least silence the warning.

geky (approved):

geky:
Thanks again for this, and everyone else who reported/proposed a fix. Sorry about taking so long to merge. There was some integration issues in the dependent littlefs-fuse I wanted to fix before making another release.

PR #859, opened 2023-07-25
init disk instance before use

ThomasAUB:
under some compilers (e.g. MSVC) the variable may not be init to 0 which can lead to a crash at line 918

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -878,6 +878,8 @@ static int lfs_dir_traverse(lfs_t *lfs,
     lfs_tag_t tag;
     const void *buffer;
     struct lfs_diskoff disk;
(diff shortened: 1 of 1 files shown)


geky:
Hi @ThomasAUB, thanks for creating a PR.

It's worth noting that uninitialized struct does go unused. It gets stored to an explicit stack, which is probably why MSVC is tripping over it, but it doesn't get accessed unless the highest bit in the tag is set, which only occurs when we assign contents to the struct.

But we can go ahead an merge this to fix the warning.

Though it looks like @mdahamshi made a PR earlier, so will go with that one: https://github.com/littlefs-project/littlefs/pull/855

PR #873, opened 2023-09-14
avoid a variable shadowing warning

yamt:


geky:
Hi @yamt, thanks for creating a PR.

I'm not a fan of -Wshadow, I believe it does more harm than good. (Why would you encourage programmers to increase a variable's scope more than necessary?).

My current stance is to use GCC/Clang with -Wall -Wextra -pedantic and not use -Wshadow unless it is added to this set of default warnings in GCC or Clang.

I could merge this, but without -Wshadow in littlefs's CI it's likely a shadowed variable gets reintroduced in another patch. I believe the correct thing to do is disable the warning locally when compiling littlefs.

PocketPi:
> Hi @yamt, thanks for creating a PR.
> 
> I'm not a fan of -Wshadow, I believe it does more harm than good. (Why would you encourage programmers to increase a variable's scope more than necessary?).
> 
> My current stance is to use GCC/Clang with -Wall -Wextra -pedantic and not use -Wshadow unless it is added to this set of default warnings in GCC or Clang.
> 
> I could merge this, but without -Wshadow in littlefs's CI it's likely a shadowed variable gets reintroduced in another patch. I believe the correct thing to do is disable the warning locally when compiling littlefs.

I would love to have this PR merged as well.

And I do not agree that Wshadow "encourage programmers to increase a variable's scope more than necessary" imo it helps. If you want to keep the scope small then its better to change the variable name but still declare it in the small scope.

yamt:
actually, it's quite the opposite to "encourage programmers to increase a variable's scope more than necessary".
you can avoid the warning by reducing the scope of the shadowed variable so that it doesn't overlap with the shadowing one.

BenBE:
Looking at the overall structure of the function, there are several instances of a variable called `err`. The one warning the compiler gives with `-Wshadow` is really just a symptom of this; and the patch only addresses part of the (actual) issue.

There are a few possible ways to go about this:
1. Move `err` to the outermost scope of the function and use it as a "global" temporary to store error state information (I don't like this, because this messes up scopes)
2. Narrow down the scope for each instance where `err` is used, by introducing a bare pair of braces around it. While this may look strange at first, this is most in spirit with what @geky said re reducing scopes
3. Rename each instance of `err` to have it's own variable name

Overall, I'd prefer to see the second one implemented with `-Wshadow` being added to the list of CI flags.

P.S.: @geky There are several quite useful compiler flags not present in `-Wall -Wextra -pedantic` that can provide an early notification about issues. Some of these include:

- [-Wwrite-strings](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wwrite-strings)
- [-Wcast-align](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wcast-align)
- [-Wcast-qual](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wcast-qual)
- [-Wpointer-arith](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wpointer-arith)
- [-Wstrict-aliasing](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wstrict-aliasing)
- [-Wformat=2](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wformat)
- [-Wuninitialized](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wuninitialized)

BenBE:
From my experience, using quite a few extra flags in my code, the disadvantages are rare and I mostly see them as an extra layer of defense myself.

That said, together with a few more folks, I'll be looking through the littlefs source in the upcoming weeks*. What we have looked at so far looks promising (corse overview so far only), but following the issue tracker there are likely some places in the code we might need to follow up on. We'll provide feedback on our findings as soon as there's something noteworthy. @geky If you have any particular areas of interest that we should take a look at, feel free to ping me.

*usually only a few hours per week, idea based on [this presentation](https://media.ccc.de/v/rc3-2021-cwtv-228-lets-review-code-toget) (sorry, German only).

geky:
More eyes are always welcome.

I'm not sure what areas would be best to look at though, maybe looking through the issue tracker for bugs where enough info is available would lead to something interesting? littlefs has become relatively stable recently, all things considered, short of changes that would require major API breakage.

Larger functionality work is currently going on in experimental branches, but those may be best described as [TODO-soup](https://github.com/littlefs-project/littlefs/tree/rbyd) at the moment.

PR #884, opened 2023-10-23
lfs_fs_raw* functions should be static

DvdGiessen:
Two new functions introduced in 2.6.0 (`lfs_fs_mkconsistent`, 259535ee73a792556bb16eebfbd076f467c5bdfa) and 2.8.0 (`lfs_fs_grow`, 23505fa9fa4f2d863f046c0bfad8db2ca63cfb90) have internal "raw" functions which are missing the `static` keyword.

This was causing build failures in our builds in the MicroPython project when updating to the latest LittleFS, because we're building with `-Werror,-Wmissing-prototypes,-Wmissing-declarations`. See [here](https://github.com/micropython/micropython/actions/runs/6561913228/job/17822696395) and [here](https://github.com/micropython/micropython/actions/runs/6575701921/job/17863464790) for the build output.

This PR adds the `static` keyword to these functions to fix these, similar to how all other internal functions in LittleFS are defined.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,6 +63,7 @@ CFLAGS += -fcallgraph-info=su
 CFLAGS += -g3
 CFLAGS += -I.
 CFLAGS += -std=c99 -Wall -Wextra -pedantic
+CFLAGS += -Wmissing-prototypes
 CFLAGS += -ftrack-macro-expansion=0
 ifdef DEBUG
 CFLAGS += -O0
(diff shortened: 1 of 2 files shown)


geky:
Hi @DvdGiessen, thanks for this! Looks like quite an oversight.

Will bring this in next patch release.

geky (approved):

geky:
I've gone ahead and added -Wmissing-prototypes into our Makefile, it seems like a useful warning (https://github.com/littlefs-project/littlefs/pull/884/commits/8f3f32d1f31e9b233a22eafa04257b5c9a073160). This should now be caught by the CI here and prevented in the future.

geky:
Should be merged now, thanks again for this!

PR #886, opened 2023-10-29
Add value-range checks for user-definable macros at compile-time

BrianPugh:
Added some sanity check for the user-providable `#define` macro values.

This was brought up due to a bug introduced by yours truly 💅  https://github.com/jrast/littlefs-python/issues/61

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -4112,8 +4112,8 @@ static int lfs_rawremoveattr(lfs_t *lfs, const char *path, uint8_t type) {
 #error "Invalid LFS_NAME_MAX, must be <= 1022"
 #endif
 
-#if LFS_FILE_MAX > 4294967295
-#error "Invalid LFS_FILE_MAX, must be <= 4294967295"
+#if LFS_FILE_MAX > 2147483647
+#error "Invalid LFS_FILE_MAX, must be <= 2147483647"
 #endif
 
 #if LFS_ATTR_MAX > 1022
--- a/lfs.c
+++ b/lfs.c
@@ -8,23 +8,6 @@
 #include "lfs.h"
 #include "lfs_util.h"
 
-// Configuration Sanity Check
-#if (LFS_NAME_MAX <= 0) || (LFS_NAME_MAX > 1022)
-#error "LFS_NAME_MAX must be in the range (0, 1022]"
-#endif
-
-#if (LFS_FILE_MAX <= 0) || (LFS_FILE_MAX > 4294967295)
-#error "LFS_FILE_MAX must be in the range (0, 4294967295]"
-#endif
-
-#if (LFS_FILE_MAX > 2147483647)
-#warning "LFS_FILE_MAX>2147483647; lfs_file_seek, lfs_file_size, and lfs_file_tell will not function properly."
-#endif
-
-#if (LFS_ATTR_MAX < 0) || (LFS_ATTR_MAX > 1022)
-#error "LFS_ATTR_MAX must be in the range [0, 1022]"
-#endif
-
 
 // some constants used throughout the code
 #define LFS_BLOCK_NULL ((lfs_block_t)-1)
@@ -4123,6 +4106,21 @@ static int lfs_rawremoveattr(lfs_t *lfs, const char *path, uint8_t type) {
 
 
 /// Filesystem operations ///
+
+// compile time checks, see lfs.h for why these limits exist
+#if LFS_NAME_MAX > 1022
+#error "Invalid LFS_NAME_MAX, must be <= 1022"
+#endif
+
+#if LFS_FILE_MAX > 4294967295
+#error "Invalid LFS_FILE_MAX, must be <= 4294967295"
+#endif
+
+#if LFS_ATTR_MAX > 1022
+#error "Invalid LFS_ATTR_MAX, must be <= 1022"
+#endif
+
+// common filesystem initialization
 static int lfs_init(lfs_t *lfs, const struct lfs_config *cfg) {
     lfs->cfg = cfg;
     lfs->block_count = cfg->block_count;  // May be 0
--- a/lfs.c
+++ b/lfs.c
@@ -8,6 +8,23 @@
 #include "lfs.h"
 #include "lfs_util.h"
 
+// Configuration Sanity Check
+#if (LFS_NAME_MAX <= 0) || (LFS_NAME_MAX > 1022)
+#error "LFS_NAME_MAX must be in the range (0, 1022]"
+#endif
+
+#if (LFS_FILE_MAX <= 0) || (LFS_FILE_MAX > 4294967296)
+#error "LFS_FILE_MAX must be in the range (0, 4294967296]"
+#endif
+
+#if (LFS_FILE_MAX > 2147483647)
+#warning "LFS_FILE_MAX>2147483647; lfs_file_seek, lfs_file_size, and lfs_file_tell will not function properly."
+#endif
+
+#if (LFS_ATTR_MAX < 0) || (LFS_ATTR_MAX > 1022)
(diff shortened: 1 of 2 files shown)


lfs.c
```diff
@@ -8,6 +8,23 @@
...
 #include "lfs_util.h"
 
+// Configuration Sanity Check
+#if (LFS_NAME_MAX <= 0) || (LFS_NAME_MAX > 1022)
+#error "LFS_NAME_MAX must be in the range (0, 1022]"
+#endif
+
+#if (LFS_FILE_MAX <= 0) || (LFS_FILE_MAX > 4294967296)
+#error "LFS_FILE_MAX must be in the range (0, 4294967296]"
+#endif
```
BenBE, line 18: Are you sure this is `4294967296` (AKA `2^32`) and not 4294967295 (`UINT32_MAX`)?

lfs.c
```diff
@@ -8,6 +8,23 @@
...
+
+#if (LFS_FILE_MAX > 2147483647)
+#warning "LFS_FILE_MAX>2147483647; lfs_file_seek, lfs_file_size, and lfs_file_tell will not function properly."
```
BenBE, line 21: This is only valid in [C23](https://en.cppreference.com/w/c/preprocessor/error) and compiler-dependent before that.

Also, given that some people opt to use `-Werror` for compilation, this unintentionally breaks their builds.

BenBE (commented):
Would be nice, to reference the design/spec for the 1022 value (I know, where it comes from, but without that ref, a casual reader of the source may believe this to be in error.

lfs.c
BrianPugh: I imagine your proposed `4294967295` is correct, but I was just copying from [the documentation](https://github.com/littlefs-project/littlefs/blob/f77214d1f0a8ccd7ddc7e1204fedd25ee5a41534/lfs.h#L55). Once we confirm thats the correct value, I can update both locations.

lfs.c
BrianPugh: I was a little hesitant on this as well; not sure how frequently users change `LFS_FILE_MAX`, would be open to alternative proposals (or even just fully removing this warning).

BrianPugh:
> Would be nice, to reference the design/spec for the 1022 value (I know, where it comes from, but without that ref, a casual reader of the source may believe this to be in error.

It's actually not referenced in `SPEC.md` or anywhere else beyond `lfs.h`. I could say "reference lfs.h" in the warning if we'd like.

BenBE:
> > Would be nice, to reference the design/spec for the 1022 value (I know, where it comes from, but without that ref, a casual reader of the source may believe this to be in error.
> 
> It's actually not referenced in `SPEC.md` or anywhere else beyond `lfs.h`. I could say "reference lfs.h" in the warning if we'd like.

Rechecked the `DESIGN.md` and `SPEC.md` and they both only mention the bit-field size this value is used in and some special values that need to be avoided. Would be nice to have those consequences to limits pointed out a bit more directly in the design/spec docs; not just in the header/implementation.

lfs.c
geky: Ah woops, yes this is a typo. 4294967295 (UINT32_MAX) should be the correct value.

lfs.c
geky: Curiously gcc does not warn about this, usually `gcc -Werror -std=c99 -Wall -Wextra -pedantic` warns about c99 compat issues. Otherwise CI would have caught this.
https://github.com/littlefs-project/littlefs/actions/runs/6685719140/job/18164313418?pr=886#step:7:50

I wonder why this is...

But @BenBE is right, there's already enough requests for C89 support, we need to stick to C99.

---

Alternative proposal: we can just remove the $&gt;2^{31}$ support. I think realistically no one is using this. It's already only partially supported and untested.

It's a desperate attempt to stretch out what 32-bits can do, but I already have some plans to eventually add a 64-bit mode, and that will probably be a better way to support larger file sizes when needed.

[Worst case](https://xkcd.com/1172/), if someone does depend on this we can always revert.

geky:
Point taken SPEC.md could document these limits better, but I don't think lfs.h/lfs.c should reference SPEC.md. They should both contain copies of all relevant documentation so a reader doesn't need to jump between documents.

If lfs.h/lfs.c and SPEC.md fall out of sync you have bigger problems.

I'm curious, is there an issue with putting these in lfs.h? Then they could be next to the relevant comments.

If not (multiple include warning spam?), then maybe just a comment saying something like "Must fit in 10-bits with 1023 reserved, i.e. <= 1022." is sufficient.

lfs.c
geky, line 21: nit: Must fit in 80-column width:

``` c
#warning "LFS_FILE_MAX>2147483647; lfs_file_seek, lfs_file_size, and \
lfs_file_tell will not function properly."                            
```

I should really get around to adding a lint for this...

BrianPugh:
> I'm curious, is there an issue with putting these in lfs.h? Then they could be next to the relevant comments.

I wanted to keep the header file clean and focus on the library API. I didn't want to crowd the header file with sanity checks, but I'm open to either option.

lfs.c
BrianPugh: addressed in c531a5e88f09e71f9487

lfs.c
BrianPugh: I'm in favor of removing `>2**31` support for the reasons you outlined.

geky:
> I wanted to keep the header file clean and focus on the library API. I didn't want to crowd the header file with sanity checks, but I'm open to either option.

That makes sense, I think you're right these should probably stay in the .c file.

Controversial thought: What if these were runtime asserts and live in lfs_init? That way they live with all of the other configuration-level asserts. And while they do have a cost at runtime... in theory runtime asserts have no cost when it matters.

---

Unrelated, how badly do you want this in the next minor release? I think think the next minor release can be cut pretty much whenever.

BrianPugh:
> What if these were runtime asserts and live in lfs_init? That way they live with all of the other configuration-level asserts.

You mean just from an organizational standpoint? Seems fine by me. This wouldn't even make it to the compiled code since these all get handled by the preprocessor, so there wouldn't be any runtime impact.

> how badly do you want this in the next minor release?
I don't have any strong opinions, would just like these asserts to make it into the codebase at some point so other people avoid making stupid mistakes 😄

geky:
> You mean just from an organizational standpoint? Seems fine by me.

I was thinking more make these actual runtime asserts, but I've compromised by just moving them near lfs_init. Compile-configuration needs a rework anyways (https://github.com/littlefs-project/littlefs/pull/491 is related).

I've updated this PR with what was discussed in the PR. Feel free to let me know if anything looks wrong. I think this can make it into the next minor release.

One maybe controversial change: I removed the check against <=0:

- LFS_FILE_MAX <= 0, zero-sized files are technically different from not-created files. It is theoretically possible to have a filesystem where every file is effectively a boolean flag of does/does-not exist. Is this reasonable? probably not. But no reason we should disallow it on principle.

- LFS_NAME_MAX <= 0, a little bit harder to argue for, but _technically_ you _could_ create a zero-length filename in each directory. As long as the directory's filename is also zero-length. Which would imply the entire filesystem is a linked-list of zero-length directories terminating in a single file, but _reasonableness_ aside, I don't think we should assert in this case. It's not an error as far as littlefs is concerned, just crippling and weird.

Anyways, let me know if anything looks concerning, otherwise I'm happy to merge this PR into the next minor release.

BrianPugh:
Looks good to me, I think your reasoning is sound. The upper limits are much more easily broken by accident, and are the more important checks.

geky (approved):

geky:
This is on master now, thanks for the PR!

PR #911, opened 2023-12-20
Fix struct sizes missing from generated release notes

geky:
This script was missed during a struct -> structs naming change.


geky:
Updated affected release notes manually

PR #912, opened 2023-12-20
Relaxed lookahead alignment, other internal block alloc readability improvements

geky:
This drops the lookahead buffer from operating on 32-bit words to operating on 8-bit bytes, and removes any alignment requirement. This may have some minor performance impact, but it is unlikely to be significant when you consider IO overhead.

The original motivation for 32-bit alignment was an attempt at future-proofing in case we wanted some more complex on-disk data structure. This never happened, and even if it did, it could have been added via additional config options.

This has been a significant pain point for users, as providing word-aligned byte-sized buffers in C can be a bit annoying.

---

This also includes a number of internal block allocator improvements, mostly around naming things and making the logic hopefully easier to understand.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -600,7 +600,7 @@ static int lfs_alloc_lookahead(void *p, lfs_block_t block) {
             + lfs->block_count) % lfs->block_count;
 
     if (off < lfs->lookahead.size) {
-        lfs->lookahead.buffer[off / 32] |= 1U << (off % 32);
+        lfs->lookahead.buffer[off / 8] |= 1U << (off % 8);
     }
 
     return 0;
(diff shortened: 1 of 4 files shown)


HiFiPhile:
Hi, could you please also update comment in `lfs_util.h` ?

Especially it still mentions 64-bit alignment for malloc requirement:
https://github.com/littlefs-project/littlefs/blob/3513ff1afc1d67adb2e6f492f0b9bc0d798fcb0d/lfs_util.h#L218

geky:
Good catch, should be updated now.

There is a risk that we will need word-alignment in the future, but I guess we'll address that when we get to that.

PR #917, opened 2024-01-03
Fix return value of lfs_rename()

tomscii:
When lfs_rename() is called trying to rename (move) a file to an existing directory, LFS_ERR_ISDIR is (correctly) returned. However, in the opposite case, if one tries to rename (move) a directory to a path currently occupied by a regular file, LFS_ERR_NOTDIR should be returned (since the error is that the destination is NOT a directory), but in reality, LFS_ERR_ISDIR is returned in this case as well.

This commit fixes the code so that in the latter case, LFS_ERR_NOTDIR is returned.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -3940,7 +3940,9 @@ static int lfs_rawrename(lfs_t *lfs, const char *oldpath, const char *newpath) {
             newoldid += 1;
         }
     } else if (lfs_tag_type3(prevtag) != lfs_tag_type3(oldtag)) {
-        return LFS_ERR_ISDIR;
+        return (lfs_tag_type3(prevtag) == LFS_TYPE_DIR)
+                ? LFS_ERR_ISDIR
+                : LFS_ERR_NOTDIR;
     } else if (samepair && newid == newoldid) {
         // we're renaming to ourselves??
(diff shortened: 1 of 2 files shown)


geky:
Hi @tomscii, thanks for this. This does look more correct in terms of following POSIX.

The only concern is if this will break any existing code. Code may be relying on LFS_ERR_NOTDIR for both cases.

I'll plan on bringing this in on the next minor release. If it's an issue, we'll learn, otherwise I think the strategy may be to allow error codes to change to match POSIX on minor releases.

geky (approved):

geky:
I've gone ahead and updating this PR for two minor reasons:
1. Updated to match style, mainly double-indention (8-spaces) in expressions. I realize this isn't really documented anywhere.
2. Added tests to ensure that the rename type mismatch returns this error as expected.

Let me know if anything looks concerning.

tomscii:
@geky Looks great, thank you!

geky:
This is on master now, thanks for the PR!

tomscii:
> This is on master now, thanks for the PR!

Excellent! Thanks for all your hard work maintaining LittleFS!

PR #921, opened 2024-01-16
Tighten bound on mlist isopen asserts

geky:
Unbalanced open/close calls continue to be a pain point for users, it doesn't help that this sometimes results in hard-to-debug infinite loops caused by the open-file linked-list (the mlist) getting tangled up in itself.

Moving the mlist isopen asserts lower into the actual list append/remove functions will help:

1. Make sure coverage of potential linked-list issues is complete.

2. Also assert against multiple close calls, which isn't an issue for the mlist, but can result in double free and memory corruption.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -505,6 +505,7 @@ static bool lfs_mlist_isopen(struct lfs_mlist *head,
 #endif
 
 static void lfs_mlist_remove(lfs_t *lfs, struct lfs_mlist *mlist) {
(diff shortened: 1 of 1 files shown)


geky:
It looks like this doesn't work because of the restriction of LFS_ASSERT to error-returning functions (to allow asserts to return errors).

Tabling this for now. Current thinking is we should eventually split LFS_ASSERT into LFS_ASSERT and LFS_DISK_ASSERT, with only LFS_DISK_ASSERT being limited to error-returning functions.

Issue #920, opened 2024-01-09
lfs_dir_getslice(): why do we subtract gdiff from gtag?


andriyndev:
I cannot understand this exact detail. From what I understand, we are looking for a tag that matches masked `gtag` (pattern), why not then we just use `gtag` instead of `gtag - gdiff`? And what is actually `gdiff` (don't understand how and why it's calculated too)?

andriyndev:
@geky please, help!

andriyndev:
By the way

> 0x401 LFS_TYPE_CREATE
> 
> Creates a new file with this id. Note that files in a metadata block don't necessarily need a create tag. All a create does is move over any files using this id. In this sense a create is similar to insertion into an imaginary array of files.
> 
> The create and delete tags allow littlefs to keep files in a directory ordered alphabetically by filename.

Do I understand correctly that it might contain an already used id, and in this case ids of already present files which bigger or equal to the "new" id, are incremented by 1, and the actual ids are calculated in runtime based on the whole data of the metadata blocks related to the folder, while the ids of older metadata entries remain unchanged on the disk? As an example, we can have 10 `LFS_TYPE_CREATE` metadata entries with id == 0 while the actual id of these files are 9, 8, 7, 6, 5, 4, 3, 2, 1, 0 respectively, and if we add one more `LFS_TYPE_CREATE` metadata entry with id == 5, the ids will become 10, 9, 8, 7, 6, 4, 3, 2, 1, 0, 5?

geky:
> Do I understand correctly, that when we delete a file, the id of each file, created after it, decreased by 1?

Yep!

> If yes, then what exactly the id field of a tag contains? For example, we create a file, then delete it, and then create another file, what will be the id of the second file - 0 or 1 (actual and written in the corresponding tag field)?

The second file will have id=0:

```
id             0 1
--------------------
create a id=0  .
               |
delete a id=0  '

create b id=0  .
               b
```

I probably should have annotated what ids were written to disk in the above example. Here's some more operations:

```
id        0 1 2
---------------
create a id=0  .
               |
create b id=1  | .
               | |
create d id=2  | | .
               | | |
delete a id=0  ' / /
                / /
create c id=1  | .\
               | | \
create a id=0  .\ \ \
               | | | |
delete a id=0  ' / / /
                / / /
               b c d

getslice(id=0) => file b
getslice(id=1) => file c
getslice(id=2) => file d
```

[...]

PR #937, opened 2024-02-04
Fix synthetic move underflows in lfs_dir_get

geky:
By "luck" the previous code somehow managed to not be broken, though it was possible to traverse the same file twice in `lfs_fs_traverse/size` (which is not an error).

The problem was an underlying assumption in `lfs_dir_get` that it would never be called when the requested id is pending removal because of a powerloss. The assumption was either:

1. `lfs_dir_find` would need to be called first to find the id, and it would correctly toss out pending-rms with LFS_ERR_NOENT.

2. `lfs_fs_mkconsistent` would be implicitly called before any filesystem traversals, cleaning up any pending-rms. This is at least true for allocator scans.

But, as noted by @andriyndev, both `lfs_fs_traverse` and `lfs_fs_size` can call `lfs_fs_get` with a pending-rm id if called in a readonly context.

---

By "luck" this somehow manages to not break anything:

1. If the pending-rm id is >0, the id is decremented by 1 in `lfs_fs_get`, returning the previous file entry during traversal. Worst case, this reports any blocks owned by the previous file entry twice.

   Note this is not an error, `lfs_fs_traverse/size` may return the same block multiple times due to underlying copy-on-write structures.

2. More concerning: If the pending-rm id is 0, the id is decremented by 1 in `lfs_fs_get` and underflows. This underflow propagates into the type field of the tag we are searching for, decrementing it from `0x200` (LFS_TYPE_STRUCT) to `0x1ff` (LFS_TYPE_INTERNAL(UNUSED)).

   Fortunately, since this happens to underflow to the INTERNAL tag type, the type intended to never exist on disk, we should never find a matching tag during our `lfs_fs_get search`. The result? `lfs_dir_get` returns LFS_ERR_NOENT, which is actually what we want.

Also note that LFS_ERR_NOENT does not terminate the mdir traversal early. If it did we would have missed files instead of duplicating files, which is a slightly worse situation.

---

The fix is to add an explicit check for pending-rms in `lfs_dir_get`, just like in `lfs_dir_find`. This avoids relying on unintended underflow propagation, and should make the internal API behavior more consistent.

This is especially important for potential future gc extensions.

Related issue: https://github.com/littlefs-project/littlefs/issues/920
Found by: @andriyndev

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -710,11 +710,14 @@ static lfs_stag_t lfs_dir_getslice(lfs_t *lfs, const lfs_mdir_t *dir,
     lfs_tag_t ntag = dir->etag;
     lfs_stag_t gdiff = 0;
 
+    // synthetic moves
     if (lfs_gstate_hasmovehere(&lfs->gdisk, dir->pair) &&
-            lfs_tag_id(gmask) != 0 &&
-            lfs_tag_id(lfs->gdisk.tag) <= lfs_tag_id(gtag)) {
-        // synthetic moves
-        gdiff -= LFS_MKTAG(0, 1, 0);
+            lfs_tag_id(gmask) != 0) {
+        if (lfs_tag_id(lfs->gdisk.tag) == lfs_tag_id(gtag)) {
+            return LFS_ERR_NOENT;
+        } else if (lfs_tag_id(lfs->gdisk.tag) < lfs_tag_id(gtag)) {
+            gdiff -= LFS_MKTAG(0, 1, 0);
+        }
     }
 
     // iterate over dir block backwards (for faster lookups)


lfs.c
```diff
@@ -710,11 +710,14 @@ static lfs_stag_t lfs_dir_getslice(lfs_t *lfs, const lfs_mdir_t *dir,
     lfs_tag_t ntag = dir->etag;
     lfs_stag_t gdiff = 0;
 
+    // synthetic moves
     if (lfs_gstate_hasmovehere(&lfs->gdisk, dir->pair) &&
-            lfs_tag_id(gmask) != 0 &&
-            lfs_tag_id(lfs->gdisk.tag) <= lfs_tag_id(gtag)) {
-        // synthetic moves
-        gdiff -= LFS_MKTAG(0, 1, 0);
+            lfs_tag_id(gmask) != 0) {
```
andriyndev, line 715: I think the condition could be written in one line but up to you.

andriyndev:
LGTM

geky:
Sorry for the delay, this should be merged now assuming CI passes.

PR #966, opened 2024-04-17
Fix DivideByZero exception when filesystem is completely full.

BrianPugh:
There were reports of a `DivideByZero` exception in [esp_littlefs](https://github.com/joltwallet/esp_littlefs) when the filesystem was completely full. @huming2207 was able to get a [stacktrace](https://github.com/joltwallet/esp_littlefs/issues/183#issuecomment-2060156353), and it points to the logging line in this PR. We're still figuring out if this is the certainly the cause, but the current code looks like an overlooked bug from #912. 

Might address https://github.com/joltwallet/esp_littlefs/issues/183

Original original thread: https://esp32.com/viewtopic.php?f=13&t=39197&p=130708

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -688,7 +688,7 @@ static int lfs_alloc(lfs_t *lfs, lfs_block_t *block) {
         if (lfs->lookahead.ckpoint <= 0) {
             LFS_ERROR("No more free space 0x%"PRIx32,
                     (lfs->lookahead.start + lfs->lookahead.next)
-                        % lfs->cfg->block_count);
+                        % lfs->block_count);
             return LFS_ERR_NOSPC;
         }
 
--- a/lfs.h
+++ b/lfs.h
@@ -204,6 +204,7 @@ struct lfs_config {
     lfs_size_t block_size;
 
     // Number of erasable blocks on the device.
+    // If 0, will attempt to infer block_count from existing filesystem.
     lfs_size_t block_count;
 
     // Number of erase cycles before littlefs evicts metadata logs and moves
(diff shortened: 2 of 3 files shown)


geky:
Hi @BrianPugh, thanks for this.

Yeah, it looks like a simple mistake. I think what happened is I rebased #912 from a branch started before #866, and the `lfs->cfg->block_size` -> `lfs->block_size` change technically wasn't a merge conflict.

Will merge, though we should probably have a test to prevent regression.

geky:
A quick grep shows `lfs->cfg->block_size` is only used in `lfs_mount_` and `lfs_migrate_`, so hopefully this is the only case of this.

huming2207:
@geky as per our quick tests, by changing the lfs->cfg->block_size to lfs->block_size will fix the crash for us.

geky (approved):

geky:
I've gone ahead and extended test_alloc to test with an inferred block count. It's a bit hacky since our test runner doesn't really expect config to be mutable, but it should work.

I think this is worthwhile as the block allocator is one of subsystems more likely to run into block_count-related edge cases.

I also tweaked the `struct lfs_config` comments for consistency and to avoid on-disk/off-disk ambiguity. But it's good to mention the inferring behavior of `block_count=0`.

Let me know if anything looks concerning.

geky:
A bit humorous this only affects an LFS_DEBUG statement.

Many users turn these off for code savings so that likely made this issue less common.

BrianPugh:
all looks good to me!

PR #967, opened 2024-04-17
Limit release script to only draft releases

geky:
We probably always want a human involved in the final step of publishing a release. 1. To allow editing of the release notes before notifying users, and 2. to avoid any issues with rogue scripts (who doesn't want 1000 release notifications?).

Especially now that releases may trigger additional post-release scripts.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -254,7 +254,7 @@ jobs:
               tag_name: env.LFS_VERSION,
               name: env.LFS_VERSION | rtrimstr(".0"),
               target_commitish: "${{github.event.workflow_run.head_sha}}",
-              draft: env.LFS_VERSION | endswith(".0"),
+              draft: true,
               body: $release,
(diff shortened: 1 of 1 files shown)


geky:
Reverting/yanking - GitHub makes it way too difficult to notice pending draft releases.

1. They create no notifications, even if they include an @geky
2. They don't show up on the repo's homepage, you must navigate to the releases page
3. It's difficult to notice when master/release are out of sync, there's nothing really relating the two on the repo's homepage
4. You can't even subscribe to the release action, action notifications are all-or-none

This makes it very easy to forget pending releases generated ~an hour after merging PRs in less-frequented repos.

The alternative, marking generated releases as a "pre-release", seems to have the opposite problem: https://github.com/orgs/community/discussions/4993

There really needs to be _some_ way to receive notifications on draft releases. Until then automatically releasing patch releases seems like the best compromise.

PR #988, opened 2024-06-03
Allow CRC-32 match value override

jmalmari:
LFS format failed because CRC=0 is, in the general case, a wrong value to test against. For example, valid zlib CRC-32 match value would be 0x2144df1c.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -1775,7 +1775,7 @@ static int lfs_dir_commitcrc(lfs_t *lfs, struct lfs_commit *commit) {
         return err;
     }
 
-    if (crc != 0) {
+    if (crc != LFS_CRC_MATCH) {
         return LFS_ERR_CORRUPT;
     }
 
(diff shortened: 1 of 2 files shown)


geky:
Hi @jmalmari, thanks for creating a PR.

This is probably equivalent to inverting the crc before and after calculation:

```c
uint32_t lfs_crc(uint32_t crc, const void *buffer, size_t size) {
    return ~my_crc(~crc, buffer, size);
}
```

If you're looking to use a different polynomial/checksum, that's a bit of a different question that would create compatibility issues for other littlefs tooling. Required a source-level change is probably warranted in this case for now.

jmalmari:
Right, I did not spot that connection. User side change is probably better. Takes only a few extra cycles.

geky:
To be fair, the inverted form is more common, since it can also detect prefixed zero bytes.

Long term, it would be nice to move to the inverted form (or even a better polynomial, such as crc32c), but since this would require disk-breaking changes, it's sort of out of the question.

PR #1004, opened 2024-07-10
Add an alternative way to override LFS_MALLOC etc

yamt:
With the existing method, (-DLFS_MALLOC=my_malloc) users often had to use compiler options like -include, which was not so portable.
This change introduces another way to provide partial overrides of lfs_util.h using a user-provided header.


geky:
Hi @yamt, thanks for creating a PR, and sorry about such a late response.

This is a clever naming solution, I was stuck and couldn't think of a name since LFS_CONFIG was taken. These should have probably been named LFS_CONFIG/LFS_UTIL but that's with hindsight.

Will bring this in on the next minor/patch release (is it unreasonable to slip this in on a patch release?)

lfs_util.h
```diff
@@ -8,18 +8,36 @@
...
+//
+// #include <stddef.h>
+// extern void *my_malloc(size_t sz);
+// #define LFS_MALLOC(sz) my_malloc(sz)
+//
+// And build littlefs with the header by defining LFS_USER_DEFINES.
+// (-DLFS_USER_DEFINES=my_defines.h)
+
+#ifdef LFS_USER_DEFINES
+#include LFS_STRINGIZE(LFS_USER_DEFINES)
```
geky, line 38: One nit, thoughts on naming this LFS_DEFINES? That would be a bit more consistent with LFS_CONFIG (or LFS_USER_DEFINES/LFS_USER_CONFIG)

yamt: i preferred LFS_USER_DEFINES/LFS_USER_CONFIG but i didn't want to make a breaking change just for aesthetic reasons.

if you think the consistency is important here, i can rename it to LFS_DEFINES.

geky: Yes, I think we should prefer consistency whenever possible. The fewer inconsistencies the fewer things for users to remember.

Another minor point in favor of LFS_DEFINES is that it's less characters to type on the command line, which may not be uncommon for this flag.

I can also update the PR if it's a hassle, GitHub just makes the process a bit awkward. I always feel like I'm walking into someone else's house and repainting their walls.

yamt: renamed.

yamt:
the ci failure seems unrelated to this PR.

geky (approved):

marcosgabrielserrano:
👀

BenBE:
Doesn't this effectively make `LFS_CONFIG` and `LFS_DEFINES` be the same (except for the latter also defining the default utils too)?

A better implementation would be to have `LFS_CONFIG` work just like now, but allow the user config to (un)define `LFS_SKIP_UTILS` to emulate both behaviors. `lfs_utils.h` on the other hand should be made aware that all of the macros it defines, may have been provided by the custom user header already.

geky:
I think it's a bit too late to be bikeshedding this. The only reason this isn't already merged is delays on my end in making a minor release (sorry! in hindsight the upcoming minor release should have been split into multiple releases).

Though discussion can influence future API changes.

---

> A better implementation would be to have `LFS_CONFIG` work just like now, but allow the user config to (un)define `LFS_SKIP_UTILS` to emulate both behaviors.

This is an interesting and clever workaround, but thinking from the perspective of a new user, you're asking them to do 2 things (define LFS_CONFIG and (un)define LFS_SKIP_UTILS) instead of 1 (define LFS_DEFINES). I suspect LFS_DEFINES will be more common for users who aren't doing a full OS integration, so having it require more steps isn't the greatest.

I'm not against dropping LFS_CONFIG on a later major API change if LFS_DEFINES (renamed -> LFS_CONFIG?) can fully replace it.

> `lfs_utils.h` on the other hand should be made aware that all of the macros it defines, may have been provided by the custom user header already.

This gets a bit tricky for things like the type definitions. Though maybe we require all user-provided types to also be macros?

``` c
#ifdef LFS_SIZE_T
typedef LFS_SIZE_T lfs_size_t
#else
#include <stdint.h>
typedef uint32_t lfs_size_t
#endif
```

geky:
Merging, thanks again for the PR! And sorry about the slow process.

yamt:
FYI, here is how i used this. https://github.com/apache/nuttx/pull/15150

PR #1013, opened 2024-08-01
Write the detect cycles function as a function to optimize code

wdfk-prog:
- Using the master branch code, the footprint size of the gcc compilation under Windows is as follows
```log
   text    data     bss     dec     hex filename
  33440       0       0   33440    82a0 lfs.o
    208       0       0     208      d0 lfs_util.o
  33648       0       0   33648    8370 (TOTALS)
```

- After this change, the compilation footprint is as follows
```log
   text    data     bss     dec     hex filename
  33228       0       0   33228    81cc lfs.o
    208       0       0     208      d0 lfs_util.o
  33436       0       0   33436    829c (TOTALS)
```

- The optimization result can be obtained,Compiles without error or warning.
```log
  0x8370- 0x829c = 0xD4
  0xD4/0x8370 = 0.63%
```

- gcc version
```log
Using built-in specs.
COLLECT_GCC=C:\MinGW\bin\gcc.exe
COLLECT_LTO_WRAPPER=c:/mingw/bin/../libexec/gcc/mingw32/9.2.0/lto-wrapper.exe
Target: mingw32
Configured with: ../src/gcc-9.2.0/configure --build=x86_64-pc-linux-gnu --host=mingw32 --target=mingw32 --disable-win32-registry --with-arch=i586 --with-tune=generic --enable-static --enable-shared --enable-threads --enable-languages=c,c++,objc,obj-c++,fortran,ada --with-dwarf2 --disable-sjlj-exceptions --enable-version-specific-runtime-libs --enable-libgomp --disable-libvtv --with-libiconv-prefix=/mingw --with-libintl-prefix=/mingw --enable-libstdcxx-debug --disable-build-format-warnings --prefix=/mingw --with-gmp=/mingw --with-mpfr=/mingw --with-mpc=/mingw --with-isl=/mingw --enable-nls --with-pkgversion='MinGW.org GCC Build-2'
Thread model: win32
gcc version 9.2.0 (MinGW.org GCC Build-2)
```

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -4396,6 +4396,30 @@ cleanup:
 }
 #endif
 
+struct lfs_tortoise_t {
+    lfs_block_t pair[2];
+    lfs_size_t i;
+    lfs_size_t period;
+};
+
+static int lfs_tortoise_detectcycles(
+    const lfs_mdir_t *dir, struct lfs_tortoise_t *tortoise) {
+    // detect cycles with Brent's algorithm
+    if (lfs_pair_issync(dir->tail, tortoise->pair)) {
+        LFS_WARN("Cycle detected in tail list");
+        return LFS_ERR_CORRUPT;
+    }
+    if (tortoise->i == tortoise->period) {
+        tortoise->pair[0] = dir->tail[0];
+        tortoise->pair[1] = dir->tail[1];
+        tortoise->i = 0;
+        tortoise->period *= 2;
+    }
+    tortoise->i += 1;
+
+    return LFS_ERR_OK;
+}
+
 static int lfs_mount_(lfs_t *lfs, const struct lfs_config *cfg) {
     int err = lfs_init(lfs, cfg);
     if (err) {
@@ -4404,23 +4428,16 @@ static int lfs_mount_(lfs_t *lfs, const struct lfs_config *cfg) {
 
     // scan directory blocks for superblock and any global updates
     lfs_mdir_t dir = {.tail = {0, 1}};
-    lfs_block_t tortoise[2] = {LFS_BLOCK_NULL, LFS_BLOCK_NULL};
-    lfs_size_t tortoise_i = 1;
-    lfs_size_t tortoise_period = 1;
+    struct lfs_tortoise_t tortoise = {
+        .pair = {LFS_BLOCK_NULL, LFS_BLOCK_NULL},
+        .i = 1,
+        .period = 1,
+    };
     while (!lfs_pair_isnull(dir.tail)) {
-        // detect cycles with Brent's algorithm
-        if (lfs_pair_issync(dir.tail, tortoise)) {
-            LFS_WARN("Cycle detected in tail list");
-            err = LFS_ERR_CORRUPT;
+        err = lfs_tortoise_detectcycles(&dir, &tortoise);
+        if (err < 0) {
             goto cleanup;
         }
-        if (tortoise_i == tortoise_period) {
-            tortoise[0] = dir.tail[0];
-            tortoise[1] = dir.tail[1];
-            tortoise_i = 0;
-            tortoise_period *= 2;
-        }
-        tortoise_i += 1;
 
         // fetch next block in tail list
         lfs_stag_t tag = lfs_dir_fetchmatch(lfs, &dir, dir.tail,
@@ -4633,22 +4650,17 @@ int lfs_fs_traverse_(lfs_t *lfs,
     }
 #endif
 
-    lfs_block_t tortoise[2] = {LFS_BLOCK_NULL, LFS_BLOCK_NULL};
-    lfs_size_t tortoise_i = 1;
-    lfs_size_t tortoise_period = 1;
+    struct lfs_tortoise_t tortoise = {
+        .pair = {LFS_BLOCK_NULL, LFS_BLOCK_NULL},
+        .i = 1,
+        .period = 1,
+    };
+    int err = LFS_ERR_OK;
     while (!lfs_pair_isnull(dir.tail)) {
-        // detect cycles with Brent's algorithm
-        if (lfs_pair_issync(dir.tail, tortoise)) {
-            LFS_WARN("Cycle detected in tail list");
+        err = lfs_tortoise_detectcycles(&dir, &tortoise);
+        if (err < 0) {
             return LFS_ERR_CORRUPT;
         }
-        if (tortoise_i == tortoise_period) {
-            tortoise[0] = dir.tail[0];
-            tortoise[1] = dir.tail[1];
-            tortoise_i = 0;
-            tortoise_period *= 2;
-        }
-        tortoise_i += 1;
 
         for (int i = 0; i < 2; i++) {
             int err = cb(data, dir.tail[i]);
@@ -4727,22 +4739,17 @@ static int lfs_fs_pred(lfs_t *lfs,
     // iterate over all directory directory entries
     pdir->tail[0] = 0;
     pdir->tail[1] = 1;
-    lfs_block_t tortoise[2] = {LFS_BLOCK_NULL, LFS_BLOCK_NULL};
-    lfs_size_t tortoise_i = 1;
-    lfs_size_t tortoise_period = 1;
+    struct lfs_tortoise_t tortoise = {
+        .pair = {LFS_BLOCK_NULL, LFS_BLOCK_NULL},
+        .i = 1,
+        .period = 1,
+    };
+    int err = LFS_ERR_OK;
     while (!lfs_pair_isnull(pdir->tail)) {
-        // detect cycles with Brent's algorithm
-        if (lfs_pair_issync(pdir->tail, tortoise)) {
-            LFS_WARN("Cycle detected in tail list");
+        err = lfs_tortoise_detectcycles(pdir, &tortoise);
+        if (err < 0) {
             return LFS_ERR_CORRUPT;
         }
-        if (tortoise_i == tortoise_period) {
-            tortoise[0] = pdir->tail[0];
-            tortoise[1] = pdir->tail[1];
-            tortoise_i = 0;
-            tortoise_period *= 2;
-        }
-        tortoise_i += 1;
 
         if (lfs_pair_cmp(pdir->tail, pair) == 0) {
             return 0;
@@ -4792,22 +4799,17 @@ static lfs_stag_t lfs_fs_parent(lfs_t *lfs, const lfs_block_t pair[2],
     // use fetchmatch with callback to find pairs
     parent->tail[0] = 0;
     parent->tail[1] = 1;
-    lfs_block_t tortoise[2] = {LFS_BLOCK_NULL, LFS_BLOCK_NULL};
-    lfs_size_t tortoise_i = 1;
-    lfs_size_t tortoise_period = 1;
+    struct lfs_tortoise_t tortoise = {
+        .pair = {LFS_BLOCK_NULL, LFS_BLOCK_NULL},
+        .i = 1,
+        .period = 1,
+    };
+    int err = LFS_ERR_OK;
     while (!lfs_pair_isnull(parent->tail)) {
-        // detect cycles with Brent's algorithm
-        if (lfs_pair_issync(parent->tail, tortoise)) {
-            LFS_WARN("Cycle detected in tail list");
-            return LFS_ERR_CORRUPT;
+        err = lfs_tortoise_detectcycles(parent, &tortoise);
+        if (err < 0) {
+            return err;
         }
-        if (tortoise_i == tortoise_period) {
-            tortoise[0] = parent->tail[0];
-            tortoise[1] = parent->tail[1];
-            tortoise_i = 0;
-            tortoise_period *= 2;
-        }
-        tortoise_i += 1;
 
         lfs_stag_t tag = lfs_dir_fetchmatch(lfs, parent, parent->tail,
                 LFS_MKTAG(0x7ff, 0, 0x3ff),


wdfk-prog:
Also worth mentioning is the use of gcc and makefiles on Windows; Can only compile and clear, view the overall occupancy size and generate assembly files, other operations can not take effect, will give an error; Is there any plan to support this part on Windows.

wdfk-prog:
- CI errors are all Error: The operation was canceled.

BenBE (changes requested):
The minor style stuff aside, I think there's a slight problem with that algorithm as it changes behaviour by moving the initialization part into the loop. This might still find trivial cases of loops, but will hide loops, that need the iteration behaviour of advancing the tortoises. You can check this by noticing that several of the assignments of the function `lfs_detect_cycles` are now dead stores, when previously they would be re-used in the next loop iteration. Nonetheless, centralizing this check pattern may allow code size reduction.

To fix that bug, the following should be done:

1. Introduce a structure to hold the state of the algorithm
```c
typedef struct lfs_tortoise {
    lfs_block_t block[2]; // Previously tortoise
    lfs_size_t i; // Previously tortoise_i
    lfs_size_t period; // Previously tortoise_period
} lfs_tortoise_t;
```

2. Pass a `lfs_tortoise_t*` as the second argument to `lfs_detect_cycles`

3. On the call sites: Outside the loop initialize a `lfs_tortoise_t` like so:
```c
lfs_tortoise_t tortoise = {
    .block = {LFS_BLOCK_NULL, LFS_BLOCK_NULL},
    .i = 1,
    .period = 1,
};
```

4. Pass that variable to `lfs_detect_cycles` by reference.

The "cancelled" CI jobs were likely due to some loop in the filesystem under test not being detected after your change.

wdfk-prog:
> The minor style stuff aside, I think there's a slight problem with that algorithm as it changes behaviour by moving the initialization part into the loop. This might still find trivial cases of loops, but will hide loops, that need the iteration behaviour of advancing the tortoises. You can check this by noticing that several of the assignments of the function `lfs_detect_cycles` are now dead stores, when previously they would be re-used in the next loop iteration. Nonetheless, centralizing this check pattern may allow code size reduction.撇开次要风格的东西不谈，我认为该算法存在一个小问题，因为它通过将初始化部分移动到循环中来改变行为。这仍然可能找到循环的微不足道的情况，但会隐藏循环，这些循环需要推进的迭代行为。您可以通过注意到函数`lfs_detect_cycles`的几个赋值现在是死存储来检查这一点，而以前它们将在下一次循环迭代中重用。尽管如此，集中管理此检查模式可能会减小代码大小。
> 
> To fix that bug, the following should be done:要修复该错误，应执行以下操作：
> 
> 1. Introduce a structure to hold the state of the algorithm引入一个结构来保存算法的状态
> 
> ```c
> typedef struct lfs_tortoise {
>     lfs_block_t block[2]; // Previously tortoise
>     lfs_size_t i; // Previously tortoise_i
>     lfs_size_t period; // Previously tortoise_period
> } lfs_tortoise_t;
> ```
> 
> 2. Pass a `lfs_tortoise_t*` as the second argument to 将 `lfs_tortoise_t*` 作为第二个参数传递给`lfs_detect_cycles`
> 3. On the call sites: Outside the loop initialize a `lfs_tortoise_t` like so:在调用站点上：在循环外初始化一个`lfs_tortoise_t`，如下所示：
> 
> ```c
> lfs_tortoise_t tortoise = {
>     .block = {LFS_BLOCK_NULL, LFS_BLOCK_NULL},
>     .i = 1,
>     .period = 1,
> };
> ```
> 
> 4. Pass that variable to `lfs_detect_cycles` by reference.通过引用将该变量传递给 `lfs_detect_cycles`。
> 
> The "cancelled" CI jobs were likely due to some loop in the filesystem under test not being detected after your change.“已取消”的 CI 作业可能是由于在更改后未检测到所测试文件系统中的某些循环。



> The minor style stuff aside, I think there's a slight problem with that algorithm as it changes behaviour by moving the initialization part into the loop. This might still find trivial cases of loops, but will hide loops, that need the iteration behaviour of advancing the tortoises. You can check this by noticing that several of the assignments of the function `lfs_detect_cycles` are now dead stores, when previously they would be re-used in the next loop iteration. Nonetheless, centralizing this check pattern may allow code size reduction.
> 
> To fix that bug, the following should be done:
> 
> 1. Introduce a structure to hold the state of the algorithm
> 
> ```c
> typedef struct lfs_tortoise {
>     lfs_block_t block[2]; // Previously tortoise
>     lfs_size_t i; // Previously tortoise_i
>     lfs_size_t period; // Previously tortoise_period
> } lfs_tortoise_t;
> ```
> 
> 2. Pass a `lfs_tortoise_t*` as the second argument to `lfs_detect_cycles`
> 3. On the call sites: Outside the loop initialize a `lfs_tortoise_t` like so:
> 
> ```c
> lfs_tortoise_t tortoise = {
>     .block = {LFS_BLOCK_NULL, LFS_BLOCK_NULL},
>     .i = 1,
>     .period = 1,
> };
> ```
> 
> 4. Pass that variable to `lfs_detect_cycles` by reference.
> 
> The "cancelled" CI jobs were likely due to some loop in the filesystem under test not being detected after your change.

Thank you for reminding me that I made a stupid mistake

lfs.c
```diff
@@ -4404,23 +4430,16 @@ static int lfs_mount_(lfs_t *lfs, const struct lfs_config *cfg) {
...
     while (!lfs_pair_isnull(dir.tail)) {
-        // detect cycles with Brent's algorithm
-        if (lfs_pair_issync(dir.tail, tortoise)) {
-            LFS_WARN("Cycle detected in tail list");
-            err = LFS_ERR_CORRUPT;
+        struct lfs_tortoise_t tortoise = {
+            .pair = {LFS_BLOCK_NULL, LFS_BLOCK_NULL},
+            .i = 1,
+            .period = 1,
+        };
```
BenBE, line 4438: This needs to go outside the loop. Similar for the other 3 places.

wdfk-prog: o(╥﹏╥)o

lfs.c
```diff
@@ -4396,6 +4396,32 @@ static int lfs_format_(lfs_t *lfs, const struct lfs_config *cfg) {
...
 #endif
 
+struct lfs_tortoise_t
+{
+    lfs_block_t pair[2];
+    lfs_size_t i;
+    lfs_size_t period;
+};
+
+static int lfs_detect_cycles(lfs_mdir_t dir, struct lfs_tortoise_t *tortoise)
```
geky, line 4406: It looks like you're passing the `dir` by value, instead of by pointer. This creates a copy of the 8-word struct, which might be adding both stack cost and code cost.

Passing by pointer:
``` c
static int lfs_detect_cycles(const lfs_mdir_t *dir, struct lfs_tortoise_t *tortoise) {
```

geky:
> Also worth mentioning is the use of gcc and makefiles on Windows; Can only compile and clear, view the overall occupancy size and generate assembly files, other operations can not take effect, will give an error; Is there any plan to support this part on Windows.

Unfortunately Windows is a difficult platform to support as a development environment. It's just not worth supporting vs leveraging all of the useful tools that are available in Linux.

The best option would be use either a virtual machine or [WSL](https://learn.microsoft.com/en-us/windows/wsl/about) to run a Linux image on windows and develop inside that. I've heard good things about WSL but haven't used it myself.

lfs.c
wdfk-prog: I missed it, it really shouldn't be, ^_^.
I've resubmitted it.

geky:
Sorry, this failure was an issue on our side. GitHub made a breaking change to their APIs recently.

To fix the pr this needs to be rebased onto the devel branch: https://github.com/littlefs-project/littlefs/tree/devel, let me know if you want me to do that for you:

``` bash
$ git rebase origin/devel
```

wdfk-prog:
> Sorry, this failure was an issue on our side. GitHub made a breaking change to their APIs recently.很抱歉，此失败是我们这边的问题。GitHub 最近对其 API 进行了重大更改。
> 
> To fix the pr this needs to be rebased onto the devel branch: https://github.com/littlefs-project/littlefs/tree/devel, let me know if you want me to do that for you:要修复 pr，这需要重新基到 devel 分支上：https://github.com/littlefs-project/littlefs/tree/devel，如果您希望我为您执行此操作，请告诉我：
> 
> ```shell
> $ git rebase origin/devel
> ```

OK

geky:
Sorry again, I was having problems pushing to your branch and couldn't understand why. You set "Maintainers are allowed to edit this pull request." correctly but pushes are still being rejected for some reason.

And wow, I think this is a real mess up on GitHub's part. There is apparently a bug where cross-organization forks can't be edited by maintainers, and I think this might also be affecting forks-of-cross-org-forks:
https://github.com/orgs/community/discussions/5634

So unfortunately it looks like you either need to add me as a collaborator or do the rebase locally... Fortunately it's a simple rebase with no conflicts.

wdfk-prog:
> Sorry again, I was having problems pushing to your branch and couldn't understand why. You set "Maintainers are allowed to edit this pull request." correctly but pushes are still being rejected for some reason.
> 
> And wow, I think this is a real mess up on GitHub's part. There is apparently a bug where cross-organization forks can't be edited by maintainers, and I think this might also be affecting forks-of-cross-org-forks: https://github.com/orgs/community/discussions/5634
> 
> So unfortunately it looks like you either need to add me as a collaborator or do the rebase locally... Fortunately it's a simple rebase with no conflicts.

- I don't know if my operation is correct, you can check it.

geky:
Looks good here, thanks for that. I've manually updated the bot comment since it's still a bit broken on our end.

lfs.c
```diff
@@ -4396,6 +4396,32 @@ static int lfs_format_(lfs_t *lfs, const struct lfs_config *cfg) {
 }
 #endif
 
+struct lfs_tortoise_t
```
geky, line 4399: nit: Squiggly brackets `{` should be on the same line:
``` c
struct lfs_tortoise_t {
```

lfs.c
```diff
@@ -4396,6 +4396,32 @@ static int lfs_format_(lfs_t *lfs, const struct lfs_config *cfg) {
...
+{
+    lfs_block_t pair[2];
+    lfs_size_t i;
+    lfs_size_t period;
+};
+
+static int lfs_detect_cycles(lfs_mdir_t *dir, struct lfs_tortoise_t *tortoise)
```
geky, line 4406: nit: Squiggly brackets `{` should be on the same line.

We should also stay with 80 columns, so this will probably need to wrap:
``` c
static int lfs_detect_cycles(
        lfs_mdir_t *dir, struct lfs_tortoise_t *tortoise) {
```

geky, line 4406: Do we modify `dir` here? `dir` should be const unless we need to modify it:
``` c
const lfs_mdir_t *dir,
```

geky, line 4406: nit: This is an odd one, but the naming convention in littlefs is underscores only for "namespaces".

So one option: `lfs_tortoise_detectcycles`.

You can see other examples of this in the codebase.

lfs.c
```diff
@@ -4404,23 +4430,16 @@ static int lfs_mount_(lfs_t *lfs, const struct lfs_config *cfg) {
...
+        .i = 1,
+        .period = 1,
+    };
     while (!lfs_pair_isnull(dir.tail)) {
-        // detect cycles with Brent's algorithm
-        if (lfs_pair_issync(dir.tail, tortoise)) {
-            LFS_WARN("Cycle detected in tail list");
-            err = LFS_ERR_CORRUPT;
+        err = lfs_detect_cycles(&dir, &tortoise);
+        if (err == LFS_ERR_CORRUPT) {
```
geky, line 4440: We probably want:
``` c
if (err) {
    goto cleanup;
}
```

Otherwise we miss any other errors returned by `lfs_detect_cycles`. It's true `lfs_detect_cycles` doesn't return anything but LFS_ERR_CORRUPT, but these sort of assumptions can cause problems for future refactoring.

lfs.c
```diff
@@ -4792,22 +4798,15 @@ static lfs_stag_t lfs_fs_parent(lfs_t *lfs, const lfs_block_t pair[2],
...
+    struct lfs_tortoise_t tortoise = {
+        .pair = {LFS_BLOCK_NULL, LFS_BLOCK_NULL},
+        .i = 1,
+        .period = 1,
+    };
     while (!lfs_pair_isnull(parent->tail)) {
-        // detect cycles with Brent's algorithm
-        if (lfs_pair_issync(parent->tail, tortoise)) {
-            LFS_WARN("Cycle detected in tail list");
+        if (lfs_detect_cycles(parent, &tortoise) == LFS_ERR_CORRUPT) {
```
geky, line 4807: This should probably be:
``` c
int err = lfs_detect_cycles(parent, &tortoise);
if (err) {
    return err;
}
```

For the above reason, and for consistency with the codebase. Putting non-trivial functions inside if/while statements make them easy to miss when reading the code.

geky:
I left some review comments, mostly related to style/consistency with the rest of the codebase.

But this looks good to me and we can probably bring it in on the next minor release.

wdfk-prog:
> I left some review comments, mostly related to style/consistency with the rest of the codebase.
> 
> But this looks good to me and we can probably bring it in on the next minor release.

- I have made the corresponding modification, maybe it should be marked with some normative information in the pr content;
- You can refer to the link below:https://github.com/RT-Thread/rt-thread/tree/master/.github
- It is recommended to use a formatted script before submitting, such as the following link:https://github.com/mysterywolf/formatting

geky:
Thanks for making those changes. This looks good to me and will bring it in next release.

> maybe it should be marked with some normative information in the pr content;

Point taken, it is a waste of both your time and my time commenting on style inconsistencies.

> It is recommended to use a formatted script before submitting, such as the following link:https://github.com/mysterywolf/formatting

I had looked into [astyle](https://astyle.sourceforge.net/) before, but struggled to make it do what I wanted. It was very opinionated about unnecessary things.

But a custom python script is a good idea, maybe that's the way to go.

geky (approved):

geky:
Merging. Thanks again for the PR! And sorry about the rough process.

PR #1017, opened 2024-08-16
Add stronger checks for offset in seek

lucic71:
Previous checks were trying to validate that file->pos does not become negative. However, the method used for checking this contains possible undefined behaivor (UB) because of the signed integer overflow.

This commit adds stronger checks for offset calculation:
  * make sure that ((lfs_soff_t) file->pos + off) is never < 0. Instead of using signed addition to check that (which can possibly lead to UB), use signed comparison: off < 0 && (lfs_soff_t) file->pos < -off. A special check of off against INT32_MIN is added to make sure that -off does not get transformed into -INT32_MIN, which is as well UB.
  * make sure that unsigned overflow does not occur in file->pos + (lfs_off_t) off.

Thoughts:
  * the lseek manual mandates an EOVERFLOW when the new offset cannot be represened in the offset type. I wonder if we want to return that instead of INVAL when an unsigned overflow occurs.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -3661,6 +3661,26 @@ static lfs_ssize_t lfs_file_write_(lfs_t *lfs, lfs_file_t *file,
 }
 #endif
 
+static bool check_valid_off(lfs_off_t current_pos, lfs_soff_t off) {
+    /* Check if current_pos + off would be less than 0. We cannot use signed addition
+     * to check that because it would result in possible UB.
+     */
+    if (off == INT32_MIN) {
+        return false;
+    }
+
+    if (off < 0 && ((lfs_soff_t) current_pos) < -off) {
+        return false;
+    }
+
+    /* Unsigned overflow. */
+    if (off > 0 && (current_pos + ((lfs_off_t) off) < current_pos)) {
+        return false;
+    }
+
+    return true;
+}
+
 static lfs_soff_t lfs_file_seek_(lfs_t *lfs, lfs_file_t *file,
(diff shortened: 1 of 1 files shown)


geky:
Hi @lucic71, thanks for creating a PR, sorry about the late response.

These checks were originally implemented before [v2.9](https://github.com/littlefs-project/littlefs/releases/tag/v2.9.0), in an attempt to provide limited support for 32-bit file sizes. But this idea has since been dropped due to API issues, with littlefs now strictly using 31-bit file sizes.

Which makes me wonder, is it possible to simplify all of this by just doing unsigned arithmetic everywhere and checking that the result is in the 31-bit range?

We probably don't need to worry about ever needing full 32-bit support as it would make more sense to provide a 63-bit variant at that point.

geky:
> the lseek manual mandates an EOVERFLOW when the new offset cannot be represened in the offset type. I wonder if we want to return that instead of INVAL when an unsigned overflow occurs.

This is an interesting question. Playing around with the API on 64-bit Linux I seem to only be able to get EINVAL as a result.

Maybe EOVERFLOW is intended for when the result doesn't fit in an `off_t`, but _does_ fit in the filesystem's theoretical max file size? Will need to investigate.

geky:
Hi @lucic71, I went ahead and pushed up an alternative implementation relying on 31-bit file sizes here: https://github.com/littlefs-project/littlefs/pull/1027

Let me know if you see any issues with it, and thanks for the original fix.

PR #1026, opened 2024-09-18
Update github actions to the latest versions

yamt:


geky:
Looks like there's a merge-on-download option: https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md#multiple-uploads-to-the-same-named-artifact

Will experiment, I have another fork of the repo for this. There's a few delicate steps (release scripts) that rely on artifacts.

geky:
Pushed up a couple changes to fix things, tested here: https://github.com/geky/littlefs-test-repo/pull/10

Also tweaked a couple things while editing/testing this:
- Switched from [dawidd6/action-download-artifact](https://github.com/dawidd6/action-download-artifact) to the standard [download-artifact](https://github.com/actions/download-artifact) for cross-workflow downloads.

  dawidd6/action-download-artifact used to be the only way to accomplish this, but it looks like the standard action added support for cross-workflow downloads recently.
  
- Dropped minor/patch pinning of action version, so only pinning on major version.

  I'm not sure there's any benefit to pinning minor/patch versions. GitHub is rather aggressive at deprecating/breaking anything out-of-date, so pinning the minor/patch version is more a liability than anything.
  
  Learned this the hard way trying to pin the Ubuntu version.

Let me know if this looks good to you, and I can go ahead and merge this into the devel branch to unblock other PRs.

yamt:
> Thanks for this, looks like @geky-bot is still having a bit of issue since we were relying on the original artifact merging behavior.

oops, i was not aware of the merging behavior.

geky:
To be fair it is pretty unintuitive behavior, but it was useful. I wonder if we'll see more issues with hitting the artifact limit with GitHub's changes.

geky:
Merging

geky:
Whoops looks like a missed a line and ended up limiting statuses to x86 only.

Fixing here: https://github.com/littlefs-project/littlefs/pull/1053

I sure do love the GitHub Actions user experience /s

PR #1036, opened 2024-10-28
add library.json to support compilation as platformio library

rwedler:
This is a simple PR to add compilation support as a platformio library.  The only thing I needed to add was a custom option to avoid the usage of `Wshadow` when compiling littlefs.

To use this in a project, it can be referenced via a URL in the project's `platformio.ini` with the following text

```
lib_deps =
    https://github.com/littlefs-project/littlefs#{githash/tag}
```

### DIFF
--- a/library.json
+++ b/library.json
@@ -1,0 +1,17 @@
+{
+  "name": "littlefs",
+  "description": "A little fail-safe filesystem designed for microcontrollers",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/littlefs-project/littlefs"
+  },
+  "homepage": "https://github.com/littlefs-project/littlefs",
+  "license": "BSD-3-Clause",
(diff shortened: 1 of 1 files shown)


geky:
Hi @rwedler, thanks for creating a PR, I appreciate the thought to upstream this.

First a quick note, to fix the CI failures you would need to rebase/retarget onto the `devel` branch. GitHub made a breaking change to their APIs somewhat recently and the fix hasn't made its way to our master branch yet.

But w.r.t. this PR, I have a general rule of not adding files that are only necessary to make downstream systems work.

This may feel unfair since it's only one file, but it opens a can of worms and makes it harder to argue against downstream-helpful changes that are unreasonable.

It also makes it so build system changes downstream can backflow into littlefs development. Breaking changes are already bad enough going in only one direction...

Issue #1040, opened 2024-11-04
trailing '/' on file confuses stat but not open


rob-zeno:
Recently integrated littlefs support into my Zephyr-based application and was doing some crude tests.

I noticed that if I create a file:

       `/mountpoint/file`

And then I pass

       `/mountpoint/file/`

(note the trailing '/') to Zephyr's `fs_stat` (which calls `lfs_stat`), it will return information about the file.  That seems incorrect -- the trailing '/' ought to require that the item be a directory.

The `fs_open` (which calls `lfs_open`) call DOES fail -- correctly in my opinion.

geky:
Hi @rob-zeno, thanks for creating an issue.

This is a good find, and it does look like a problem on littlefs's side.

There is a similar issue in `lfs_mkdir`, though humorously reversed, where `lfs_mkdir` rejects trailing slashes (https://github.com/littlefs-project/littlefs/pull/679 and https://github.com/littlefs-project/littlefs/issues/428).

These both stem from a lack of testing around trailing slashes.

---

It would be good to fix this, though one hiccup is does a change to path behavior constitute a breaking API change. I think you can argue no, since the current behavior is inconsistent both with POSIX and, as you noted, with itself.

geky:
Hi @rob-zeno, I've gone ahead and put together a PR that should fix this: https://github.com/littlefs-project/littlefs/pull/1046

With any luck it should land in the next minor release.

rob-zeno:
Thanks!  I'm following along in 1046, and looking forward to the fixes!

BTW, I found some other issues with Zephyr and large flash parts which turned out to not be LittleFS related at all, although it was LittleFS failues that confused me.  For a very large flash part (64MB), unless I used JEDEC runtime discovery (which is NOT the default in Zephyr), it would only use 3-byte addressing, and that caused LittleFS file systems in the lower part of the flash part to get actually written in the upper part.  Very scary when I first saw it, but when I finally got to the bottom of it, it made sense, and after enabling the JEDEC runtime support (which allowed 4-byte addressing to work), LittleFS is behaving rock-solid in my testing so far.  There's probably something to take up with the Zephyr community -- if large flash parts won't work with 3-byte addressing, then they should probably fail (it's actually detectable at build-time and run-time).  But LittleFS isn't involved.

geky:
That's good to know. I don't really have any connections to Zephyr, but it's useful if someone else has a similar problem.

[...]

Issue #729, opened 2022-09-22
Documenting file name limitations


Xywzel:
Can't find anywhere in lfs.h, but do the filenames have any limitations other than their size? I'm guessing that because many functions take paths that the filename should not contain '/' and with the names being null-terminated there likely should not be nulls inside them either, but are all other char values acceptable?

steverpalmer:
I've been looking for the same information.  I've not read every document in full, but I did find a reference in the SPEC.md file that says for "0x0xx LFS_TYPE_NAME" that "The data contains the file name stored as an ASCII string (may be expanded to UTF8 in the future)."  As @Xywzel suggests, I also guess that a filename cannot contain a '/' or a null, but what about other characters like '\', '*', control characters or whole filenames like ".", "..".  It would be create if this detail could be specified somewhere.

Xywzel:
If I'm reading the `lfs_dir_find`correctly in the lfs.c, then "." and ".." are indeed special cases. There also seems to be root directory for the file system, but it is hard to say if one would open that for reading with path "/" or with empty path.

geky:
I'll make a note to add documentation on this. littlefs follows Linux/Unix tradition with '/' and NULL being the only illegal characters, and '.' and '..' being the only reserved filenames. "/" is the proper way to reference the root directory. If there's anything I've missed littlefs will try to follow Linux/Unix.

The fact that the empty string is also accepted for the root directory wasn't really intended, checking with Linux it should probably be an LFS_ERR_NOENT. But at this point it shouldn't be changed unless there is a major revision.

Also UTF8 in paths is fine, that line of the documentation should be changed. UTF8's use of the high bit for continuation bytes means that it's not possible to introduce NULL bytes without a true NULL character.

This scheme has the unfortunate downside of requiring base64 or hex encoding to use raw data for filenames, but it's a tradeoff for directories and OS portability.

steverpalmer:
Thanks @geky,

I've been implementing a virtual file system to sit above littlefs, which is why I've been looking at the issue of paths fairly closely, and in particular the [POSIX specification](https://pubs.opengroup.org/onlinepubs/9699919799/).  (I understand that you are not trying to claim full compliance with POSIX, but those guys have had to think closely about some of these things, so I've found them a useful reference). When you consider the documentation, please could you also consider a few corner cases:

- I think I am right in saying that littlefs interprets sequences of "/" characters the same as a single "/" (e.g. line 1339 in `lfs_dir_find`);
[...]

Issue #428, opened 2020-05-15
lfs_mkdir() returns LFS_ERR_NOENT


PoppaChubby:
Hello,

my attempt to create a dir fails with LFS_ERR_NOENT... Of course there is no directory entry, or I would not been trying to create one !

```
/// Top level directory operations ///
int lfs_mkdir(lfs_t *lfs, const char *path) {
    LFS_TRACE("lfs_mkdir(%p, \"%s\")", (void*)lfs, path);
    // deorphan if we haven't yet, needed at most once after poweron
    int err = lfs_fs_forceconsistency(lfs);
    if (err) {
        LFS_TRACE("lfs_mkdir -> %d", err);
        return err;
    }

    struct lfs_mlist cwd;
    cwd.next = lfs->mlist;
    uint16_t id;
    err = lfs_dir_find(lfs, &cwd.m, &path, &id);
    if (!(err == LFS_ERR_NOENT && id != 0x3ff)) {
        LFS_TRACE("lfs_mkdir -> %d", (err < 0) ? err : LFS_ERR_EXIST);
        return (err < 0) ? err : LFS_ERR_EXIST;
    }
(...)
```
This test is failing in lfs_mkdir() because id ==0x3ff. I have no idea of the meaning of id, and why lfs_dir_find() puts  0x3ff in id. 
It seems that I missed something obvious... Do I need to call something before lfs_mkdir() ? (of course I mounted successfully the filsystem...)

Thanks for your help

PoppaChubby:
OK, problem solved, sorry for the noise.
My error was to try to create "myDir/". With "myDir" without the trailing "/", lfs_mkdir() returns 0.

- is there a doc somewhere explaining requirements on file and directory names ?
- what is the reason for returning LFS_ERR_NOENT when I try to create "myDir/" ? Does it try to create "" under "myDir" and complains because "myDir" does not exists ?

pavel-kirienko:
I observe the same behavior: `lfs_mkdir(..., "/a/")` fails with `LFS_ERR_NOENT`, but `lfs_mkdir(..., "/a")` succeeds. I'm not sure it's a defect, though. If you look at the standard, it says that ENOENT shall be returned if:

> A component of the path prefix specified by path does not name an existing directory or path is an empty string.

If you decompose `/a/`, the component `a` names a non-existent directory, so the function is expected to return ENOENT. Either that or I am misreading the standard.

inf265:
[...]

PR #1046, opened 2024-11-24
paths: Revisit path parsing, fix trailing slash behavior

geky:
This PR includes a number of tweaks to path parsing behavior to better align with POSIX. Better aligning with POSIX should make things less surprising for users, which is always a good thing.

This is the relevant section in POSIX: [4.16 Pathname Resolution](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap04.html#tag_04_16).

Though note there are still some differences, listed below.

In addition to these changes, this PR includes a rewrite of the [`tests/test_paths.toml`](https://github.com/littlefs-project/littlefs/blob/fix-trailing-slashes/tests/test_paths.toml) (+7269/-207 lines, +20 cases, +260 perms) to more aggressively cover path parsing corner cases. That being said, it's unreasonable to expect me to think of every possible corner case on the first (or nth?) try, so if you notice anything missing, please let me know.

I believe this also fixes all path-related issues reported so far (https://github.com/littlefs-project/littlefs/issues/1040, https://github.com/littlefs-project/littlefs/issues/729, https://github.com/littlefs-project/littlefs/issues/428), though it's likely I've overlooked some.

My current thinking is that most users are expecting POSIX behavior anyways, so it is reasonable to make these changes in a minor release, as long as there is sufficient documentation in the release notes. I haven't heard of any complaints related to the last set of [tiny breaking changes](https://xkcd.com/1172/) (in [v2.9](https://github.com/littlefs-project/littlefs/releases/tag/v2.9.0)) yet, though feedback is welcome.

---

### Path changes

- `lfs_mkdir` now accepts trailing slashes:

  ``` c
  before: lfs_mkdir("a/") => LFS_ERR_NOENT
  after:  lfs_mkdir("a/") => 0
  ```

- `lfs_stat`, `lfs_getattr`, etc, now reject trailing slashes if the file is not a directory:

  ``` c
  before: lfs_stat("reg_a/") => 0
  after:  lfs_stat("reg_a/") => LFS_ERR_NOTDIR
  ```

  Note trailing slashes _are_ accepted if the file is a directory:

  ``` c
  before: lfs_stat("dir_a/") => 0
  after:  lfs_stat("dir_a/") => 0
  ```

- `lfs_file_open` now returns LFS_ERR_NOTDIR if the path contains trailing slashes and the file is not a directory, or LFS_O_CREAT is specified and the file does not exist:

  ``` c
  before: lfs_file_open("reg_a/") => LFS_ERR_NOENT
  after:  lfs_file_open("reg_a/") => LFS_ERR_NOTDIR
  ```
  
  Note trailing slashes return LFS_ERR_ISDIR if the file is a directory:
  
  ``` c
  before: lfs_file_open("dir_a/") => LFS_ERR_ISDIR
  after:  lfs_file_open("dir_a/") => LFS_ERR_ISDIR
  ```
  
  Note LFS_ERR_NOENT takes priority if LFS_O_CREAT is not specified:
  
  ``` c
  before: lfs_file_open("missing_a/") => LFS_ERR_NOENT
  after:  lfs_file_open("missing_a/") => LFS_ERR_NOENT
  ```
  
- Attempting to navigate above the root directory now returns LFS_ERR_INVAL:

  ``` c
  before: lfs_stat("/../a") => 0
  after:  lfs_stat("/../a") => LFS_ERR_INVAL
  ```
  
  This actually moves away from the behavior of other filesystems, but towards, in my opinion, more reasonable behavior. It should also be more consistent with future theoretical `openat`-like APIs.
  
- The empty path ("") no longer aliases the root directory, now returns LFS_ERR_INVAL:

  ``` c
  before: lfs_stat("") => 0
  after:  lfs_stat("") => LFS_ERR_INVAL
  ```

---

### POSIX deviations

While this gets us closer to POSIX, there are still some subtle differences in behaviors. These shouldn't affect most users, but are worth documenting. Most of these are due to 1. littlefs not actually creating `.` and `..` entries on disk and 2. not using recursion during path resolution.

In my opinion these simply aren't worth trying to match POSIX exactly, but I'm curious if others disagree.

- Root modifications return EINVAL instead of EBUSY:

  ``` c
  littlefs: remove("/") => EINVAL
  POSIX:    remove("/") => EBUSY
  ```
  
  Reason: This would be the only use of EBUSY in the system.

- We accept modifications of directories with trailing dots:

  ``` c
  littlefs: remove("a/.") => 0
  POSIX:    remove("a/.") => EBUSY
  ```
  
  Reason: Not worth implementing.

- We do not check for existence of directories followed by dotdots:

  ``` c
  littlefs: stat("a/missing/..") => 0
  POSIX:    stat("a/missing/..") => ENOENT
  ```
  
  Reason: Difficult to implement non-recursively.

- We accept modifications of directories with trailing dotdots:

  ``` c
  littlefs: rename("a/b/..", "c") => 0
  POSIX:    rename("a/b/..", "c") => EBUSY
  ```
  
  Reason: Not worth implementing.

---

Related issues: https://github.com/littlefs-project/littlefs/issues/1040, https://github.com/littlefs-project/littlefs/issues/729, https://github.com/littlefs-project/littlefs/issues/428
Should supersede: https://github.com/littlefs-project/littlefs/pull/679
Related path issues found by: @rob-zeno, @XinStellaris, @PoppaChubby, @pavel-kirienko, @inf265, @Xywzel, @steverpalmer, and likely others
Feedback: Welcome

EDIT: Changed filenames in examples to hint the filetype.
EDIT: Tried to make the root ascension example a bit more clear.
EDIT: Changed `lfs_file_open("missing/")` to return LFS_ERR_NOTDIR
EDIT: Noted `lfs_file_open("dir_a/")` returns LFS_ERR_ISDIR
EDIT: Clarified that `lfs_file_open("missing/")` without LFS_O_CREAT returns LFS_ERR_NOENT

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -282,6 +282,21 @@ static int lfs_bd_erase(lfs_t *lfs, lfs_block_t block) {
 
 
 /// Small type-level utilities ///
+
+// some operations on paths
+static inline lfs_size_t lfs_path_namelen(const char *path) {
+    return strcspn(path, "/");
+}
+
+static inline bool lfs_path_islast(const char *path) {
+    lfs_size_t namelen = lfs_path_namelen(path);
+    return path[namelen + strspn(path + namelen, "/")] == '\0';
+}
+
+static inline bool lfs_path_isdir(const char *path) {
+    return path[lfs_path_namelen(path)] != '\0';
+}
+
 // operations on block pairs
 static inline void lfs_pair_swap(lfs_block_t pair[2]) {
     lfs_block_t t = pair[0];
@@ -1461,13 +1476,16 @@ static int lfs_dir_find_match(void *data,
     return LFS_CMP_EQ;
 }
 
+// lfs_dir_find tries to set path and id even if file is not found
+//
+// returns:
+// - 0                  if file is found
+// - LFS_ERR_NOENT      if file or parent is not found
+// - LFS_ERR_NOTDIR     if parent is not a dir
 static lfs_stag_t lfs_dir_find(lfs_t *lfs, lfs_mdir_t *dir,
         const char **path, uint16_t *id) {
     // we reduce path to a single name if we can find it
     const char *name = *path;
-    if (id) {
-        *id = 0x3ff;
-    }
 
     // default to root dir
     lfs_stag_t tag = LFS_MKTAG(LFS_TYPE_DIR, 0x3ff, 0);
@@ -1476,8 +1494,10 @@ static lfs_stag_t lfs_dir_find(lfs_t *lfs, lfs_mdir_t *dir,
 
     while (true) {
 nextname:
-        // skip slashes
-        name += strspn(name, "/");
+        // skip slashes if we're a directory
+        if (lfs_tag_type3(tag) == LFS_TYPE_DIR) {
+            name += strspn(name, "/");
+        }
         lfs_size_t namelen = strcspn(name, "/");
 
         // skip '.' and root '..'
@@ -1519,7 +1539,7 @@ nextname:
         // update what we've found so far
         *path = name;
 
-        // only continue if we hit a directory
+        // only continue if we're a directory
         if (lfs_tag_type3(tag) != LFS_TYPE_DIR) {
             return LFS_ERR_NOTDIR;
         }
@@ -1539,8 +1559,7 @@ nextname:
             tag = lfs_dir_fetchmatch(lfs, dir, dir->tail,
                     LFS_MKTAG(0x780, 0, 0),
                     LFS_MKTAG(LFS_TYPE_NAME, 0, namelen),
-                     // are we last name?
-                    (strchr(name, '/') == NULL) ? id : NULL,
+                    id,
                     lfs_dir_find_match, &(struct lfs_dir_find_match){
                         lfs, name, namelen});
             if (tag < 0) {
@@ -2603,12 +2622,12 @@ static int lfs_mkdir_(lfs_t *lfs, const char *path) {
     cwd.next = lfs->mlist;
     uint16_t id;
     err = lfs_dir_find(lfs, &cwd.m, &path, &id);
-    if (!(err == LFS_ERR_NOENT && id != 0x3ff)) {
+    if (!(err == LFS_ERR_NOENT && lfs_path_islast(path))) {
         return (err < 0) ? err : LFS_ERR_EXIST;
     }
 
     // check that name fits
-    lfs_size_t nlen = strlen(path);
+    lfs_size_t nlen = lfs_path_namelen(path);
     if (nlen > lfs->name_max) {
         return LFS_ERR_NAMETOOLONG;
     }
@@ -3057,7 +3076,7 @@ static int lfs_file_opencfg_(lfs_t *lfs, lfs_file_t *file,
 
     // allocate entry for file if it doesn't exist
     lfs_stag_t tag = lfs_dir_find(lfs, &file->m, &path, &file->id);
-    if (tag < 0 && !(tag == LFS_ERR_NOENT && file->id != 0x3ff)) {
+    if (tag < 0 && !(tag == LFS_ERR_NOENT && lfs_path_islast(path))) {
         err = tag;
         goto cleanup;
     }
@@ -3077,8 +3096,14 @@ static int lfs_file_opencfg_(lfs_t *lfs, lfs_file_t *file,
             goto cleanup;
         }
 
+        // don't allow trailing slashes
+        if (lfs_path_isdir(path)) {
+            err = LFS_ERR_ISDIR;
+            goto cleanup;
+        }
+
         // check that name fits
-        lfs_size_t nlen = strlen(path);
+        lfs_size_t nlen = lfs_path_namelen(path);
         if (nlen > lfs->name_max) {
             err = LFS_ERR_NAMETOOLONG;
             goto cleanup;
@@ -3842,6 +3867,12 @@ static int lfs_stat_(lfs_t *lfs, const char *path, struct lfs_info *info) {
         return (int)tag;
     }
 
+    // only allow trailing slashes on dirs
+    if (strchr(path, '/') != NULL
+            && lfs_tag_type3(tag) != LFS_TYPE_DIR) {
+        return LFS_ERR_NOTDIR;
+    }
+
     return lfs_dir_getinfo(lfs, &cwd, lfs_tag_id(tag), info);
 }
 
@@ -3944,7 +3975,7 @@ static int lfs_rename_(lfs_t *lfs, const char *oldpath, const char *newpath) {
     uint16_t newid;
     lfs_stag_t prevtag = lfs_dir_find(lfs, &newcwd, &newpath, &newid);
     if ((prevtag < 0 || lfs_tag_id(prevtag) == 0x3ff) &&
-            !(prevtag == LFS_ERR_NOENT && newid != 0x3ff)) {
+            !(prevtag == LFS_ERR_NOENT && lfs_path_islast(newpath))) {
         return (prevtag < 0) ? (int)prevtag : LFS_ERR_INVAL;
     }
 
@@ -3955,8 +3986,14 @@ static int lfs_rename_(lfs_t *lfs, const char *oldpath, const char *newpath) {
     struct lfs_mlist prevdir;
     prevdir.next = lfs->mlist;
     if (prevtag == LFS_ERR_NOENT) {
+        // if we're a file, don't allow trailing slashes
+        if (lfs_path_isdir(newpath)
+                && lfs_tag_type3(oldtag) != LFS_TYPE_DIR) {
+            return LFS_ERR_NOTDIR;
+        }
+
         // check that name fits
-        lfs_size_t nlen = strlen(newpath);
+        lfs_size_t nlen = lfs_path_namelen(newpath);
         if (nlen > lfs->name_max) {
             return LFS_ERR_NAMETOOLONG;
         }
@@ -4016,7 +4053,8 @@ static int lfs_rename_(lfs_t *lfs, const char *oldpath, const char *newpath) {
             {LFS_MKTAG_IF(prevtag != LFS_ERR_NOENT,
                 LFS_TYPE_DELETE, newid, 0), NULL},
             {LFS_MKTAG(LFS_TYPE_CREATE, newid, 0), NULL},
-            {LFS_MKTAG(lfs_tag_type3(oldtag), newid, strlen(newpath)), newpath},
+            {LFS_MKTAG(lfs_tag_type3(oldtag),
+                newid, lfs_path_namelen(newpath)), newpath},
             {LFS_MKTAG(LFS_FROM_MOVE, newid, lfs_tag_id(oldtag)), &oldcwd},
             {LFS_MKTAG_IF(samepair,
                 LFS_TYPE_DELETE, newoldid, 0), NULL}));
--- a/lfs.c
+++ b/lfs.c
@@ -3109,7 +3109,7 @@ static int lfs_file_opencfg_(lfs_t *lfs, lfs_file_t *file,
 
         // don't allow trailing slashes
         if (lfs_path_isdir(path)) {
-            err = LFS_ERR_ISDIR;
+            err = LFS_ERR_NOTDIR;
             goto cleanup;
         }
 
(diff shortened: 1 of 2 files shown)


bmcdonnell-fb:
Overall, this seems like good stuff, based on the description.

> * `lfs_file_open` now returns LFS_ERR_NOTDIR if the file exists but the path contains trailing slashes:
>   ```c
>   before: lfs_file_open("a/") => LFS_ERR_NOENT
>   after:  lfs_file_open("a/") => LFS_ERR_NOTDIR
>   ```

This one doesn't seem right to me. `lfs_file_open()` expects a non-dir file, right? I think maybe it should return `LFS_ERR_INVAL` in this case? Since the problem is that `"a/"` is (or would be) a dir, not a file.

Consider the following cases:

| FS state | Func call |
| --- | --- |
| `a` exists, is a dir | `lfs_file_open("a")` |
| `a` exists, is a dir | `lfs_file_open("a/")` |
| `a` exists, is a file | `lfs_file_open("a/")` |
| `a` does not exist | `lfs_file_open("a/")` |

I think maybe all of these should return the same error?

geky:
> This one doesn't seem right to me. `lfs_file_open()` expects a non-dir file, right? I think maybe it should return `LFS_ERR_INVAL` in this case? Since the problem is that `"a/"` is (or would be) a dir, not a file.


Yeah, I noticed this too and thought it was a bit funny:

| FS state | Func call | Return value |
| --- | --- | --- |
| `a` exists, is a dir | `lfs_file_open("dir_a")` | LFS_ERR_ISDIR |
| `a` exists, is a dir | `lfs_file_open("dir_a/")` | LFS_ERR_ISDIR |
| `a` exists, is a file | `lfs_file_open("reg_a/")` | LFS_ERR_NOTDIR |
| `a` does not exist | `lfs_file_open("missing_a/")` | LFS_ERR_ISDIR |

But it is consistent with `lfs_stat("reg_a/")`:

``` c
lfs_stat("reg_a/") => LFS_ERR_NOTDIR
```

If `lfs_stat("reg_a/")` returned LFS_ERR_ISDIR that would be very confusing...

---

Actually, I think `lfs_file_open("missing_a/") => LFS_ERR_ISDIR` is the mistake. This should probably be LFS_ERR_NOTDIR:

| FS state | Func call | Return value |
| --- | --- | --- |
| `a` exists, is a dir | `lfs_file_open("dir_a")` | LFS_ERR_ISDIR |
| `a` exists, is a dir | `lfs_file_open("dir_a/")` | LFS_ERR_ISDIR |
| `a` exists, is a file | `lfs_file_open("reg_a/")` | LFS_ERR_NOTDIR |
| `a` does not exist | `lfs_file_open("missing_a/")` | LFS_ERR_NOTDIR |

Then it would be consistent with the only other function that can "create" files, `lfs_rename`:

``` c
lfs_rename("reg_a", "missing_a/") => LFS_ERR_NOTDIR
```

For these weird corner cases I was mostly going off what my Linux + ext4 machine does, but reading the POSIX spec, I think this might be one of the cases where, as a retroactive spec, POSIX leaves this up to the implementation.

But I'll be honest, sometimes it's hard to know what exactly the spec is saying:

[open, openat — open file](https://pubs.opengroup.org/onlinepubs/9799919799/functions/open.html)

>[EISDIR]
The named file is a directory and oflag includes O_WRONLY or O_RDWR, or includes O_CREAT without O_DIRECTORY, or includes O_EXEC when O_EXEC is not the same value as O_SEARCH.
> 
> [ENOTDIR]
> A component of the path prefix names an existing file that is neither a directory nor a symbolic link to a directory; or O_CREAT and O_EXCL are not specified, the path argument contains at least one non-<slash> character and ends with one or more trailing <slash> characters, and the last pathname component names an existing file that is neither a directory nor a symbolic link to a directory; or O_DIRECTORY was specified and the path argument names a non-directory file; or the path argument names a non-directory file and O_SEARCH was specified when O_SEARCH is not the same value as O_EXEC.

Though POSIX `open` can also open directories, since it predates `opendir`, so things get a bit weird.

Will update.

bmcdonnell-fb:
> But it is consistent with `lfs_stat("reg_a/")`:
> 
> ```c
> lfs_stat("reg_a/") => LFS_ERR_NOTDIR
> ```

But `lfs_stat()` supports dirs, right? And `lfs_file_open()` doesn't, right? So these are not necessarily analogous.

> If `lfs_stat("reg_a/")` returned LFS_ERR_ISDIR that would be very confusing...

Fair.

> Actually, I think `lfs_file_open("missing_a/") => LFS_ERR_ISDIR` is the mistake. This should probably be LFS_ERR_NOTDIR:

| # | FS state	| Func call | You proposed return | I propose return |
| --- | --- | --- | --- | --- |
| 1 | `dir_a` exists, is a dir | `lfs_file_open("dir_a")` | `LFS_ERR_ISDIR` | `LFS_ERR_ISDIR` |
| 2 | `dir_a` exists, is a dir | `lfs_file_open("dir_a/")` | `LFS_ERR_ISDIR` | `LFS_ERR_ISDIR` |
| 3 | `reg_a` exists, is a file | `lfs_file_open("reg_a/")` | `LFS_ERR_NOTDIR` | `LFS_ERR_INVAL` |
| 4 | `missing_a` does not exist | `lfs_file_open("missing_a/")` | `LFS_ERR_NOTDIR` | However you indicate "file not found"... |

1. We agree. `ISDIR` accurately describes the problem.
2. Combining problems 1 & 3 - can argue for either, as both would need to be resolved, but you're unlikely to change a dir into a file, so I think 1 is better aligned
3. I'm on the fence.
   - On the one hand, I feel like we shouldn't use `LFS_ERR_NOTDIR` in any case here. The error message should indicate what the problem is. It's true that `"reg_a/"` is not a dir, but that's not why you can't `lfs_file_open()` it. You still couldn't open it if it were a dir (per 2).
   - OTOH, `LFS_ERR_INVAL` seems more correct in the sense of "true", but `LFS_ERR_NOTDIR` is more specific, though misleading. (Make it a dir, and you're back to 2.)
4. Combining problem 3 w/ the file not existing. What does `lfs_file_open("missing_a")` return? I think this should be the same.

geky:
> Combining problem 3 w/ the file not existing. What does lfs_file_open("missing_a") return? I think this should be the same.

This would be LFS_ERR_NOENT.

To make things more confusing, it depends on which flags you specify:

``` c
lfs_file_open("missing_a/", LFS_O_RDWR | LFS_O_CREAT) => LFS_ERR_NOTDIR
```

`lfs_file_open` actually already prioritizes LFS_ERR_NOENT,  though this combo is missing from the tests (TODO note to self add NOENT/EXCL+trailing slash/dot/dotdot tests):

``` c
lfs_file_open("missing_a/", LFS_O_RDWR) => LFS_ERR_NOENT
```

A more complete table:
| Func call | Flags | Proposed return value |
| --- | --- | --- |
| `lfs_file_open("dir_a")` | `LFS_O_RDWR` | LFS_ERR_ISDIR |
| `lfs_file_open("dir_a/")` | `LFS_O_RDWR` | LFS_ERR_ISDIR |
| `lfs_file_open("reg_a/")` | `LFS_O_RDWR` | LFS_ERR_NOTDIR |
| `lfs_file_open("missing_a/")` | `LFS_O_RDWR` | LFS_ERR_NOENT |
| `lfs_file_open("dir_a")` | `LFS_O_RDWR \| LFS_O_CREAT` | LFS_ERR_ISDIR |
| `lfs_file_open("dir_a/")` | `LFS_O_RDWR \| LFS_O_CREAT` | LFS_ERR_ISDIR |
| `lfs_file_open("reg_a/")` | `LFS_O_RDWR \| LFS_O_CREAT` | LFS_ERR_NOTDIR |
| `lfs_file_open("missing_a/")` | `LFS_O_RDWR \| LFS_O_CREAT` | LFS_ERR_NOTDIR |
| `lfs_file_open("dir_a")` | `LFS_O_RDWR \| LFS_O_CREAT \| LFS_O_EXCL` | LFS_ERR_EXIST |
| `lfs_file_open("dir_a/")` | `LFS_O_RDWR \| LFS_O_CREAT \| LFS_O_EXCL` | LFS_ERR_EXIST |
| `lfs_file_open("reg_a/")` | `LFS_O_RDWR \| LFS_O_CREAT \| LFS_O_EXCL` | LFS_ERR_NOTDIR |
| `lfs_file_open("missing_a/")` | `LFS_O_RDWR \| LFS_O_CREAT \| LFS_O_EXCL` | LFS_ERR_NOTDIR |

> 3. I'm on the fence.
>   - On the one hand, I feel like we shouldn't use `LFS_ERR_NOTDIR` in any case here. The error message should indicate what the problem is. It's true that `"reg_a/"` is not a dir, but that's not why you can't `lfs_file_open()` it. You still couldn't open it if it were a dir (per 2).
>   - OTOH, `LFS_ERR_INVAL` seems more correct in the sense of "true", but `LFS_ERR_NOTDIR` is more specific, though misleading. (Make it a dir, and you're back to 2.)

My understanding of POSIX/Linux/Unix is that EINVAL is generally left as a last resort, otherwise it's too easy to use as a blanket error when more info would be useful to the user.

> But `lfs_stat()` supports dirs, right? And `lfs_file_open()` doesn't, right? So these are not necessarily analogous.

That's true, though they do share the code path that decides if a trailing-slash is LFS_ERR_NOTDIR.

Thoughts on `lfs_rename("reg_a", "missing_a/") => LFS_ERR_NOTDIR`?

I think if we change `lfs_file_open("missing_a/")` we should change `lfs_rename("reg_a", "missing_a/")` to match. Though this would be another deviation from POSIX. POSIX is pretty explicit with rename ([rename, renameat — rename file](https://pubs.opengroup.org/onlinepubs/9799919799/functions/rename.html)).

---

I need to think about this a bit more, I'm not entirely opposed to `lfs_file_open("reg_a/") => LFS_ERR_ISDIR`, though I think it does get a bit confusing with no dirs are involved. I'm also not entirely opposed to `lfs_file_open("reg_a/") => LFS_ERR_INVAL`, but it does lose information and doesn't seem very helpful. Worst case a user thinks it's a different arg that's the problem.

Another way of looking at LFS_ERR_NOTDIR is that it's the file that you're attempting to create that is not a directory. Though I can understand where this could be confusing.

On the other hand, these are realistically going to be pretty rare errors for users to encounter unintentionally...

geky:
I've gone ahead and adopted LFS_ERR_NOTDIR for missing + trailing slashes, since it's at least an improvement over the open/rename mismatch.

Also updated tests, added the missing tests over NOENT /EXIST + trailing slashes, and updated the PR comment to match changes.

geky:
I think I'm going to leave `lfs_file_open("reg_a/")` as LFS_ERR_NOTDIR for now, if anything just because it matches POSIX rename. I agree with your assessment that LFS_ERR_NOTDIR is more specific than LFS_ERR_INVAL at the risk of being confusing, but we should only deviate from POSIX if there is a strong reason to.

POSIX has baggage and sometimes doesn't make the most sense, but it's what users are expecting.

If there are no other comments I'll bring this in and make a v2.10 release soon.

BenBE:
From experiece I'd expect both `lfs_file_open('missing')` and `lfs_file_open('missing/')` to return the equivalent of `ENOENT` (AKA "file not found").

bmcdonnell-fb:
@BenBE,

> From experiece I'd expect both `lfs_file_open('missing')` and `lfs_file_open('missing/')` to return the equivalent of `ENOENT` (AKA "file not found").

Experience with what?

BenBE:
> @BenBE,
> 
> > From experiece I'd expect both `lfs_file_open('missing')` and `lfs_file_open('missing/')` to return the equivalent of `ENOENT` (AKA "file not found").
> 
> Experience with what?

With other filesystems/OS. At least this would be my expectation. First test if it exists, then if it's also what you expect it to be.

geky:
> From experiece I'd expect both lfs_file_open('missing') and lfs_file_open('missing/') to return the equivalent of ENOENT (AKA "file not found").

@BenBE To be clear LFS_ERR_NOENT does take priority. `lfs_file_open` only returns LFS_ERR_NOTDIR if LFS_O_CREAT is specified, since LFS_ERR_NOENT would not make sense in this case.

See the big table in https://github.com/littlefs-project/littlefs/pull/1046#issuecomment-2496973365.

But clearly my explanation of the change is not worded well. Will update to hopefully make it more clear.

geky:
Should be updated now. I added an explicit note that `lfs_file_open("missing_a/")` returns LFS_ERR_NOENT if LFS_O_CREAT is not provided.

geky:
LGTM. Thanks all for the feedback. It's useful to know what people's thoughts are on these sort of changes.

PR #1051, opened 2024-12-08
Allow sync to be NULL when not needed

wangdongustc:
When using non-caching storage, the user can omit the `sync` function in `struct lfs_config`.

This would be more intuitive. 

(After spending hours debugging why the program stuck after `prog` is called... I went back to the doc and see the descriptions about `sync`. It says "simply return 0" which I thought to be "simply 0". Is it just me?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ If your storage caches writes, make sure that the provided `sync` function
 flushes all the data to memory and ensures that the next read fetches the data
 from memory, otherwise data integrity can not be guaranteed. If the `write`
 function does not perform caching, and therefore each `read` or `write` call
-hits the memory, the `sync` function can simply return 0.
+hits the memory, the `sync` function can be `NULL` or simply return 0.
 
 ## Design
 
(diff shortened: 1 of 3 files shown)


bmcdonnell-fb:
Alternatively, `LFS_ASSERT(lfs->cfg->sync != NULL)`.

wangdongustc:
> Alternatively, `LFS_ASSERT(lfs->cfg->sync != NULL)`.

Thanks. Done in https://github.com/littlefs-project/littlefs/pull/1052

geky:
Hi @wangdongustc, thanks for the PR.

My preference is #1052 (asserting).

Eventually the config system needs a rework to better support compile-time configuration. If LFS_CFG_SYNC can be defined at compile time, it shouldn't be too difficult to make the default behavior a noop with an `ifdef` and without a runtime check.

wangdongustc:
Revised in https://github.com/littlefs-project/littlefs/pull/1052

PR #1052, opened 2024-12-09
Assert on NULL IO functions

wangdongustc:
Assert that the IO functions (especially the `sync` function when using non-caching storage) are provided by the user.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -4217,6 +4217,14 @@ static int lfs_init(lfs_t *lfs, const struct lfs_config *cfg) {
     // which littlefs currently does not support
     LFS_ASSERT((bool)0x80000000);
 
+    // check that the required io functions are provided
+    LFS_ASSERT(lfs->cfg->read != NULL);
+#ifndef LFS_READONLY
+    LFS_ASSERT(lfs->cfg->prog != NULL);
+    LFS_ASSERT(lfs->cfg->erase != NULL);
+    LFS_ASSERT(lfs->cfg->sync != NULL);
+#endif
+
     // validate that the lfs-cfg sizes were initiated properly before
     // performing any arithmetic logics with them
     LFS_ASSERT(lfs->cfg->read_size != 0);
--- a/scripts/prettyasserts.py
+++ b/scripts/prettyasserts.py
@@ -86,6 +86,13 @@ def write_header(f, limit=LIMIT):
     f.writeln("}")
     f.writeln()
     f.writeln("__attribute__((unused))")
+    f.writeln("static void __pretty_assert_print_ptr(")
+    f.writeln("        const void *v, size_t size) {")
+    f.writeln("    (void)size;")
+    f.writeln("    printf(\"%p\", v);")
+    f.writeln("}")
+    f.writeln()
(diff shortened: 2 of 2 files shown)


wangdongustc:
emm...

> ```
> lfs.c:220:5: error: initialization of ‘intmax_t’ {aka ‘long int’} from ‘int (*)(const struct lfs_config *)’ makes integer from pointer without a cast [-Werror=int-conversion]
>   220 |     LFS_ASSERT(lfs->cfg->sync != NULL);
>       |     ^~~~~~~~~~~~~~~~~~~~~~
> lfs.c:220:5: note: (near initialization for ‘(anonymous)’)
> lfs.c:220:5: error: initialization of ‘intmax_t’ {aka ‘long int’} from ‘int (*)(const struct lfs_config *)’ makes integer from pointer without a cast [-Werror=int-conversion]
> lfs.c:220:5: note: (near initialization for ‘(anonymous)’)
> ```

`LFS_ASSERT(lfs->cfg->sync != NULL);` was expanded to `__PRETTY_ASSERT_INT_NE(lfs->cfg->sync , NULL);` by ./scripts/prettyasserts.py. Then it will raise those warnings..

Changing to `LFS_ASSERT((lfs->cfg->sync != NULL));` can fix the problem. But it looks a little bit weird to me...

geky:
Hi @wangdongustc, thanks for creating a PR.

I think this would be more useful in [`lfs_init`](https://github.com/littlefs-project/littlefs/blob/d01280e64934a09ba16cac60cf9d3a37e228bb66/lfs.c#L4156), which is where we have other config-related asserts. `lfs_init` is always called before any filesystem operations.

Are you able to also include asserts on read/prog/erase being non-null? These should all probably follow the same rules.

But only if LFS_READONLY is not defined:

``` c
LFS_ASSERT(lfs->cfg->read != NULL);
#ifndef LFS_READONLY
LFS_ASSERT(lfs->cfg->prog != NULL);
LFS_ASSERT(lfs->cfg->erase != NULL);
LFS_ASSERT(lfs->cfg->sync != NULL);
#endif
```

This gets quite a bit more nuanced than it first appears...

---

And clever solution for the prettyasserts.py issue, thanks for that.

I had a fix for pointer comparisons on an unrelated branch (https://github.com/littlefs-project/littlefs/commit/b5e264bec400b17b0741f574d1d5bf63d038a273), but I like your solution more.

(We will probably want to end up with both eventually to allow pointer comparisons without a literal NULL)

geky:
> But then I thought, why bother asserting the "read" function? Nobody would expect that lfs will work without a "read" function.

Now I'm wondering what it would take to support write-only memory 😬 

Thanks for this! Merging!

geky (approved):

bmcdonnell-fb:
> write-only memory 😬

WOM. 😆

geky:
https://web.archive.org/web/20160313132244/https://dl.dropboxusercontent.com/u/85027770/25120SpecTwoSides%20NC.pdf 😄

Apparently someone wrote a kernel driver for it: https://github.com/spacerace/write-only-memory

PR #1053, opened 2024-12-09
gha: Fix x86-only statuses

geky:
Looks like I missed a line during refactoring, resulted in only x86 sizes being reported in GitHub statuses.

\< Insert rant about GHA making a mess of commit histories \>

Related: https://github.com/littlefs-project/littlefs/pull/1026


geky:
Tested in https://github.com/geky/littlefs-test-repo/pull/11

PR #1056, opened 2024-12-18
Add LFS_F_WRUNCHECKED option to avoid validation and hence reading blocks when writing

andyp1per:
In ArduPilot we predominantly stream logs to disk while flying. In our current block logger we do this without validation and this PR adds an option to skip validation while writing to streamline the writing process.

Also fixes a build issue with the file backend when compiling without debug.

### DIFF
--- a/bd/lfs_filebd.c
+++ b/bd/lfs_filebd.c
@@ -141,6 +141,7 @@ int lfs_filebd_erase(const struct lfs_config *cfg, lfs_block_t block) {
 
     // erase is a noop
     (void)block;
+    (void)bd;
 
     LFS_FILEBD_TRACE("lfs_filebd_erase -> %d", 0);
     return 0;
--- a/lfs.c
+++ b/lfs.c
@@ -25,7 +25,6 @@ enum {
     LFS_CMP_GT = 2,
 };
 
-
 /// Caching block device operations ///
 
 static inline void lfs_cache_drop(lfs_t *lfs, lfs_cache_t *rcache) {
@@ -2891,7 +2890,7 @@ static int lfs_ctz_find(lfs_t *lfs,
 static int lfs_ctz_extend(lfs_t *lfs,
         lfs_cache_t *pcache, lfs_cache_t *rcache,
         lfs_block_t head, lfs_size_t size,
-        lfs_block_t *block, lfs_off_t *off) {
+        lfs_block_t *block, lfs_off_t *off, bool validate) {
     while (true) {
         // go ahead and grab a block
         lfs_block_t nblock;
@@ -2931,7 +2930,7 @@ static int lfs_ctz_extend(lfs_t *lfs,
                     }
 
                     err = lfs_bd_prog(lfs,
-                            pcache, rcache, true,
+                            pcache, rcache, validate,
                             nblock, i, &data, 1);
                     if (err) {
                         if (err == LFS_ERR_CORRUPT) {
@@ -2952,7 +2951,7 @@ static int lfs_ctz_extend(lfs_t *lfs,
             lfs_block_t nhead = head;
             for (lfs_off_t i = 0; i < skips; i++) {
                 nhead = lfs_tole32(nhead);
-                err = lfs_bd_prog(lfs, pcache, rcache, true,
+                err = lfs_bd_prog(lfs, pcache, rcache, validate,
                         nblock, 4*i, &nhead, 4);
                 nhead = lfs_fromle32(nhead);
                 if (err) {
@@ -3228,6 +3227,7 @@ static int lfs_file_close_(lfs_t *lfs, lfs_file_t *file) {
 
 #ifndef LFS_READONLY
 static int lfs_file_relocate(lfs_t *lfs, lfs_file_t *file) {
+
     while (true) {
         // just relocate what exists into new block
         lfs_block_t nblock;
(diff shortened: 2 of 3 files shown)


geky:
Hi @andyp1per, thanks for creating a PR.

Have you considered making this a filesystem-wide ifdef, say, LFS_NO_CKPROGS? That way you could also avoid the reads to validate metadata commits ([here](https://github.com/littlefs-project/littlefs/blob/630a0d87c23900c05ed5906db461ab91d3cc2c27/lfs.c#L1781-L1810)). I have a similar thing on a fork (albeit via mount flags).

I'm curious, did this impact performance noticeably? I've been viewing this as low priority as I assumed the read overhead for validation would always be overshadowed by the related prog+erase overhead.

andyp1per:
> Hi @andyp1per, thanks for creating a PR.
> 
> Have you considered making this a filesystem-wide ifdef, say, LFS_NO_CKPROGS? That way you could also avoid the reads to validate metadata commits ([here](https://github.com/littlefs-project/littlefs/blob/630a0d87c23900c05ed5906db461ab91d3cc2c27/lfs.c#L1781-L1810)). I have a similar thing on a fork (albeit via mount flags).
> 
> I'm curious, did this impact performance noticeably? I've been viewing this as low priority as I assumed the read overhead for validation would always be overshadowed by the related prog+erase overhead.

I am interested in other types of file being checked, that's why I went for a per-file option. It might also be possible to set it temporarily - so for instance our logs are self-describing, we probably want the description to be stored reliably but the data not so much.

read is about 10x faster than write and it also chews up the SPI bus, so not nothing - in my testing so far this plus avoiding checking the file size seems to get me where I need to, but I need to do another full flight test to fully check

geky:
Hi @andyp1per, just a quick update. I think I'm going to have to leave this open for a while due to how it interacts with future work.

In the ongoing experimental branch ([rbyd](https://github.com/littlefs-project/littlefs/tree/rbyd)), the separation of file data-structure and metadata data-structure becomes quite a bit more muddy, so I'm not sure it will be possible to set this flag on a per-file basis.

I'm also growing a bit concerned we're going to run out of 32-bit flags with the current planned features...

bmcdonnell-fb:
> In the ongoing experimental branch ([rbyd](https://github.com/littlefs-project/littlefs/tree/rbyd)),

BTW, what does "rbyd" mean?

geky:
> BTW, what does "rbyd" mean?

That's a good question! It's the name of the new metadata-data structure that started the branch. The original goal was to bring mdir compaction down from $O(n^2)$ to $O(n \log n)$, which is one of the main performance issues in littlefs, but it also serves as a backbone for file B-trees and a bunch of other things.

At this point it's sort of snowballed into a number of seemingly unrelated features.

I'm not particularly happy with how it's snowballed, but also not sure there's a better way to go about it. With filesystems/storage, things get "stuck" as soon as they're released, so new features need somewhere to "stew" while details/bugs get worked out...

---

The name itself comes from combining [Dhara trees](https://github.com/dlbeer/dhara) (which you may know about, it's a really neat data-structure), with a [red-black-yellow tree](https://en.wikipedia.org/wiki/Red%E2%80%93black_tree) to make it self-balancing and [order-statistic](https://en.wikipedia.org/wiki/Order_statistic_tree).

Maybe not the greatest name, but a name's a name.

There's a lot to write about it, but that's also a todo item...

geky:
> With filesystems/storage, things get "stuck" as soon as they're released

I should also mention I do plan to have a sort of "request-for-comments/feedback" period before stabilizing, but at the moment I'm still trying to get the details/kinks worked out.

geky:
One neat example, order-statistic operations make things like `lfs_file_fruncate` possible:

https://github.com/littlefs-project/littlefs/blob/1f3c47b1a66bf1483e611fb3445ddf8cc2cd6888/lfs.h#L1133-L1139

Which I think will be quite useful for maintaining logs without needing multiple files.

andyp1per:
> One neat example, order-statistic operations make things like `lfs_file_fruncate` possible:
> 
> https://github.com/littlefs-project/littlefs/blob/1f3c47b1a66bf1483e611fb3445ddf8cc2cd6888/lfs.h#L1133-L1139
> 
> Which I think will be quite useful for maintaining logs without needing multiple files.

Presumably that would make a posix compliant version of SEEK_SET possible.

geky:
> Presumably that would make a posix compliant version of SEEK_SET possible.

It does make random writes efficient, that is another big benefit.

Though I believe the current implementation of SEEK_SET is technically POSIX compliant, if only practical for random reads and random writes to small files.

andyp1per:
This is the bit that is not supported:
```
POSIX allows seeking beyond the existing end of file. If an output is performed after this seek, any read from the gap will return zero bytes. Where supported by the filesystem, this creates a sparse file. 
```

geky:
> This is the bit that is not supported:
>
> ```
> POSIX allows seeking beyond the existing end of file. If an output is performed after this seek, any read from the gap will return zero bytes. Where supported by the filesystem, this creates a sparse file.
> ```

Ah, I see. Yes, that's a good point, sparse files are a nice side-effect of B-trees.

I'm curious if what use cases you've seen for sparse files (SQLite?). I've been considering them low-priority since advanced file manipulation seems relatively rare in this space.

Sparse files are actually already implemented (it'd be a bit silly not to with B-trees), but more advanced APIs like SEEK_DATA, `lfs_file_punchhole`, `lfs_file_collapserange`, etc, are currently out-of-scope (but would be nice in the future).

PR #1064, opened 2025-01-13
scripts: Fixed several SyntaxWarning for python test helpers

tim-nordell-nimbelink:
Many of these require a r'' string context to avoid errors like:

  scripts/test.py:105: SyntaxWarning: invalid escape sequence '\s'


geky:
Ah! I did not realize Python's escape behavior was being deprecated. Seems this warning was introduced in [Python 3.12](https://docs.python.org/3/whatsnew/3.12.html#other-language-changes).

Thanks for this, will bring this in on the next release.

geky (approved):

geky:
Sorry about the delay, this should be merged shortly. Thanks for the PR!

PR #1065, opened 2025-01-15
Fix for "unsafe use of type bool" warning when compiling with MSVC.

amubiera:
Using negate on a bool causes a warning when compiling with MSVC 19.X:
```
littlefs\lfs.c(2561): error C2220: the following warning is treated as an error
littlefs\lfs.c(2561): warning C4804: '-': unsafe use of type 'bool' in operation
```

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -2558,7 +2558,7 @@ static int lfs_dir_orphaningcommit(lfs_t *lfs, lfs_mdir_t *dir,
         if (err != LFS_ERR_NOENT) {
             if (lfs_gstate_hasorphans(&lfs->gstate)) {
                 // next step, clean up orphans
-                err = lfs_fs_preporphans(lfs, -hasparent);
+                err = lfs_fs_preporphans(lfs, -(int8_t)hasparent);
                 if (err) {
                     return err;
                 }


lfs.c
```diff
@@ -2558,7 +2558,7 @@ static int lfs_dir_orphaningcommit(lfs_t *lfs, lfs_mdir_t *dir,
         if (err != LFS_ERR_NOENT) {
             if (lfs_gstate_hasorphans(&lfs->gstate)) {
                 // next step, clean up orphans
-                err = lfs_fs_preporphans(lfs, -hasparent);
```
geky, line 2561: Would it make more sense to change this to:

``` c
err = lfs_fs_preporphans(lfs, -(int8_t)hasparent);
```

?

lfs.c
```diff
@@ -2558,7 +2558,7 @@ static int lfs_dir_orphaningcommit(lfs_t *lfs, lfs_mdir_t *dir,
...
+                err = lfs_fs_preporphans(lfs, hasparent ? -1 : 0);
```
geky, line 2561: Would it make more sense to change this to:

``` c
err = lfs_fs_preporphans(lfs, -(int8_t)hasparent);
```

?

lfs.c
geky: Sorry, please ignore the duplicate. Getting used to GitHub's new issue UI.

lfs.c
amubiera: That works as well. Was just worried that on some compiler, a true bool could be all 0xFF and -0xFF would not turn out to be -1.

However, the C standard states that a `true` bool is always `1`:
```
4 Standard conversions
- An rvalue of type bool can be converted to an rvalue of type int, with
  false becoming zero and true becoming one.
```

I also checked Godbolt compiler explorer and both `-(int8_t)hasparent` and `hasparent ? -1 : 0` produce the same assembly in an optimized build. But slower in debug builds.

I pushed a new version with `-(int8_t)hasparent)`

Cheers!

geky: Thanks for this!

I was curious, for some reason the ternary operator cost a bit more. You would think GCC would be smart enough to emit the same code for these:

```
-hasparent              17104 B
hasparent ? -1 : 0      17112 B (+0.0%)
-(int8_t)hasparent      17104 B (+0.0%)
```

Not a deal-breaker, but very curious...

> That works as well. Was just worried that on some compiler, a true bool could be all 0xFF and -0xFF would not turn out to be -1.

This was a real concern in the wild-west before `stdbool.h`. Fortunately, littlefs only supports up to C99, which standardized `stdbool.h` and the `_Bool` type.

This has been enough of an issue though that we ended up adding an explicit assert:
https://github.com/littlefs-project/littlefs/blob/0494ce7169f06a734a7bd7585f49a9fa91fa7318/lfs.c#L4209-L4213

Related discussion: https://github.com/littlefs-project/littlefs/pull/772

geky:
Looks good here, thanks for creating a PR!

Will bring this in on the next patch release.

geky (approved):

geky:
Sorry about the delay, will be releasing this shortly. Thanks for the PR!

Issue #1061, opened 2024-12-31
Corruption reported upon concurrent directory modification


tpwrules:
I have an embedded application where one part of the application might be iterating a directory while another deletes files from it. The outcome of the iteration isn't meaningful to the application in this case, but sometimes `lfs_dir_read` returns `LFS_ERR_CORRUPT`. I would expect it to return some subset of the files in the directory (possibly none at all) but complete without any errors.

POSIX says that a directory iteration is unspecified to report an entry if it's added or removed while the iteration is in progress, but otherwise I believe it's expected to. I'm not sure LittleFS needs to go that far and guarantee entries that aren't touched will be returned (it would be nice if it did), but it at least should not claim corruption.

Is there any actual corruption occurring? Can file operations safely be continued as long as the directory is closed after the error is reported? Can LittleFS be easily fixed to not claim corruption?

To put it another way, I would not expect the below tests to fail, but they do on v2.10.1 at least. Iteration shouldn't give an error if a file is removed before reads start, nor if multiple files are removed between reads.

```
[cases.test_dirs_read_remove_multi]
defines.K = [0, 5]
if = 'N < BLOCK_COUNT/2'
code = '''
    lfs_t lfs;
    lfs_format(&lfs, cfg) => 0;
    lfs_mount(&lfs, cfg) => 0;
    lfs_mkdir(&lfs, "prickly-pear") => 0;
    for (int i = 0; i < 10; i++) {
        char path[1024];
        sprintf(path, "prickly-pear/cactus%03d", i);
        lfs_file_t file;
        lfs_file_open(&lfs, &file, path,
                LFS_O_WRONLY | LFS_O_CREAT | LFS_O_EXCL) => 0;
        lfs_file_close(&lfs, &file) => 0;
    }

    lfs_dir_t dir;
    struct lfs_info info;
    char path[1024];

    lfs_dir_open(&lfs, &dir, "prickly-pear") => 0;
    lfs_dir_read(&lfs, &dir, &info) => 1;
    assert(info.type == LFS_TYPE_DIR);
    assert(strcmp(info.name, ".") == 0);
    lfs_dir_read(&lfs, &dir, &info) => 1;
    assert(info.type == LFS_TYPE_DIR);
    assert(strcmp(info.name, "..") == 0);

    lfs_dir_read(&lfs, &dir, &info) => 1;
    sprintf(path, "prickly-pear/cactus%03d", (int)F);
    lfs_remove(&lfs, path) => 0;
[...]

PR #1068, opened 2025-02-04
Fix dir iteration being broken by concurrent removes

geky:
When removing a file, we mark all open handles as "removed" (pair={-1,-1}) to avoid trying to later read metadata that no longer exists. Unfortunately, this also includes open dir handles that happen to be pointing at the removed file, causing them to return LFS_ERR_CORRUPT on the next read.

The good news is this is _not_ actual filesystem corruption, only a logic error in `lfs_dir_read`.

We actually already have logic in place to nudge the dir to the next id, but it was unreachable with the existing logic. I suspect this worked at one point but was broken during a refactor due to lack of testing.

---

Fortunately, all we need to do is _not_ clobber the handle if the internal type is a dir. Then the dir-nudging logic can correctly take over.

I've also added `test_dirs_remove_read` to test this and prevent another regression, adapted from tests provided by @tpwrules that identified the original bug.

Found by @tpwrules
See https://github.com/littlefs-project/littlefs/issues/1061 for more info

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -2369,7 +2369,8 @@ fixmlist:;
             if (d->m.pair != pair) {
                 for (int i = 0; i < attrcount; i++) {
                     if (lfs_tag_type3(attrs[i].tag) == LFS_TYPE_DELETE &&
-                            d->id == lfs_tag_id(attrs[i].tag)) {
+                            d->id == lfs_tag_id(attrs[i].tag) &&
+                            d->type != LFS_TYPE_DIR) {
                         d->m.pair[0] = LFS_BLOCK_NULL;
                         d->m.pair[1] = LFS_BLOCK_NULL;
                     } else if (lfs_tag_type3(attrs[i].tag) == LFS_TYPE_DELETE &&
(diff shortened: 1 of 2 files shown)


tpwrules:
Thanks a bunch, I was able to test in our application with this patch on top of 2.10.1 and it seems to fix the bug we observed.

Do you think the case of removing multiple files between directory reads is adequately tested? My tests from the issue do pass after this patch too.

geky:
> Do you think the case of removing multiple files between directory reads is adequately tested? My tests from the issue do pass after this patch too.

I'm not opposed to accepting more tests :)

The problem is you can always have more testing, and my attention is elsewhere on a changeset that will eventually add more testing ([rbyd](https://github.com/littlefs-project/littlefs/blob/1f3c47b1a66bf1483e611fb3445ddf8cc2cd6888/tests/test_dread.toml)). So hardening the current state of things is unfortunately not a priority for me right now.

I'll respond with more in https://github.com/littlefs-project/littlefs/issues/1061#issuecomment-2658228809

geky:
@tpwrules Sorry about the delay on bringing this in. Unrelated PRs delayed the release a bit. Thanks for raising the issue on this!

PR #1070, opened 2025-02-05
Changed cast to correct type when trace is enabled for filebd

Noxet:
This does not compile when LFS_FILEBD_YES_TRACE is set, due to a cast to a `lfs_file_t` instead of `lfs_filebd_t`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -374,6 +374,29 @@ jobs:
         run: |
           CFLAGS="$CFLAGS -DLFS_NO_INTRINSICS" make test
 
+  # run with all trace options enabled to at least make sure these
+  # all compile
+  test-yes-trace:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


geky:
Hi @Noxet, thanks for this. That is a funny typo.

Will bring this in on the next patch release, though we should probably have something in CI to test this.

geky (approved):

geky:
I went ahead and added a CI job (test-yes-trace) to your PR to build/test the codebase with trace enabled. Let me know if you see any issues with it.

I was a bit worried this wouldn't catch filebd issues, since our test-runner doesn't actually use filebd, but apparently `make test` still compiles filebd and rambd, which I guess is a good thing?

Noxet:
Looks good!

geky:
Sorry about the delay, will be released shortly. Thanks for the PR!

PR #1078, opened 2025-02-28
Add a little bit of documentation on how to run tests.

BrianPugh:


README.md
```diff
@@ -199,6 +199,18 @@ The tests assume a Linux environment and can be started with make:
 make test
 ```
 
+Tests are defined as C code in toml files in the `tests/` directory.
+When developing a feature or fixing a bug, it is frequently useful to run a single suite of tests.
```
geky, line 203: nit: We should stick to 80-char widths where possible

README.md
```diff
@@ -199,6 +199,18 @@ The tests assume a Linux environment and can be started with make:
...
+To run individual tests or suite of tests:
+
+``` bash
+make test-runner
+python scripts/test.py --help  # See all available options
+python scripts/test.py --list-suites  # prints out all available test suites.
+python scripts/test.py --list-cases  # prints out all available individual test cases.
+python scripts/test.py runners/test_runner test_attrs  # runs the test_attrs suite.
```
geky, line 211: nit: Do we need `python` here? vs `./scripts/test.py`?

My understanding is you only need an explicit `python` in Windows, but there's no way these scripts happen to work in Windows.

geky (commented):
Hi @BrianPugh, thanks for this.

Happy to bring this in, just have a couple of nits.

geky:
Some other things that may be useful:

- `-j/--jobs` for parallel testing:

  ``` bash
  $ make test-runner -j && ./scripts/test.py runners/test_runner -j
  ```
  
  or just
  
  ``` bash
  $ make test -j
  ```

- `-O/--stdout` for capturing stdout:

  ``` bash
  $ ./scripts/test.py runners/test_runner -v -O-
  ```

- `-d/--disk` for capturing the resulting disk image:

  ``` bash
  $ ./scripts/test.py runners/test_runner -ddisk
  $ xxd -g1 -l64 disk
  ```

Though it's probably sufficient just to point to ./scripts/test.py --help.

Eventually the testing framework deserves its own wiki page.

BrianPugh:
hopefully d3109d8 addresses the feedback!

geky:
Sorry about jumping in there. Since this is the README.md and the first thing users see, I wanted to be a bit more pedantic than is probably reasonable for a code review loop.

Ended up adding some notes on test ids and dropping into gdb.

Let me know if this looks good to you.

geky (approved):

BrianPugh:
looks great! and totally understand! I like all the additional examples; it definitely makes adding new tests much more approachable!

geky:
Will be released shortly, thanks for the original proposal!

PR #1084, opened 2025-03-14
fix: avoid assuming struct packing

elupus:
lfs_gstate_t was assumed to be a packed array of uint32_t, but this is not always guaranteed. Access the fields directly instead of attempting to loop over an array of uint32_t

Fixes clang tidy warnings about use of uninitialized memory accessed.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -410,16 +410,9 @@ static inline void lfs_gstate_xor(lfs_gstate_t *a, const lfs_gstate_t *b) {
 }
 
 static inline bool lfs_gstate_iszero(const lfs_gstate_t *a) {
-    if (a->tag != 0) {
-        return false;
-    }
-    if (a->pair[0] != 0) {
-        return false;
-    }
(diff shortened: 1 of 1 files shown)


lfs.c
```diff
@@ -404,16 +404,20 @@ struct lfs_diskoff {
...
+    if (a->tag != 0) {
+        return false;
+    }
+    if (a->pair[0] != 0) {
+        return false;
+    }
+    if (a->pair[1] != 0) {
+        return false;
     }
     return true;
```
bmcdonnell-fb, line 422: Nit: this could potentially be condensed (if not simplified) to:

```
    return ((a->tag == 0) && (a->pair[0] == 0) && (a->pair[1] == 0));
```

geky (approved):

elupus:
Thanx!

geky:
It's not quite in master yet, but will be shortly if this CI job succeeds. Thanks for the PR!

PR #1090, opened 2025-04-07
drop a few unsupported CFLAGS for clang

yamt:


geky:
Hi @yamt, thanks for the PR, sorry about the late response.

I'm not so sure on this one. When they break, `shell` commands can be confusing in Makefiles (I'm not sure what compiler wouldn't support --version, but still).

It also breaks stack measurement, though there's no way around that...

Thoughts on `ifeq ($(findstring clang,$(CC)),clang)`? Or an explicit NO_STACK or CLANG env variable?

geky:
> Or an explicit NO_STACK

Actually, maybe NO_CALLGRAPH or NO_SU, NO_STACK could be interpreted weirdly (and NO_CI would be even worse!)

yamt:
> Thoughts on `ifeq ($(findstring clang,$(CC)),clang)`?

on macOS, clang has a few names including "gcc" and "cc".

```shell
spacetanuki% /usr/bin/gcc --version  
Apple clang version 17.0.0 (clang-1700.0.13.3)
Target: x86_64-apple-darwin24.4.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
spacetanuki% /usr/bin/cc --version 
Apple clang version 17.0.0 (clang-1700.0.13.3)
Target: x86_64-apple-darwin24.4.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
spacetanuki% /usr/bin/clang --version
Apple clang version 17.0.0 (clang-1700.0.13.3)
Target: x86_64-apple-darwin24.4.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
spacetanuki% 
```

geky:
> on macOS, clang has a few names including "gcc" and "cc".

Ah... I get it, but if you force one compiler to cosplay as another you should probably expect some weird errors...

I don't like it, but this does seem like the only way for `make test` to work out-of-the-box on Mac...

yamt:
> > on macOS, clang has a few names including "gcc" and "cc".
> 
> Ah... I get it, but if you force one compiler to cosplay as another you should probably expect some weird errors...

earlier versions of xcode had a real gcc.
i guess it was necessary for them to smooth the migration to clang.

> I don't like it, but this does seem like the only way for `make test` to work out-of-the-box on Mac...

i don't like it either.
but it seems like a common practice for traditional make users.
cmake uses even more complex approach for its CMAKE_C_COMPILER_ID.
(actually process some code to see predefined macros.)

geky:
> cmake uses even more complex approach for its CMAKE_C_COMPILER_ID.
> (actually process some code to see predefined macros.)

Oh no! Anything but cmake!

> i don't like it either.
> but it seems like a common practice for traditional make users.

After sleeping on it, I think you're right. This does seem like the only route for `make test` to work out of the box on Mac. And that is useful.

Will bring this in on the next patch release. Thanks for the PR!

geky (approved):

geky:
Bumping to needs-minor. Not sure if Makefile-only changes _need_ a minor release, but there will probably be a minor release soon.

geky:
I've also added the NO_GCC variable. This should still be implicitly set if --version contains "clang", but allows other, non-gcc, users to explicitly disable the GCC-specific flag.

Let me know if this tweak causes any problems on your end.

Makefile
```diff
@@ -18,6 +18,12 @@ VALGRIND ?= valgrind
 GDB		 ?= gdb
 PERF	 ?= perf
 
+# guess clang or gcc (clang sometimes masquerades as gcc because of
+# course it does)
+ifeq ($(shell $(CC) --version | grep clang),clang)
```
yamt, line 23: i don't think this works because the line contains more characters than just `clang`.
```shell
spacetanuki% cc --version | grep clang
Apple clang version 17.0.0 (clang-1700.0.13.3)
spacetanuki% 
```

geky: AH, I was trying to avoid the double-negative and thought this was make's findstring. Reverted to `ifneq`. Good catch!

And actually tested with `CC=clang` this time..

yamt:
> Let me know if anything else looks wrong.

looks fine to me. thank you.

geky:
Merging these, thanks for the PRs!

PR #1094, opened 2025-04-16
Add support for shrinking a filesystem

### LINKED ISSUE #1093: Ability to downsize a filesystem
Hi,

In our product, we have some devices that were initialized with a `block_count` that was larger than the real block count.
After formatting the flash, in an update we reduced the block count, this was fine because at this point the filesystem was empty.

But recently we wanted to push an update that bumps the littlefs library version past https://github.com/littlefs-project/littlefs/pull/584 (fail mount when block count changes), which now leads to a failure to mount.

https://github.com/littlefs-project/littlefs/pull/584 mentions the idea of adding a flag in the future that would allow overriding the introduced check.

Would it be possible to introduce this flag?

What would be the risk to just ignore this check for filesystems that were shrunk ? Is there any risk that the block_count in the superblock is used somewhere else?

Is it possible to make it so that the flag will  overwrite the block_count the next time the superblock is relocated.

Maybe it would make sense when that happens to add a check that none of the now-deleted blocks are in use.

sosthene-nitrokey:
This PR adds a new `lfs_fs_shrink`, which functions similarly to `lfs_fs_grow`, but supports reducing the block count.

This functions first checks that none of the removed block are in use. If it is the case, it will fail.

Closes #1093

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -5233,7 +5233,7 @@ static int lfs_fs_gc_(lfs_t *lfs) {
 #endif
 
 #ifndef LFS_READONLY
-#ifdef LFS_SHRINKIFCHEAP
+#ifdef LFS_SHRINKNONRELOCATING
 static int lfs_shrink_checkblock(void * data, lfs_block_t block) {
     lfs_size_t threshold = *((lfs_size_t *) data);
     if (block >= threshold) {
@@ -5251,11 +5251,11 @@ static int lfs_fs_grow_(lfs_t *lfs, lfs_size_t block_count) {
     }
 
     
-#ifndef LFS_SHRINKIFCHEAP
+#ifndef LFS_SHRINKNONRELOCATING
     // shrinking is not supported
     LFS_ASSERT(block_count >= lfs->block_count);
 #endif
-#ifdef LFS_SHRINKIFCHEAP
+#ifdef LFS_SHRINKNONRELOCATING
     lfs_block_t threshold = block_count;
     err = lfs_fs_traverse_(lfs, lfs_shrink_checkblock, &threshold, true);
     if (err) {
--- a/lfs.c
+++ b/lfs.c
@@ -5234,8 +5234,8 @@ static int lfs_fs_gc_(lfs_t *lfs) {
 
 #ifndef LFS_READONLY
 #ifdef LFS_SHRINKNONRELOCATING
-static int lfs_shrink_checkblock(void * data, lfs_block_t block) {
-    lfs_size_t threshold = *((lfs_size_t *) data);
+static int lfs_shrink_checkblock(void *data, lfs_block_t block) {
+    lfs_size_t threshold = *((lfs_size_t*)data);
     if (block >= threshold) {
         return LFS_ERR_NOTEMPTY;
     }
@@ -5256,10 +5256,11 @@ static int lfs_fs_grow_(lfs_t *lfs, lfs_size_t block_count) {
     LFS_ASSERT(block_count >= lfs->block_count);
 #endif
 #ifdef LFS_SHRINKNONRELOCATING
-    lfs_block_t threshold = block_count;
-    err = lfs_fs_traverse_(lfs, lfs_shrink_checkblock, &threshold, true);
-    if (err) {
-        return err;
+    if (block_count < lfs->block_count) {
+        err = lfs_fs_traverse_(lfs, lfs_shrink_checkblock, &block_count, true);
+        if (err) {
+            return err;
+        }
     }
 #endif
 
--- a/lfs.c
+++ b/lfs.c
@@ -5233,12 +5233,41 @@ static int lfs_fs_gc_(lfs_t *lfs) {
 #endif
 
 #ifndef LFS_READONLY
-static int lfs_fs_rewrite_block_count(lfs_t *lfs, lfs_size_t block_count) {
+#ifdef LFS_SHRINKIFCHEAP
+static int lfs_shrink_checkblock(void * data, lfs_block_t block) {
+    lfs_size_t threshold = *((lfs_size_t *) data);
+    if (block >= threshold) {
+        return LFS_ERR_NOTEMPTY;
+    }
+    return 0;
+}
+#endif
+
+static int lfs_fs_grow_(lfs_t *lfs, lfs_size_t block_count) {
+    int err;
+
+    if (block_count == lfs->block_count) {
+        return 0;
+    }
+
+    
+#ifndef LFS_SHRINKIFCHEAP
+    // shrinking is not supported
+    LFS_ASSERT(block_count >= lfs->block_count);
+#endif
+#ifdef LFS_SHRINKIFCHEAP
+    lfs_block_t threshold = block_count;
+    err = lfs_fs_traverse_(lfs, lfs_shrink_checkblock, &threshold, true);
+    if (err) {
+        return err;
+    }
+#endif
+
     lfs->block_count = block_count;
 
     // fetch the root
     lfs_mdir_t root;
-    int err = lfs_dir_fetch(lfs, &root, lfs->root);
+    err = lfs_dir_fetch(lfs, &root, lfs->root);
     if (err) {
         return err;
     }
@@ -5261,41 +5290,6 @@ static int lfs_fs_rewrite_block_count(lfs_t *lfs, lfs_size_t block_count) {
     if (err) {
         return err;
     }
-    return 0;
-}
-
-static int lfs_fs_grow_(lfs_t *lfs, lfs_size_t block_count) {
-    // shrinking is not supported
-    LFS_ASSERT(block_count >= lfs->block_count);
-
-    if (block_count > lfs->block_count) {
-        return lfs_fs_rewrite_block_count(lfs, block_count);
-    }
-
-    return 0;
-}
-
-static int lfs_shrink_check_block(void * data, lfs_block_t block) {
-    lfs_size_t threshold = *((lfs_size_t *) data);
-    if (block >= threshold) {
-        return LFS_ERR_NOTEMPTY;
-    }
-    return 0;
-}
-
-static int lfs_fs_shrink_(lfs_t *lfs, lfs_size_t block_count) {
-    if (block_count != lfs->block_count) {
-
-        lfs_block_t threshold = block_count;
-
-        int err = lfs_fs_traverse_(lfs, lfs_shrink_check_block, &threshold, true);
-        if (err) {
-            return err;
-        }
-
-        return lfs_fs_rewrite_block_count(lfs, block_count);
-    }
-
     return 0;
 }
 #endif
@@ -6509,22 +6503,6 @@ int lfs_fs_grow(lfs_t *lfs, lfs_size_t block_count) {
     err = lfs_fs_grow_(lfs, block_count);
 
     LFS_TRACE("lfs_fs_grow -> %d", err);
-    LFS_UNLOCK(lfs->cfg);
-    return err;
-}
-#endif
-
-#ifndef LFS_READONLY
-int lfs_fs_shrink(lfs_t *lfs, lfs_size_t block_count) {
-    int err = LFS_LOCK(lfs->cfg);
-    if (err) {
-        return err;
-    }
-    LFS_TRACE("lfs_fs_shrink(%p, %"PRIu32")", (void*)lfs, block_count);
-
-    err = lfs_fs_shrink_(lfs, block_count);
-
-    LFS_TRACE("lfs_fs_shrink -> %d", err);
     LFS_UNLOCK(lfs->cfg);
     return err;
 }
--- a/lfs.c
+++ b/lfs.c
@@ -5233,38 +5233,67 @@ static int lfs_fs_gc_(lfs_t *lfs) {
 #endif
 
 #ifndef LFS_READONLY
+static int lfs_fs_rewrite_block_count(lfs_t *lfs, lfs_size_t block_count) {
+    lfs->block_count = block_count;
+
+    // fetch the root
+    lfs_mdir_t root;
+    int err = lfs_dir_fetch(lfs, &root, lfs->root);
+    if (err) {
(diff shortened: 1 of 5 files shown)


geky:
Hi @sosthene-nitrokey, thanks for the PR.

I'm going to contradict past me a bit, but I think this would be better implemented as an extension to `lfs_fs_grow`.

I've come around on the idea of add `lfs_fs_shrink` to littlefs, since it is also valuable for allowing users to allocate fixed storage for application specific uses (XIP mainly, https://github.com/littlefs-project/littlefs/pull/692). That being said, it is still a hard problem.

But in obnoxiously POSIX fashion I think we want to avoid adding a new function and instead allow `lfs_fs_grow` to specify block_counts both larger _and_ smaller than the origin block_count. This would match the behavior of `lfs_file_truncate`. It's unintuitive but there is a bit of reason behind the madness as it allows you to specify the block_count without caring about the current state. (It may also lead to better code reuse as you may not need the `lfs_fs_rewrite_block_count` function.)

To avoid unnecessary code cost, shrinking behavior would be opt-in behind `ifdef LFS_SHRINK`. This opens the door for using a different define, maybe `ifdef LFS_SHRINKIFCHEAP` (open to better names), for the erroring behavior proposed here.

----

So, instead of adding `lfs_fs_shrink`, add `ifdef LFS_SHRINKIFCHEAP` which enables the shrink-if-not-in-use behavior in `lfs_fs_grow`. I think this should be a relatively easy change to the proposed PR. But let me know if you have concerns.

Also I appreciate the relevant test cases. Usually I have to beg for those :)

lfs.c
```diff
@@ -5233,38 +5233,67 @@ static int lfs_fs_gc_(lfs_t *lfs) {
...
-        superblock.block_count = lfs->block_count;
+static int lfs_fs_shrink_(lfs_t *lfs, lfs_size_t block_count) {
+    if (block_count != lfs->block_count) {
 
-        lfs_superblock_tole32(&superblock);
-        err = lfs_dir_commit(lfs, &root, LFS_MKATTRS(
-                {tag, &superblock}));
+        lfs_block_t threshold = block_count;
+
+        int err = lfs_fs_traverse_(lfs, lfs_shrink_check_block, &threshold, true);
```
geky, line 5291: nit: Please limit to 80 chars/columns (I should really add a CI job for this)

lfs.c
```diff
@@ -5233,38 +5233,67 @@ static int lfs_fs_gc_(lfs_t *lfs) {
 #endif
 
 #ifndef LFS_READONLY
+static int lfs_fs_rewrite_block_count(lfs_t *lfs, lfs_size_t block_count) {
```
geky, line 5236: nit: Naming is a bit weird, but useful, in littlefs. Underscores indicate namespaces/subsystems for the most part. So this should be `lfs_fs_rewriteblockcount`.

lfs.c
```diff
@@ -5233,38 +5233,67 @@ static int lfs_fs_gc_(lfs_t *lfs) {
...
-        // update the superblock
-        lfs_superblock_t superblock;
-        lfs_stag_t tag = lfs_dir_get(lfs, &root, LFS_MKTAG(0x7ff, 0x3ff, 0),
-                LFS_MKTAG(LFS_TYPE_INLINESTRUCT, 0, sizeof(superblock)),
-                &superblock);
-        if (tag < 0) {
-            return tag;
-        }
-        lfs_superblock_fromle32(&superblock);
+static int lfs_shrink_check_block(void * data, lfs_block_t block) {
```
geky, line 5278: nit: see above, namespace quirkyness: This should be `lfs_fs_shrink_checkblock`.

lfs.h
```diff
@@ -772,6 +772,17 @@ int lfs_fs_gc(lfs_t *lfs);
 int lfs_fs_grow(lfs_t *lfs, lfs_size_t block_count);
 #endif
 
+#ifndef LFS_READONLY
+// Shrinks the filesystem to a new size, updating the superblock with the new
+// block count.
+//
+// Note: This first checks that none of the blocks that are being removed are in use
```
geky, line 779: nit: 80 cols

lfs.c
```diff
@@ -5233,38 +5233,67 @@ static int lfs_fs_gc_(lfs_t *lfs) {
...
+    lfs_size_t threshold = *((lfs_size_t *) data);
```
geky, line 5279: nit: spacing: `*((lfs_size_t*)data);`

lfs.c
```diff
@@ -5233,38 +5233,67 @@ static int lfs_fs_gc_(lfs_t *lfs) {
...
+    if (block >= threshold) {
+        return LFS_ERR_NOTEMPTY;
```
geky, line 5281: Clever reuse of an error code! I like it 👍

geky (commented):
Added some nits to consider. But this is some pretty great code. It would just be nice to move under `ifdef LFS_SHRINKIFCHEAP` to fit into above future plans.

lfs.c
geky:
```
hi OverLength ctermbg=darkblue
match OverLength /\%81v./     
```
is a decent way to highlight >80 char lines if you're in vim.

lfs.c
```diff
@@ -5233,38 +5233,67 @@ static int lfs_fs_gc_(lfs_t *lfs) {
 #endif
 
 #ifndef LFS_READONLY
+static int lfs_fs_rewrite_block_count(lfs_t *lfs, lfs_size_t block_count) {
+    lfs->block_count = block_count;
```
geky, line 5237: I was worried about a tricky case where the superblock update could potentially relocate a superblock outside of the new filesystem bounds. But by setting `lfs->block_count` early we avoid this 👍

bmcdonnell-fb:
@geky

> ... I think this would be better implemented as an extension to `lfs_fs_grow`.

Maybe rename it to `lfs_fs_resize`?

You could `#define  lfs_fs_grow  lfs_fs_resize`, and deprecate `lfs_fs_grow`.

geky:
Ah, but the name abuse is part of the charm of POSIX, see `lfs_file_truncate` :)

Users might also think `lfs_fs_resize` and `lfs_fs_size` are somehow related (though I do want to rename `lfs_fs_size` -> `lfs_fs_usage` at some point, but it's low priority).

lfs.c
sosthene-nitrokey: Makes sense. This function will be removed anyway following shrink being part of `lfs_fs_grow`.

lfs.c
sosthene-nitrokey: Done

sosthene-nitrokey:
Thank you for the review!

I made a fixup commit that applies your suggestions. I'm just not sure how to make it so that `LFS_FS_SHRINKCHEAP` is defined in the relevant tests (the tests pass if I add it to `runners/test_runners.h` so it's only about where I should put the define).

Regarding merging this functionality into `lfs_fs_grow`, I initially made it into its own function because I think shrinking is a more "dangerous" operation and should be explicit (as any usage would require careful review, more so than grow). I don't care much about being "POSIX-like", but if that's what you prefer no issue from me.

bmcdonnell-fb:
> Regarding merging this functionality into `lfs_fs_grow`, I initially made it into its own function because I think shrinking is a more "dangerous" operation and should be explicit (as any usage would require careful review, more so than grow).

This makes sense to me.

@geky maybe consider prioritizing clarity over POSIX-likeness, in cases where POSIX-likeness is quirky, unexpected, or unclear? Just my 2¢; apologies if you've already decided otherwise for the project.

lfs.h
```diff
@@ -766,6 +766,10 @@ int lfs_fs_gc(lfs_t *lfs);
 // Grows the filesystem to a new size, updating the superblock with the new
 // block count.
 //
+// if LFS_SHRINKIFCHEAP is defined, this function will also accept
+// block_counts smaller than the current configuration, after checking
+// that none of the blocks that are being removed are in use.
+//
 // Note: This is irreversible.
```
geky, line 773: We may want to drop the "Note: This is irreversible" comment now :)

This is great documentation, but I'm going to ask if you can also add the sentence "Note this is unreliable due to littlefs's pseudorandom block allocation.".

I was going to comment this on your issue, but LFS_SHRINKIFCHEAP is unlikely to be generally useful/reliable for users. littlefs places its blocks pseudorandomly, so LFS_SHRINKIFCHEAP will likely fail, pseudorandomly, on most filesystem images. We probably want to avoid users thinking they will be able to shrink if the filesystem is small enough.

It sounds like your case only occurred because of the old ignore-block-count API? Hopefully that won't be a problem for future users.

That being said, this may change based on future allocator designs.

Oh and also `if LFS_SHRINKIFCHEAP` -> `If LFS_SHRINKIFCHEAP`.

geky:
> For the name `LFS_FS_SHRINKCHEAP` what about `LFS_FS_SHRINK_NONMOVING` ?

That is a better name, but what if I raise you `LFS_SHRINKIFTRIVIAL`?

geky:
> Regarding merging this functionality into `lfs_fs_grow`, I initially made it into its own function because I think shrinking is a more "dangerous" operation and should be explicit (as any usage would require careful review, more so than grow).

That's a fair argument. Though I'm not sure if "dangerous" is the right word. Shrinking would still use copy-on-write updates, so worst case `lfs_fs_grow` errors and you end up with an unmodified filesystem (ignoring current issues with error recovery).

However it is a much more expensive operation.

And there is also the potential benefit of compile-time gc with the separate `lfs_fs_shrink`, though not sure if this is useful with it already being an opt-in ifdef...

Hmm. May have to think on this. Sorry if this ends up going the other way.

The main reason for merging is to prefer fewer, more powerful (not more complicated!), functions to make it easier for users the keep the API in their head. The more fragmented your API becomes, the harder it is to know whether or not a relevant function exists as a user.

Though maybe the benefits of separating outweigh the benefits of minimizing the API surface here...

> @geky maybe consider prioritizing clarity over POSIX-likeness, in cases where POSIX-likeness is quirky, unexpected, or unclear? Just my 2¢; apologies if you've already decided otherwise for the project.

At the risk of getting into controversial opinion territory, my opinion is consistency > clarity when it comes to API design. A clear name doesn't exactly help you if you can't find the function, and usually the name won't be able to capture the full idiosyncrasies of the function anyways.

lfs.h
sosthene-nitrokey: Done

sosthene-nitrokey:
> > For the name `LFS_FS_SHRINKCHEAP` what about `LFS_FS_SHRINK_NONMOVING` ?
> 
> That is a better name, but what if I raise you `LFS_SHRINKIFTRIVIAL`?

I still prefer `SHRINKNONMOVING`. It's not that important to me.

geky:
Having slept on it, I still think this should be one function (`lfs_fs_grow`):

1. It simplifies maintaining a variable sized fixed region of storage after the filesystem (for XIP, etc).
2. It better shares code internally.
2. It minimizes the API surface.
3. It matches `lfs_fs_truncate`, for better or worse.

It's subjective, but it feels like the best option to me. Thanks for changing it @sosthene-nitrokey.

---

> I still prefer `SHRINKNONMOVING`. It's not that important to me.

We can go with `LFS_SHRINKNONMOVING`. Though thoughts on `LFS_SHRINKNONRELOCATING` or `LFS_SHRINKNORELOCATE`? "Relocate" is the term used elsewhere in the codebase specifically for block -> block migrations.

lfs.c
```diff
@@ -5233,40 +5233,63 @@ static int lfs_fs_gc_(lfs_t *lfs) {
...
+    }
+
+    
+#ifndef LFS_SHRINKIFCHEAP
     // shrinking is not supported
     LFS_ASSERT(block_count >= lfs->block_count);
+#endif
+#ifdef LFS_SHRINKIFCHEAP
+    lfs_block_t threshold = block_count;
+    err = lfs_fs_traverse_(lfs, lfs_shrink_checkblock, &threshold, true);
```
geky, line 5260: Ah, sorry I should have caught this earlier.

Even with LFS_SHRINKNONRELOCATING enabled, we shouldn't traverse if unless `block_count < lfs->block_count`. `lfs_fs_traverse_` involves a lot of io, and users probably expect `lfs_fs_grow` to be cheap if it is a simple grow operation.

Also I think this can just be `&block_count`? It doesn't look like threshold is used after this.

lfs.c
```diff
@@ -5233,40 +5233,63 @@ static int lfs_fs_gc_(lfs_t *lfs) {
 #endif
 
 #ifndef LFS_READONLY
+#ifdef LFS_SHRINKIFCHEAP
+static int lfs_shrink_checkblock(void * data, lfs_block_t block) {
```
geky, line 5237: nit: spacing: `void *data`

lfs.c
```diff
@@ -5233,40 +5233,63 @@ static int lfs_fs_gc_(lfs_t *lfs) {
...
+    lfs_size_t threshold = *((lfs_size_t *) data);
```
geky, line 5238: nit: spacing: `*((lfs_size_t*)data)`

tests/test_shrink.toml
```diff
@@ -0,0 +1,109 @@
+# simple shrink
+[cases.test_shrink_simple]
+defines.BLOCK_COUNT = [10, 15, 20]
+defines.AFTER_BLOCK_COUNT = [5, 10, 15, 19]
+   
+if = "AFTER_BLOCK_COUNT <= BLOCK_COUNT"
+code = '''
+#ifdef LFS_SHRINKIFCHEAP
```
geky, line 8: Ah, yes. This is the best way to `ifdef` right now. I just noticed the earlier test failures.

I have a branch that adds `ifdef` attrs to the test case itself, but it hasn't been upstreamed yet unfortunately...

geky:
Looks good to me, thanks for the tweaks. Will bring this into the next minor release.

geky (approved):

geky:
Merging this, thanks for the PR and updates based on feedback!

PR #1109, opened 2025-05-27
refactor: value stored to 'diff' is never read

Ryan-CW-Code:


Ryan-CW-Code:
I also used clang-tidy and cppcheck

PR #1110, opened 2025-05-28
perf: gc might try to populate the lookahead buffer each time

Ryan-CW-Code:
Sorry my English is very poor, so I use machine translation.
When the user incorrectly sets lookahead_size, if lookahead_size > (block_count / 8), gc might try to populate the lookahead buffer each time.
I'm not sure if this PR is the optimal solution. Perhaps it should help users set the correct value during the mount stage.
Because similar processing is also done in the lfs_format_ and lfs_alloc_scan functions, in fact, this processing can be completely optimized out.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -5225,7 +5225,9 @@ static int lfs_fs_gc_(lfs_t *lfs) {
     }
 
     // try to populate the lookahead buffer, unless it's already full
-    if (lfs->lookahead.size < lfs_min(8 * lfs->cfg->lookahead_size, lfs->block_count)) {
+    if (lfs->lookahead.size < lfs_min(
+            8 * lfs->cfg->lookahead_size,
+            lfs->block_count)) {
         err = lfs_alloc_scan(lfs);
         if (err) {
             return err;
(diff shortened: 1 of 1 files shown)


geky:
Interesting, it looks like you're right, this is needed to avoid redundant lookahead scans when `8*lookahead_size > block_count`. Can bring this in (next patch?) and thanks for the PR!

> I'm not sure if this PR is the optimal solution. Perhaps it should help users set the correct value during the mount stage.
Because similar processing is also done in the lfs_format_ and lfs_alloc_scan functions, in fact, this processing can be completely optimized out.

This gets tricky when `block_count` is not a multiple of 8. Unlikely, for large filesystems, but useful for carving out blocks for partition tables, reserved storage, XiP, etc.

But you're right this doesn't feel very optimal.

_Eventually_ (the "Config API rework" in https://github.com/littlefs-project/littlefs/pull/1111), I would like to drop the idea that `struct lfs_config` needs to be statically allocated. If instead we store everything in `lfs_t` in RAM, we can apply `lfs_min` during the mount stage and simplify this logic. By default this would add RAM, but in theory in the "Config API rework" that RAM could be saved by defining LFS_CFG_LOOKAHEAD_SIZE at compile time.

geky (approved):

Issue #934, opened 2024-01-31
Provide option to use string arguments that are not null-terminated


BSven:
I would highly appreciate to have the option to use explicit length specifications for string parameters in all API functions.

**Example for open():**
I would be great to not only have:
`int lfs_file_open(lfs_t *lfs, lfs_file_t *file,
        const char *path, int flags);`
... where path is assumed to be null-terminated, but also:
`int lfs_file_open(lfs_t *lfs, lfs_file_t *file,
        const char *path, lfs_size_t path_length, int flags);`
... where the length of path is specified explicitely.

geky:
Hi @BSven, this should be doable. We would just need to change the path parsing code ([`lfs_dir_find`](https://github.com/littlefs-project/littlefs/blob/f53a0cc961a8acac85f868b431d2f3e58e447ba3/lfs.c#L1461C19-L1461C31)) to not rely on null-terminated strings. Unfortunately the null-terminator _is_ useful for writing parsers. Curiously this often gets overlooked when discussing their tradeoffs.

One question is what would these functions be named? `lfs_file_opennotnullterminated`? `lfs_file_opennnt`?

BSven:
Hi @geky,
thanks for your immediate reply and your constructive outlook!

I would go with `lfs_file_open_nnt` so that the main function (i.e., open) stands out clearly.

geky:
> thanks for your immediate reply and your constructive outlook!

Well, to be clear this will be low-priority relative to other implementation-level issues, of which there is a bit of a long list.

> I would go with `lfs_file_open_nnt` so that the main function (i.e., open) stands out clearly.

I get where you're coming from, but right now littlefs's API only uses underscores to indicate the relevant struct namespace. Is this a weird holdout from old C? maybe. though I do think it helps delineate what the verb is. Either way it's a bit out of scope to change the naming convention of the whole API and ultimately consistency is more important.

I think the wildest this gets with all current proposals is `lfs_file_opennntatcfg`.

---

Other names could be `lfs_file_openz` or `lfs_file_openn`.

I'm sort of leaning towards `lfs_file_openn` to somewhat vaguely relate to `strncp` and friends. Though it is curious how little prior-art there is.

geky:
[...]

Issue #908, opened 2023-12-14
Unexpected data size information loss (zero)


sengulhamza:
Hello,

In our R&D project, we are using nRF9160 and mx25r64 external flash.
The project uses Zephyr as an OS. 
Nordic SDK 2.4.0
v2.4.0/modules/littlefs git commit hash _ca583fd297ceb48bced3c2548600dc615d67af24_
We are running LittleFS to use mx25r64.
We did our tests all day long, there was no problem, but we saw a problem in two of our devices.
The file system is mounting, we can list and see the files, but the file size returns zero as the value in the reading stage. It seems like the file is corrupted. The file is there, but the size is 0.
This problem was seen in our devices that have finished their software and are working in the field.

```
/**
 * The function checks if a file exists at the specified path and returns the appropriate status.
 *
 * @param path A string representing the file path to check for existence.
 *
 * @return an app_status_t enum value. It can return either APP_OK or APP_ERR, depending on whether the
 * file exists or not.
 */
app_status_t file_mngr_check_file_exist(const char *path, size_t *file_size)
{
    if ( !path ) {
        LOG_ERR("path is NULL to write");
        return APP_ERR;
    }
    char fpath[FILE_MNGR_MAX_PATH_LEN];
    snprintf(fpath, sizeof(fpath), "%s/%s", s_mp->mnt_point, path);
    struct fs_dirent entry;
    int rc = fs_stat(fpath, &entry);
    if (!rc) {
        if (entry.type == FS_DIR_ENTRY_FILE) {
            if (file_size) {
                *file_size = entry.size;
            }
            LOG_INF("%s is exist. Size is %d", path, entry.size);
        }
        return APP_OK; //File is exit.
    }
    LOG_ERR("%s is not exist.", path);
    return APP_ERR; //File is not exit.
}

```

**entry.size returns 0.**

When I delete and rewrite the file, or overwrite it, and read it again, there is no problem, but it contains important calibration data. I need to remove the device from the field.

What could be the cause, I am open to your suggestions.

Sincerely,

geky:
Hi @sengulhamza,

Is it possible you're losing power between creating a new file and the first call to `lfs_file_sync`/`lfs_file_close`?

When you create a new file with `lfs_file_open` it creates a zero-length file to put the filename somewhere. If you lose power before `lfs_file_sync`/`lfs_file_close` the zero-length file will still be there.

There some work planned to avoid this case (basically an "ignore me" file type), but it is not yet implemented.

[...]

Issue #862, opened 2023-07-31
Too inefficient way to append file content, when i need to save file immediately


muguou:
I use W25Q128 Flash.
My LFS adopts the following configuration.

const struct lfs_config fsCfg =
{
	.read_size = 32,
	.prog_size = 32,
	.block_size = 4096,
	.block_count =10,       //address start at 0x800
	.lookahead_size = 64,
	.block_cycles = 8,
       .cache_size = 1024,
};

I have some data that needs to be stored in flash at any time when there is a power outage. So every time I use lfs to open a file (use open type writeOnly, append) and write to it, I immediately close the file (I also used sync testing, and the results were the same).

If I don't close file it  or synchronize, then the power outage data will be lost at this time

The data is stored in a structure with a size of 64B, and the size is always 64 when writing. Because it immediately shuts down after each write, lfs saves the data back to flash after each write. In this case, when the file size is over 512 B, every time I save data, lfs erases a new sector. You know, I only wrote 64B of data, but every time lfs erases an entire sector of 4096B, there is a difference of 2 ^ 6=64 times. This is a bit too inefficient, and even less efficient than simply erasing the sectors in place. 

Is it caused by my usage error.

My log is as follows。
The
 "i" mean times of write.
 "size"  means the total number of bytes in the file now( befor write data and close file ).
"fe" means erase sector .

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>i= 0,size=0 (file size now)
w25WrAdd = 0x801040,size=64    //maybe creat file
w25WrAdd = 0x801080,size=96   //maybe write file =64; update file info = 32; total 96
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>i= 1,size=64 
w25WrAdd = 0x8010E0,size=160   // maybe write file =64*2; update file info = 32;total 160. 
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>i= 2,size=128 
w25WrAdd = 0x801180,size=224  // maybe write file =64*3; update file info = 32;total 224. Copying the entire file every time is too inefficient

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>i= 3,size=192 
w25WrAdd = 0x801260,size=288
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>i= 4,size=256 
[...]

PR #1111, opened 2025-05-29
v3-alpha: Rbyds and B-trees and gcksums, oh my!

geky:
```
                  ###    
               ####### #####    #####
             ################ ########  ####
       #### ## ############################## ####
     ############-#### ##_#########################
    #######'#########_\'|######_####################
     #### ##########_.. \ .'#_._#### #/#_##########
   ###############_    \ v .'_____#_.'.'_ ###########
  ###############_ "'. |  .""  _________ '"-##########
  #######"#####..--.  /    .-"'|  | |   '"-####'######
    ############    \     /  .---.|_|_    ###########
     ##########     |    /  -|        |-   ########
                    )    |  -|littlefs|-
                    |    |  -|   v3   |-
                    |    |   '--------'
                    |   ||     ' ' '
       ------------/-/---\\----------------------
```

## Table of contents <a href="#table-of-contents" id="table-of-contents">^</a>

<ol>
  <li><a href="#hello">Hello!</a></li>
  <li><a href="#disk-breaking">Wait, a disk breaking change?</a></li>
  <li>
    <details>
    <summary><a href="#whats-new">What's new?</a></summary>
    <ol>
      <li>
        <details>
        <summary><a href="#implemented">Implemented</a></summary>
        <ol>
          <li><a href="#n-meta-compaction">Efficient metadata compaction</a></li>
          <li><a href="#n-random-writes">Efficient random writes</a></li>
          <li><a href="#n-logging">Better logging: No more sync-padding issues</a></li>
          <li><a href="#n-inline-files">Efficient inline files, no more RAM constraints</a></li>
          <li><a href="#n-caches">Independent file caches</a></li>
          <li><a href="#n-fruncate">Easier logging APIs: lfs3_file_fruncate</a></li>
          <li><a href="#n-holes">Sparse files</a></li>
          <li><a href="#n-namelookups">Efficient file name lookup</a></li>
          <li><a href="#n-mtree">A simpler/more robust metadata tree</a></li>
          <li><a href="#n-sync-model">A well-defined sync model</a></li>
          <li><a href="#n-stickynotes">Stickynotes, no more 0-sized files</a></li>
          <li><a href="#n-compat">A new and improved compat flag system</a></li>
          <li><a href="#n-gcksums">Error detection! - Global-checksums</a></li>
          <li><a href="#n-traversals">Better traversal APIs</a></li>
          <li><a href="#n-gc">Incremental GC</a></li>
          <li><a href="#n-error-recovery">Better recovery from runtime errors</a></li>
          <li><a href="#n-sattrs">Standard custom attributes</a></li>
          <li><a href="#n-tests">More tests!</a></li>
        </ol>
        </details>
      </li>
      <li>
        <details>
        <summary><a href="#planned">Planned</a></summary>
        <ol>
          <li><a href="#n-bmap">Efficient block allocation, via optional on-disk block-map (bmap)</a></li>
          <li><a href="#n-bad-blocks">Bad block tracking</a></li>
          <li><a href="#n-preerased-blocks">Pre-erased block tracking</a></li>
          <li><a href="#n-meta-redund">Error correction! - Metadata redundancy</a></li>
          <li><a href="#n-data-redund">Error correction! - Data redundancy</a></li>
          <li><a href="#n-dedup">Transparent block deduplication</a></li>
        </ol>
        </details>
      </li>
      <li>
        <details>
        <summary><a href="#stretch-goals">Stretch goals</a></summary>
        <ol>
          <li><a href="#n-64-bit">16-bit and 64-bit variants</a></li>
          <li><a href="#n-config">Config API rework</a></li>
          <li><a href="#n-bd">Block device API rework</a></li>
          <li><a href="#n-uattrs">Custom attr API rework</a></li>
          <li><a href="#n-alternative-ws">Alternative (cheaper) write-strategies</a></li>
          <li><a href="#n-advanced-holes">Advanced file tree operations</a></li>
          <li><a href="#n-advanced-cow">Advanced file copy-on-write operations</a></li>
          <li><a href="#n-kv">Simple key-value APIs</a></li>
          <li><a href="#n-cow-lockups">Reserved blocks to prevent CoW lockups</a></li>
          <li><a href="#n-meta-lockups">Metadata checks to prevent metadata lockups</a></li>
          <li><a href="#n-bd-ecc">Integrated block-level ECC</a></li>
          <li><a href="#n-raid">Disk-level RAID</a></li>
        </ol>
        </details>
      </li>
      <li>
        <details>
        <summary><a href="#out-of-scope">Out-of-scope (for now)</a></summary>
        <ol>
          <li><a href="#n-alternative-cksums">Alternative checksums</a></li>
          <li><a href="#n-alternative-builds">Feature-limited configurations for smaller code/stack sizes</a></li>
          <li><a href="#n-openat">lfs3_file_openat for dir-relative APIs</a></li>
          <li><a href="#n-openn">lfs3_file_openn for non-null-terminated-string APIs</a></li>
          <li><a href="#n-compr">Transparent compression</a></li>
          <li><a href="#n-shrink">Filesystem shrinking</a></li>
          <li><a href="#n-advanced-caches">High-level caches</a></li>
          <li><a href="#n-symlinks">Symbolic links</a></li>
          <li><a href="#n-cov">100% line/branch coverage</a></li>
        </ol>
        </details>
      </li>
    </ol>
    </details>
  </li>
  <li>
    <details>
    <summary><a href="#code-size">Code/stack size</a></summary>
    <ol>
      <li><a href="#c-error-recovery">Runtime error recovery</a></li>
      <li><a href="#c-btree-flexibility">B-tree flexibility</a></li>
      <li><a href="#c-traversal-inversion">Traversal inversion</a></li>
    </ol>
    </details>
  </li>
  <li>
    <details>
    <summary><a href="#benchmarks">Benchmarks</a></summary>
    <ol>
      <li>
        <details>
        <summary><a href="#simulated-benchmarks">Simulated benchmarks</a></summary>
        <ol>
          <li><a href="#b-linear-writes">Linear writes</a></li>
          <li><a href="#b-random-writes">Random writes</a></li>
          <li><a href="#b-logging">Logging</a></li>
        </ol>
        </details>
      </li>
    </ol>
    </details>
  </li>
  <li><a href="#funding">Funding</a></li>
  <li><a href="#next-steps">Next steps</a></li>  
</ol>

## Hello! <a href="#table-of-contents" id="hello">^</a>

Hello everyone! As some of you may have already picked up on, there's been a large body of work fermenting in the background for the past couple of years. Originally started as an experiment to try to solve littlefs's $O(n^2)$ metadata compaction, this branch eventually snowballed into more-or-less a full rewrite of the filesystem from the ground up.

There's still several chunks of planned work left, but now that this branch has reached **feature parity** with v2, there's nothing really stopping it from being merged eventually.

So I figured it's a good time to start calling this v3, and put together a public roadmap.

**NOTE: THIS WORK IS INCOMPLETE AND UNSTABLE**

Here's a quick TODO list of planned work before stabilization. More details below:

- [x] Test framework rework (merged, mostly)
- [x] Rbyds 
- [x] B-trees
- [x] M-tree
- [x] B-shrubs
- [x] Fruncate
- [x] Sync model rework
- [x] Stickynotes
- [x] Traversal API rework
- [x] Incremental GC
- [x] Global-checksums
- [ ] On-disk block-map (in progress)
- [ ] Metadata redundancy
- [ ] Data redundancy
- [ ] Document, document, document
- [ ] Stretch goals

**This work may continue to break the on-disk format.**

That being said, I highly encourage others to experiment with v3 where possible. Feedback is welcome, and immensely important at this stage. Once it's stabilized, it's stabilized.

To help with this, the current branch uses a v0.0 as its on-disk version to indicate that it is experimental. When it is eventually released, v3 will reject this version and fail to mount.

Unfortunately, the API _will_ be under heavy flux during this period. 

**A note on benchmarking:** The on-disk block-map is **key** for scalable allocator performance, so benchmarks at this stage needs to be taken with a grain of salt when many blocks are involved. Please refer to this version as "v3 (no bmap)" or something similar in any published benchmarks until this work is completed.

## Wait, a disk breaking change? <a href="#table-of-contents" id="disk-breaking">^</a>

Yes. v3 breaks disk compatibility from v2.

I think this is a necessary evil. Attempting to maintain backwards compatibility has a heavy cost:

1. Development time - The littlefs team is ~1 guy, and v3 has already taken ~2.5 years. The extra work to make everything compatible would stretch this out _much_ longer and likely be unsustainable.

2. Code cost - The goal of littlefs is to be, well, little. This is unfortunately in conflict with backwards compatibility.

   Take the new B-tree data-structure, for example. It would be easy to support both B-tree and CTZ skip-list files, but now you need ~2x the code. This cost gets worse for the more enmeshed features, and potentially exceeds the cost of just including both v3 and v2 in the codebase.

So I think it's best for both littlefs as a project and long-term users to break things here.

Note v2 isn't going anywhere! I'm happy to continue maintaining the [v2][v2] branch, merge bug fixes when necessary, etc. But the economic reality is my focus will be shifting to v3.

## What's new <a href="#table-of-contents" id="whats-new">^</a>

Ok, with that out of the way, what does breaking everything actually get us?

### Implemented: <a href="#table-of-contents" id="implemented">^</a>

- Efficient metadata compaction: $O(n^2) \rightarrow O(n \log n)$ <a href="#table-of-contents" id="n-meta-compaction">^</a>

  v3 adopts a new metadata data-structure: Red-black-yellow Dhara trees (rbyds). Based on the data-structure invented by Daniel Beer for the [Dhara FTL][dhara], rbyds extend log-encoded [Dhara trees][dhara-tree] with [self-balancing][rb-tree] and [self-counting][cb-tree] (also called [order-statistic][os-tree]) properties.

  This speeds up most metadata operations, including metadata lookup ( $O(n) \rightarrow O(\log n)$ ), and, critically, metadata compaction ( $O(n^2) \rightarrow  O(n \log n)$ ).

  This improvement may sound minor on paper, but it's a difference measured in seconds, sometimes even minutes, on devices with extremely large blocks.

- Efficient random writes: $O(n) \rightarrow O(\log_b^2 n)$ <a href="#table-of-contents" id="n-random-writes">^</a>

  A much requested feature, v3 adopts [B-trees][b-tree], replacing the CTZ skip-list that previously backed files.

  This avoids needing to rewrite the entire file on random reads, bringing the performance back down into tractability.

  For extra cool points, littlefs's B-trees use rbyds for the inner nodes, which makes CoW updates much cheaper than traditional array-packed B-tree nodes when large blocks are involved ( $O(n) \rightarrow O(\log n)$ ).

- Better logging: No more [sync-padding issues][sync-padding-disc] <a href="#table-of-contents" id="n-logging">^</a>

  v3's B-trees support inlining data directly in the B-tree nodes. This gives us a place to store data during sync, without needing to pad things for prog alignment.

  In v2 this padding would force the rewriting of blocks after sync, which had a tendency to wreck logging performance.

- Efficient inline files, no more RAM constraints: $O(n^2) \rightarrow O(n \log n)$ <a href="#table-of-contents" id="n-inline-files">^</a>

  In v3, B-trees can have their root inlined in the file's mdir, giving us what I've been calling a "B-shrub". This, combined with the above inlined leaves, gives us a much more efficient inlined file representation, with better code reuse to boot.

  Oh, and B-shrubs also make small B-trees more efficient by avoiding the extra block needed for the root.

- Independent file caches <a href="#table-of-contents" id="n-caches">^</a>

  littlefs's `pcache`, `rcache`, and file caches can be configured independently now. This should allow for better RAM utilization when tuning the filesystem.

- Easier logging APIs: `lfs3_file_fruncate` <a href="#table-of-contents" id="n-fruncate">^</a>

  Thanks to the new [self-counting][cb-tree]/[order-statistic][os-tree] properties, littlefs can now truncate from both the end and _front_ of files via the new `lfs3_file_fruncate` API.

  Before, the best option for logging was renaming log files when they filled up. Now, maintaining a log/FIFO is as easy as:

  ``` c
  lfs3_file_write(&lfs, &file, entry, entry_size) => entry_size;
  lfs3_file_fruncate(&lfs, &file, log_size) => 0;
  ```

- Sparse files <a href="#table-of-contents" id="n-holes">^</a>

  Another advantage of adopting B-trees, littlefs can now cheaply represent [file holes][sparse-file], where contiguous runs of zeros can be implied without actually taking up any disk space.

  Currently this is limited to a couple operations:

  - `lfs3_file_truncate`
  - `lfs3_file_fruncate`
  - `lfs3_file_seek` + `lfs3_file_write` past the end of the file

  But more advanced hole operations may be added in the future.

- Efficient file name lookup: $O(n) \rightarrow O(\log_b n)$ <a href="#table-of-contents" id="n-namelookups">^</a>

  littlefs now uses a B-tree (yay code reuse) to organize files by file name. This allows for much faster file name lookup than the previous linked-list of metadata blocks.

- A simpler/more robust metadata tree <a href="#table-of-contents" id="n-mtree">^</a>

  As a part of adopting B-trees for metadata, the previous threaded file tree has been completely ripped out and replaced with one big metadata tree: the M-tree.

  I'm not sure how much users are aware of it, but the previous threaded file tree was a real _pain-in-the-ass_ with the amount of bugs it caused. Turns out having a [fully-connected graph][graph] in a CoBW filesystem is a really bad idea.

  In addition to removing an entire category of possible bugs, adopting the M-tree allows for multiple directories in a single metadata block, removing the 1-dir = 1-block minimum requirement.

- A well-defined sync model <a href="#table-of-contents" id="n-sync-model">^</a>

  One interesting thing about littlefs, it doesn't have a strictly POSIX API. This puts us in a relatively unique position, where we can explore tweaks to the POSIX API that may make it easer to write powerloss-safe applications.

  To leverage this (and because the previous sync model had some real problems), v3 includes a new, well-defined sync model.

  I think [this discussion][desync-disc] captures most of the idea, but for a high-level overview:

  1. Open file handles are strictly snapshots of the on-disk state. Writes to a file are copy-on-write (CoW), with no immediate affect to the on-disk state or any other file handles.

  2. Syncing or closing an in-sync file atomically updates the on-disk state and any other in-sync file handles.

  3. Files can be desynced, either explicitly via `lfs3_file_desync`, or because of an error. Desynced files do not recieve sync broadcasts, and closing a desynced file has no affect on the on-disk state.

  4. Calling `lfs3_file_sync` on a desynced file will atomically update the on-disk state, any other in-sync file handles, and mark the file as in-sync again.

  5. Calling `lfs3_file_resync` on a file will discard its current contents and mark the file as in-sync. This is equivalent to
     closing and reopening the file.

- Stickynotes, no more [0-sized files][0-sized-disc] <a href="#table-of-contents" id="n-stickynotes">^</a>

  As an extension of the littlefs's new sync model, v3 introduces a new file type: `LFS3_TYPE_STICKYNOTE`.

  A stickynote represents a file that's in the awkward state of having been created, but not yet synced. If you lose power, stickynotes are hidden from the user and automatically cleaned up on the next mount.

  The avoids the [0-sized file issue][0-sized-disc], while still allowing most of the POSIX interactions users expect.

- A new and improved compat flag system <a href="#table-of-contents" id="n-compat">^</a>

  v2.1 was a bit of a mess, but it was a learning experience. v3 still includes a global version field, but also includes a set of compat flags that allow non-linear addition/removal of future features.

  These are probably familiar to users of Linux filesystems, though I've given them slightly different names:

  - `rcompat flags` - Must understand to read the filesystem ([incompat_flags][compat-flags])
  - `wcompat flags` - Must understand to write to the filesystem ([ro_compat_flags][compat-flags])
  - `ocompat flags` - No understanding necessary ([compat_flags][compat-flags])

  This also provides an easy route for marking a filesystem as read-only, non-standard, etc, on-disk.

- Error detection! - Global-checksums <a href="#table-of-contents" id="n-gcksums">^</a>

  v3 now supports filesystem-wide error-detection. This is actually quite tricky in a CoBW filesystem, and required the invention of [global-checksums (gcksums)][gcksums] to prevent rollback issues caused by naive checksumming.

  With [gcksums][gcksums], and a traditional [Merkle-tree-esque][merkle-tree] B-tree construction, v3 now provides a filesystem-wide self-validating checksum via `lfs3_fs_cksum`. This checksum can be stored external to the filesystem to provide protection against last-commit rollback issues, [metastability][metastability-disc], or just for that extra peace of mind.

  Funny thing about checksums. It's incredibly cheap to calculate checksums when writing, as we're already processing that data anyways. The hard part is, when do you check the checksums?

  This is a problem that mostly ends up on the user, but to help, v3 adds a large number checksum checking APIs (probably too many if I'm honest):

  - `LFS3_M_CKMETA/CKDATA` - Check checksums during mount
  - `LFS3_O_CKMETA/CKDATA` - Check checksums during file open
  - `lfs3_fs_ckmeta/ckdata` - Explicitly check all checksums in the filesystem
  - `lfs3_file_ckmeta/ckdata` - Explicitly check a file's checksums
  - `LFS3_T_CKMETA/CKDATA` - Check checksums incrementally during a traversal
  - `LFS3_GC_CKMETA/CKDATA` - Check checksums during GC operations
  - `LFS3_M_CKPROGS` - Closed checking of data during progs
  - `LFS3_M_CKFETCHES` - Optimistic (not closed) checking of data during fetches
  - `LFS3_M_CKREADS` (planned) - Closed checking of data during reads

- Better traversal APIs <a href="#table-of-contents" id="n-traversals">^</a>

  The traversal API has been completely reworked to be easier to use (both externally and internally).

  No more callback needed, blocks can now be iterated over via the dir-like `lfs3_traversal_read` function.

  Traversals can also perform janitorial work and check checksums now, based on the flags provided to `lfs3_traversal_open`.

- Incremental GC <a href="#table-of-contents" id="n-gc">^</a>

  GC work can now be accomplished incrementally, instead of requiring one big go. This is managed by `lfs3_fs_gc`, `cfg.gc_flags`, and `cfg.gc_steps`.

  Internally, this just shoves one of the new traversal objects into `lfs3_t`. It's equivalent to managing a traversal object yourself, but hopefully makes it easier to write library code.

  However, this does add a significant chunk of RAM to `lfs3_t`, so GC is now an opt-in feature behind the `LFS3_GC` ifdef.

- Better recovery from runtime errors <a href="#table-of-contents" id="n-error-recovery">^</a>

  Since we're already doing a full rewrite, I figured let's actually take the time to make sure things don't break on exceptional errors.

  Most in-RAM filesystem state should now revert to the last known-good state on error.

  The one exception involves file _data_ (not metadata!). Reverting file data correctly turned out to roughly double the cost of files. And now that you can manual revert with `lfs3_file_resync`, I figured this cost just isn't worth it. So file _data_ remains undefined after an error.

  In total, these changes add a significant amount of code and stack, but I'm of the opinion this is necessary for the maturing of littlefs as a filesystem.

- Standard custom attributes <a href="#table-of-contents" id="n-sattrs">^</a>

  Breaking disk gives us a chance to reserve attributes `0x80-0xbf` for future standard custom attributes:

  - `0x00-0x7f` - Free for user-attributes (uattr) 
  - `0x80-0xbf` - Reserved for standard-attributes (sattr)
  - `0xc0-0xff` - Encouraged for system-attributes (yattr)

  In _theory_, it was technically possible to reserve these attributes without a disk-breaking change, but it's much safer to do so while we're already breaking the disk.

  v3 also includes the possibility of extending the custom attribute space from 8-bits to ~25-bits in the future, but I'd hesitate to  to use this, as it risks a significant increase in stack usage.

- More tests! <a href="#table-of-contents" id="n-tests">^</a>

  v3 comes with a couple more tests than v2 (+~6812.2%):

  ```
       suites  cases  permutations ;     pls   runtime
  v2:      22    198         11641 ;   28741    54.16s
  v3:      23    784        804655 ; 2228513  1323.18s
  ```

  You may or may not have seen the [test framework rework][test-rework-3] that went curiously under-utilized. That was actually in preparation for the v3 work.

  The goal is not 100% line/branch coverage, but just to have more confidence in littlefs's reliability.

### Planned: <a href="#table-of-contents" id="planned">^</a>

- Efficient block allocation, via optional on-disk block-map (bmap) <a href="#table-of-contents" id="n-bmap">^</a>

  The one remaining bottleneck in v3 is block allocation. This is a tricky problem for littlefs (and any CoW/CoBW filesystem), because we don't actually know when a block becomes free.

  This is in-progress work, but the solution I'm currently looking involves 1. adding an optional on-disk block map (bmap) stored in gstate, and 2. updating it via tree diffing on sync. In theory this will drop huge file writes: $O(n^2 \log n) \rightarrow O(n \log_b^2 n)$

  There is also the option of using the bmap as a simple cache, which doesn't avoid the filesystem-wide scan but at least eliminates the RAM constraint of the lookahead buffer.

  As a plus, we should be able to leverage the self-counting property of B-trees to make the on-disk bmap compressible.

- Bad block tracking <a href="#table-of-contents" id="n-bad-blocks">^</a>

  This is a much requested feature, and adding the optional on-disk bmap finally gives us a place to track bad blocks.

- Pre-erased block tracking <a href="#table-of-contents" id="n-preerased-blocks">^</a>

  Just like bad-blocks, the optional on-disk bmap gives us a place to track pre-erased blocks. Well, at least in theory.

  In practice it's a bit more of a nightmare. To avoid multiple progs, we need to mark erased blocks as unerased before progging. This introduces an unbounded number of catch-22s when trying to update the bmap itself.

  Fortunately, if instead we store a simple counter in the bmap's gstate, we can resolve things at the mrootanchor worst case.

- Error correction! - Metadata redundancy <a href="#table-of-contents" id="n-meta-redund">^</a>

  Note it's already possible to do error-correction at the block-device level outside of littlefs, see [ramcrc32cbd][ramcrc32cbd] and [ramrsbd][ramrsbd] for examples. Because of this, integrating in-block error correction is low priority.
  
  But I think there's potential for _cross-block_ error-correction in addition to the _in-block_ error-correction.

  The plan for cross-block error-correction/block redundancy is a bit different for metadata vs data. In littlefs, all metadata is logs, which is a bit of a problem for parity schemes. I think the best we can do is store metadata redundancy as naive copies.

  But we already need two blocks for every mdir, one usually just sits unused when not compacting. This, combined with metadata usually being much smaller than data, makes the naive scheme less costly than one might expect.

- Error correction! - Data redundancy <a href="#table-of-contents" id="n-data-redund">^</a>

  For raw data blocks, we can be a bit more clever. If we add an optional dedup tree for block -> parity group mapping, and an optional parity tree for parity blocks, we can implement a [RAID-esque][raid-6] parity scheme for up to 3 blocks of data redundancy relatively cheaply.

- Transparent block deduplication <a href="#table-of-contents" id="n-dedup">^</a>

  This one is a bit funny. Originally block deduplication was intentionally out-of-scope, but it turns out you need something that looks a lot like a dedup tree for error-correction to work in a system that allows multiple block references.

  If we already need a virtual -> physical block mapping for error correction, why not make the key the block checksum and get block deduplication for free?

  Though if this turns out to not be as free as I think it is, block deduplication will fall out-of-scope.

### Stretch goals: <a href="#table-of-contents" id="stretch-goals">^</a>

These may or may not be included in v3, depending on time and funding:

- 16-bit and 64-bit variants <a href="#table-of-contents" id="n-64-bit">^</a>

- Config API rework <a href="#table-of-contents" id="n-config">^</a>

- Block device API rework <a href="#table-of-contents" id="n-bd">^</a>

- Custom attr API rework <a href="#table-of-contents" id="n-uattrs">^</a>

- Alternative (cheaper) write-strategies (write-once, global-aligned, eager-crystallization) <a href="#table-of-contents" id="n-alternative-ws">^</a>

- Advanced file tree operations (`lfs3_file_punchhole`, `lfs3_file_insertrange`, `lfs3_file_collapserange`, `LFS3_SEEK_DATA`, `LFS3_SEEK_HOLE`) <a href="#table-of-contents" id="n-advanced-holes">^</a>

- Advanced file copy-on-write operations (shallow `lfs3_cowcopy` + opportunistic `lfs3_copy`) <a href="#table-of-contents" id="n-advanced-cow">^</a>

- Simple key-value APIs (`lfs3_get`, `lfs3_set`, `lfs3_remove`, etc) <a href="#table-of-contents" id="n-kv">^</a>

- Reserved blocks to prevent CoW lockups <a href="#table-of-contents" id="n-cow-lockups">^</a>

- Metadata checks to prevent metadata lockups <a href="#table-of-contents" id="n-meta-lockups">^</a>

- Integrated block-level ECC ([ramcrc32cbd][ramcrc32cbd], [ramrsbd][ramrsbd]) <a href="#table-of-contents" id="n-bd-ecc">^</a>

- [Disk-level RAID][raid-6] (this is just data redund + a disk aware block allocator) <a href="#table-of-contents" id="n-raid">^</a>

### Out-of-scope (for now): <a href="#table-of-contents" id="out-of-scope">^</a>

If we don't stop somewhere, v3 will never be released. But these may be added in the future:

- Alternative checksums (crc16, crc64, sha256, etc) <a href="#table-of-contents" id="n-alternative-cksums">^</a>

- Feature-limited configurations for smaller code/stack sizes (`LFS3_NO_DIRS`, `LFS3_KV`, `LFS3_2BLOCK`, etc) <a href="#table-of-contents" id="n-alternative-builds">^</a>

- [`lfs3_file_openat`][openat-disc] for dir-relative APIs <a href="#table-of-contents" id="n-openat">^</a>

- [`lfs3_file_openn`][openn-disc] for non-null-terminated-string APIs <a href="#table-of-contents" id="n-openn">^</a>

- [Transparent compression][compr-disc] <a href="#table-of-contents" id="n-compr">^</a>

- Filesystem shrinking <a href="#table-of-contents" id="n-shrink">^</a>

- High-level caches (block cache, mdir cache, btree leaf cache, etc) <a href="#table-of-contents" id="n-advanced-caches">^</a>

- Symbolic links <a href="#table-of-contents" id="n-symlinks">^</a>

- 100% line/branch coverage <a href="#table-of-contents" id="n-cov">^</a>

## Code/stack size <a href="#table-of-contents" id="code-size">^</a>

<p align="center">
<a href="http://benchmarks.littlefs.org/plots/codemap_lfs1.svg"                
    ></a>
<a href="http://benchmarks.littlefs.org/plots/codemap_lfs2.svg"                
    ></a>
<a href="http://benchmarks.littlefs.org/plots/codemap_lfs3.svg"
    ></a>
<br>
littlefs v1, v2, and v3, 1 pixel ~= 1 byte of code,
click for a larger interactive codemap
(<a
  href="https://github.com/littlefs-project/littlefs-benchmarks/tree/be7a47bbd02b093828d9e61536e954cf47be9d5e"
  >commit</a>)
</p>

Unfortunately, v3 is a little less little than v2:

```
     code            stack           ctx
v2: 17144             1440           580
v3: 37300 (+117.6%)   2280 (+58.3%)  636 (+9.7%)
```

On one hand, yes, more features generally means more code.

And it's true there's an opportunity here to carve out more feature-limited builds to save code/stack in the future.

But I think it's worth discussing some of the other reasons for the code/stack increase:

1. Runtime error recovery <a href="#table-of-contents" id="c-error-recovery">^</a>

   Recovering from runtime errors isn't cheap. We need to track both the before and after state of things during fallible operations, and this adds both stack and code.

   But I think this is necessary for the maturing of littlefs as a filesystem.

   Maybe it will make sense to add a sort of `LFS3_GLASS` mode in the future, but this is out-of-scope for now.

2. B-tree flexibility <a href="#table-of-contents" id="c-btree-flexibility">^</a>

   The bad news: The new B-tree files are extremely flexible. Unfortunately, this is a double-edged sword.

   B-trees, on their own, don't add that much code. They are a relatively poetic data-structure. But _deciding_ how to write to a B-tree, efficiently, with an unknown write pattern, is surprisingly tricky.

   The current implementation, what I've taken to calling the "lazy-crystallization algorithm", leans on the more complicated side to see what is possible performance-wise.

   The good news: The new B-tree files are extremely flexible.

   There's no reason you need the full crystallization algorithm if you have a simple write pattern, or don't care as much about performance. This will either be a future or stretch goal, but it would be interesting to explore alternative write-strategies that could save code in these cases.

3. Traversal inversion <a href="#table-of-contents" id="c-traversal-inversion">^</a>

   Inverting the traversal, i.e. moving from a callback to incremental state machine, adds both code and stack as 1. all of the previous on-stack state needs to be tracked explicitly, and 2. we now need to worry about what happens if the filesystem is modified mid-traversal.

   In theory, this could be reverted if you don't need incremental traversals, but extricating incremental traversals from the current codebase would be an absolute nightmare, so this is out-of-scope for now.

## Benchmarks <a href="#table-of-contents" id="benchmarks">^</a>

**A note on benchmarking:** The on-disk block-map is **key** for scalable allocator performance, so benchmarks at this stage needs to be taken with a grain of salt when many blocks are involved. Please refer to this version as "v3 (no bmap)" or something similar in any published benchmarks until this work is completed.

First off, I would highly encourage others to do their own benchmarking with v3/v2. Filesystem performance is tricky to measure because it depends heavily on your application's write pattern and hardware nuances. If you do, please share in this thread! Others may find the results useful, and now is the critical time for finding potential disk-related performance issues.

### Simulated benchmarks <a href="#table-of-contents" id="simulated-benchmarks">^</a>

To test the math behind v3, I've put together some preliminary simulated benchmarks.

Note these are **simulated** and **optimistic**. They do not take caching or hardware buffers into account, which can have a big impact on performance. Still, I think they provide at least a good first impression of v3 vs v2.

To find an estimate of runtime, I first measured the amount of bytes read, progged, and erased, and then [scaled][benchmarks-scaling] based on values found in relevant datasheets. The options here were a bit limited, but WinBond fortunately provides runtime estimates in the datasheets on their website:

- NOR flash - [w25q64jv][w25q64jv]

- NAND flash - [w25n01gv][w25n01gv]

- SD/eMMC - Also [w25n01gv][w25n01gv], assuming a perfect FTL

  I said optimistic, didn't I? I could't find useful estimates for SD/eMMC, so I'm just assuming a perfect FTL here.

These also assume an optimal bus configuration, which, as any embedded engineer knows, is often not the case.

**Full benchmarks here:** https://benchmarks.littlefs.org ([repo][benchmarks-repo], [commit][benchmarks-commit])

And here are the ones I think are the most interesting:

Note that SD/eMMC is heavily penalized by the lack of on-disk block-map! SD/eMMC breaks flash down into many small blocks, which tends to make block allocator performance dominate.

1. Linear writes, where we write a 1 MiB file and don't call sync until closing the file. <a href="#table-of-contents" id="b-linear-writes">^</a>

   This one is the most frustrating to compare against v2. CTZ skip-lists are really fast at appending! The problem is they are _only_ fast at appending:

   <p align="center">
   
   (<a
     href="https://github.com/littlefs-project/littlefs-benchmarks/tree/be7a47bbd02b093828d9e61536e954cf47be9d5e"
     >commit</a>, <a
     href="http://benchmarks.littlefs.org/plots_1M/bench_vs_lfs2_fwrite_linear_write_r.svg"
     >reads</a>, <a
     href="http://benchmarks.littlefs.org/plots_1M/bench_vs_lfs2_fwrite_linear_write_p.svg"
     >progs</a>, <a
     href="http://benchmarks.littlefs.org/plots_1M/bench_vs_lfs2_fwrite_linear_write_e.svg"
     >erases</a>, <a
     href="http://benchmarks.littlefs.org/plots_1M/bench_vs_lfs2_fwrite_linear_write.svg"
     >sim</a>, <a
     href="http://benchmarks.littlefs.org/plots_1M/bench_vs_lfs2_fwrite_linear_usage.svg"
     >usage</a>)
   </p>
      
2. Random writes, note we start with a 1MiB file. <a href="#table-of-contents" id="b-random-writes">^</a>

   As expected, v2 is comically bad at random writes. v3 is indistinguishable from zero in the NOR case:

   <p align="center">
   
   (<a
     href="https://github.com/littlefs-project/littlefs-benchmarks/tree/be7a47bbd02b093828d9e61536e954cf47be9d5e"
     >commit</a>, <a
     href="http://benchmarks.littlefs.org/plots_1M/bench_vs_lfs2_fwrite_random_write_r.svg"
     >reads</a>, <a
     href="http://benchmarks.littlefs.org/plots_1M/bench_vs_lfs2_fwrite_random_write_p.svg"
     >progs</a>, <a
     href="http://benchmarks.littlefs.org/plots_1M/bench_vs_lfs2_fwrite_random_write_e.svg"
     >erases</a>, <a
     href="http://benchmarks.littlefs.org/plots_1M/bench_vs_lfs2_fwrite_random_write.svg"
     >sim</a>, <a
     href="http://benchmarks.littlefs.org/plots_1M/bench_vs_lfs2_fwrite_random_usage.svg"
     >usage</a>)
   </p>

3. Logging, write 4 MiB, but limit the file to 1 MiB. <a href="#table-of-contents" id="b-logging">^</a>

   In v2 this is accomplished by renaming the file, in v3 we can leverage `lfs3_file_fruncate`.

   v3 performs significantly better with large blocks thanks to avoiding the sync-padding problem:

   <p align="center">
   
   (<a
     href="https://github.com/littlefs-project/littlefs-benchmarks/tree/be7a47bbd02b093828d9e61536e954cf47be9d5e"
     >commit</a>, <a
     href="http://benchmarks.littlefs.org/plots_1M/bench_vs_lfs2_logging_write_r.svg"
     >reads</a>, <a
     href="http://benchmarks.littlefs.org/plots_1M/bench_vs_lfs2_logging_write_p.svg"
     >progs</a>, <a
     href="http://benchmarks.littlefs.org/plots_1M/bench_vs_lfs2_logging_write_e.svg"
     >erases</a>, <a
     href="http://benchmarks.littlefs.org/plots_1M/bench_vs_lfs2_logging_write.svg"
     >sim</a>, <a
     href="http://benchmarks.littlefs.org/plots_1M/bench_vs_lfs2_logging_usage.svg"
     >usage</a>)
   </p>

## Funding <a href="#table-of-contents" id="funding">^</a>

If you think this work is worthwhile, consider [sponsoring littlefs][gh-sponsor]. Current benefits include:

1. Being able to complain about v3 not being released yet
2. Being able to complain about the disk breakage v2 -> v3

I joke, but I truly appreciate those who have contributed to littlefs so far. littlefs, in its current form, is a mostly self-funded project, so every little bit helps.

If you would like to contribute in a different way, or have other requests, feel free to reach me at geky at geky.net.

As stabilization gets closer, I will also be open to contract work to help port/integrate/adopt v3. If this is interesting to anyone, let me know.

Thank you @micropython, @fusedFET for sponsoring littlefs, and thank you @eclo, @kmetabg, and @nedap for your past sponsorships!

## Next steps <a href="#table-of-contents" id="next-steps">^</a>

For me, I think it's time to finally put together a website/wiki/discussions/blog. I'm not sure on the frequency quite yet, but I plan to write/publish the new DESIGN.md in chapters in tandem with the remaining work.

[v2]: https://github.com/littlefs-project/littlefs/tree/v2

[dhara]: https://github.com/dlbeer/dhara
[dhara-tree]: https://github.com/dlbeer/dhara/blob/master/map_internals.txt
[rb-tree]: https://en.wikipedia.org/wiki/Red%E2%80%93black_tree
[cb-tree]: https://www.chiark.greenend.org.uk/~sgtatham/algorithms/cbtree.html
[os-tree]: https://en.wikipedia.org/wiki/Order_statistic_tree
[b-tree]: https://en.wikipedia.org/wiki/B-tree
[sync-padding-disc]: https://github.com/littlefs-project/littlefs/issues/862
[sparse-file]: https://en.wikipedia.org/wiki/Sparse_file
[graph]: https://en.wikipedia.org/wiki/Graph_(discrete_mathematics)
[desync-disc]: https://github.com/littlefs-project/littlefs/pull/960
[0-sized-disc]: https://github.com/littlefs-project/littlefs/issues/908
[compat-flags]: https://docs.kernel.org/filesystems/ext2.html#feature-compatibility
[gcksums]: https://github.com/littlefs-project/littlefs/commit/1c5adf71b38f8eb26e5a00798bfdb5eac6c20a74
[merkle-tree]: https://en.wikipedia.org/wiki/Merkle_tree
[metastability-disc]: https://github.com/littlefs-project/littlefs/issues/671
[test-rework-3]: https://github.com/littlefs-project/littlefs/pull/752

[ramcrc32cbd]: https://github.com/littlefs-project/ramcrc32bd
[ramrsbd]: https://github.com/littlefs-project/ramrsbd
[raid-6]: https://en.wikipedia.org/wiki/Standard_RAID_levels#RAID_6

[openat-disc]: https://github.com/littlefs-project/littlefs/issues/42
[openn-disc]: https://github.com/littlefs-project/littlefs/issues/934
[compr-disc]: https://github.com/littlefs-project/littlefs/issues/672

[benchmarks]: https://benchmarks.littlefs.org
[benchmarks-repo]: https://github.com/geky/littlefs-benchmarks
[benchmarks-commit]: https://github.com/geky/littlefs-benchmarks/tree/abc7e667c862d6f176d7576485c7410203a48012
[benchmarks-scaling]: https://github.com/geky/littlefs-benchmarks/blob/abc7e667c862d6f176d7576485c7410203a48012/Makefile#L41-L55
[w25q64jv]: https://www.winbond.com/resource-files/W25Q256JV%20SPI%20RevQ%2002072025%20Plus.pdf
[w25n01gv]: https://www.winbond.com/resource-files/W25N01GV%20Rev%20R%20070323.pdf

[gh-sponsor]: https://github.com/sponsors/littlefs-project


geky:
There is high risk this thread drowns in noise. So, **please limit this thread to high-level v3-wide discussion.**

For bugs, low-level questions, concerns, etc, feel free to create issues on littlefs with "v3" in the title.

dpgeorge:
Fantastic!  Great to see this v3 work making it's way into the world.

The code size increase would be the biggest concern for MicroPython.  Any optimisations to get that down would be welcome.  In particular we put read-only littlefs2 inside a 32k bootloader.  Do you have any idea how big the read-only version of v3 is?

geky:
@dpgeorge, I don't know quite yet. The non-default builds aren't fully reimplemented, but will update this thread when they are. Read-only is the most interesting but also the most time consuming.

The good news is that rbyds (and red-black trees) and B-trees are much more complicated on the write side than read side, so the impact should be less than the default build, but I'll wait to stick my foot in my mouth until after I have actual numbers.

geky:
@Ryan-CW-Code, it's worth noting the simple key-value APIs are stretch goals. They may or may not make it into v3 depending on how long things take and funding.

Or to put it another way, stretch goals won't block the release of v3. But it would be convenient to get them in while breaking the API, to minimize future API breaks.

Though if they don't make it into v3, I would like to introduce them in future releases, if possible. Note most stretch and out-of-scope features are mainly API related unlikely to affect the on-disk format.

Ryan-CW-Code:
My English is not very good, please forgive me for using machine translation.
Thanks for your contribution!
Using kvdb and tsdb in the file system is a common function.
We are preparing to migrate from flashdb to littlefs, and have implemented kvdb and tsdb based on littlefs, but due to the long wait for metadata compression, we perform gc operations in a low-priority task every time we make a file change to obtain constant write time.
I noticed that v3 was a good opportunity, with improved metadata compression, and a lot of performance improvements, and I was very willing to test the v3 version.

geky:
@dpgeorge Initial readonly results:

<p>
<a href="https://benchmarks.littlefs.org/tree.html#commit=97b944f9a6b8390c4380fde2348f224b5dca3ded&path=codemaps/codemap_lfs2_rdonly.svg"
    ></a>
<a href="https://benchmarks.littlefs.org/tree.html#commit=97b944f9a6b8390c4380fde2348f224b5dca3ded&path=codemaps/codemap_lfs3_rdonly.svg"
    ></a>
</p>

```
            code            stack           ctx
v2-rdonly:  6270              448           580
v3-rdonly: 10616 (+69.3%)     808 (+80.4%)  508 (-12.4%)
```

Digging into the code cost, it's interesting to note most of it comes from the added complexity to filesystem traversal. The threaded linked-list is a significant mess for the write path, but it does make filesystem traversal dead simple.

This hits v3 with the double whammy of both:

1. Filesystem traversal is more complicated since we now need to keep track of where in the B-tree, and which B-tree we're in
2. We need to manage traversal state explicitly since we no longer rely on callbacks

Name lookup is also a bit more costly now that we're binary searching over tree lookups.

It may be possible to shave off another ~1KiB or so by disabling dirs, traversals, cksum checking, etc, (or by letting the compiler gc them), but I think the above costs will dominate.

MouriNaruto:
It's really nice to see this v3 work. It seems improved a lot. I want to say some of my opinions:

- I hope littlefs v3 can test and document some information like maximum volume size and file size. (both in theoretical and recommended) It will help for littlefs users to know how to use littlefs better.
- I respect with v3 will have breaking changes and think it will make us have a better design. But it will have more challenges to support all (in current stage should be three) major on-disk versions of littlefs images in lightweight scenarios like some MCU devices which need to accept users external storages which format with littlefs.
- In the current stage, we know littlefs v2 implementation's little v1 implementation only have the v1 to v2 migration. I hope littlefs will have a separate version migration implementation source files instead of making inline in lfs*.h and lfs*.c. I think it will be useful for you to have more on-disk versions. (I think it's predictable. Maybe you will create v4 even newer on-disk formats before 2030.) Also, it will makes littlefs implementation more simplified. I hope side by side support of littlefs v3 like v1-prefix and v2-prefix branches will be the default mode, and it will be helpful for users to add littlefs more vNext on-disk format support versions in the future.

Kenji Mouri

PR #1131, opened 2025-08-08
add: cmakelist

Anton-Fat:


BenBE:
Не могли бы вы написать комментарии в коде на английском? Так они будут полезны гораздо более широкой аудитории.

geky:
I think the best path for cmake support at the moment would be if someone volunteered to host/maintain a CMakeLists.txt outside of this org.

We could add a link to it in the README.md (and eventually website) to direct potential users, like we have with littlefs-python, littlefs2-rust, etc.

If such a CMakeLists.txt matures, and there's interest, I can eventually bring it into this org as an "official(?)" repo and volunteer to maintain it. But 1. I don't think it's high-priority, and 2. I'm not yet convinced there's a single CMakeLists.txt that covers enough use cases to be worth the maintenance headache.

Issue #904, opened 2023-12-13
Access to buffers returned as corrupted by the block layer


BenBE:
While reviewing littlefs we (@Maaxxs, @stoeckmann and me) noticed an issue, where littlefs reads from potentially uninitialized memory.

A simple way to reproduce this is to apply the following patch:
```diff
diff --git a/bd/lfs_rambd.c b/bd/lfs_rambd.c
index a6a05727..d6282191 100644
--- a/bd/lfs_rambd.c
+++ b/bd/lfs_rambd.c
@@ -7,6 +7,8 @@
  */
 #include "bd/lfs_rambd.h"
 
+#include "sanitizer/asan_interface.h"
+
 int lfs_rambd_create(const struct lfs_config *cfg,
         const struct lfs_rambd_config *bdcfg) {
     LFS_RAMBD_TRACE("lfs_rambd_create(%p {.context=%p, "
@@ -69,6 +71,21 @@ int lfs_rambd_read(const struct lfs_config *cfg, lfs_block_t block,
     memcpy(buffer, &bd->buffer[block*bd->cfg->erase_size + off], size);
 
     LFS_RAMBD_TRACE("lfs_rambd_read -> %d", 0);
+
+    if (block == 0 && off == 0 && size == 4) {
+        LFS_RAMBD_TRACE("Corrupted foo! -> %d", LFS_ERR_CORRUPT);
+        ASAN_POISON_MEMORY_REGION(buffer, size);
+        return LFS_ERR_CORRUPT;
+    }
+    if (block == 1 && off == 0 && size == 4) {
+        LFS_RAMBD_TRACE("Negative Serial -> %d", 0);
+        ASAN_UNPOISON_MEMORY_REGION(buffer, size);
+        ((uint8_t*)buffer)[3] |= 0x80;
+        return 0;
+    }
+
+    ASAN_UNPOISON_MEMORY_REGION(buffer, size);
+
     return 0;
 }
```
and compile a small demo program using that rambd driver with ASAN enabled:
```bash
BUILDDIR=$(pwd) gcc -fsanitize=address -g -DLFS_RAMBD_YES_TRACE -DLFS_YES_TRACE -I${BUILDDIR} lfs.c lfs_util.c bd/lfs_rambd.c -o thing thing.c
```
(`thing.c` is the example from the README plus the necessary structs to use `lfs_rambd.c`)

When running `./thing` you get output like this:
```
[...]

PR #1133, opened 2025-08-12
fix: false uninitialized read warning

elupus:
Add asserts on file system reads to make sure
no positive values are returned, which would
make assumptions on error checks invalid.

This fixes clang tidy warnings on uninitialized
reads in uses of lfs_dir_get where only negative
returns are considered errors.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -739,6 +739,7 @@ static lfs_stag_t lfs_dir_getslice(lfs_t *lfs, const lfs_mdir_t *dir,
         int err = lfs_bd_read(lfs,
                 NULL, &lfs->rcache, sizeof(ntag),
                 dir->pair[0], off, &ntag, sizeof(ntag));
+        LFS_ASSERT(err <= 0);
         if (err) {
             return err;
         }
@@ -767,6 +768,7 @@ static lfs_stag_t lfs_dir_getslice(lfs_t *lfs, const lfs_mdir_t *dir,
             err = lfs_bd_read(lfs,
                     NULL, &lfs->rcache, diff,
                     dir->pair[0], off+sizeof(tag)+goff, gbuffer, diff);
+            LFS_ASSERT(err <= 0);
             if (err) {
                 return err;
             }


BenBE:
Just looking at the changed code in this PR, there are the following likely scenarios:
1.  `lfs_bd_read` may on some code paths not properly return a value --> That would be a grave bug in that function
2. `clang-tidy` is buggy -> Very low priority
3. There's some other hidden issue, that triggers this -> If that's the case, a more thorough architectural fix would be needed.

Given there's a similar issue I reported a while back in #904, with potentially uninitialized memory, and the overall design in the core LittleFS codebase not ensuring properly initialized memory (at least as far as I had a look at for LittleFS v2), there may be further similar issues hidden.

I this particular case I'm strongly leaning towards option 1 (buggy `lfs_bd_read` with potential for uninitialized return value).

If that's the case, the PR as presented is insufficient and only fixes symptoms, not bugs.

elupus:
1. The problem is lfs_bd_read is actually implemented as a callout to user supplied reading function. There is one assert in that file already for this case, but not in all paths:

Ie it's checked here: https://github.com/littlefs-project/littlefs/blob/8e251dd675da00342d45dac78b6f627f119aed03/lfs.c#L118

but not here: https://github.com/littlefs-project/littlefs/blob/8e251dd675da00342d45dac78b6f627f119aed03/lfs.c#L96

So I would have hoped clang-tidy would have been happy with fixing that extra assert inside lfs_bd_read, but it was not oddly. Seems it could not follow that assertion.

One better option could have been to change all cases to if (err < 0) to make sure no positive values ever slip out since they brake assumptions and asserts can be turned off.

geky:
> The problem is lfs_bd_read is actually implemented as a callout to user supplied reading function. There is one assert in that file already for this case, but not in all paths:

That would be a bug on our end, thanks for creating a PR!

> So I would have hoped clang-tidy would have been happy with fixing that extra assert inside lfs_bd_read, but it was not oddly. Seems it could not follow that assertion.

I think that makes sense if clang-tidy wants to protect against future implementations of lfs_bd_read. I've adopted a similar pattern in littlefs3 for functions that create sign-unioned errors from non-unioned errors.

Frustratingly, I couldn't come up with an LFS_ASSERT definition that informs GCC about these assumptions without dragging in extra code, so there's a few unnecessary initializations to keep GCC happy. It's neat that clang-tidy can understand these asserts though.

> One better option could have been to change all cases to if (err < 0) to make sure no positive values ever slip out since they brake assumptions and asserts can be turned off.

I've toyed around with this before, but it created more issues than it solved. If a positive error is returned, the logic will fall through to the rest of the code with garbage state, which creates problems/warnings.

geky (approved):

geky:
> Given there's a similar issue I reported a while back in https://github.com/littlefs-project/littlefs/issues/904, with potentially uninitialized memory, and the overall design in the core LittleFS codebase not ensuring properly initialized memory (at least as far as I had a look at for LittleFS v2), there may be further similar issues hidden.

It's worth noting rambd is a special case where we're treating memory as storage. We shouldn't be relying on uninitialized memory in the core code, but uninitialized storage is something littlefs needs to be prepared to handle. (Even if we erased the entire disk during format, storage after powerloss could be considered uninitialized.)

If you need reproducibility you should probably use emubd (coincidentally I'm working on a lighter-weight variant of emubd for benchmarking that looks like rambd + erase emulation + read/prog/erase counters).

PR #658, opened 2022-03-21
Restructure littlefs to not use recursion, measure stack usage

geky:
This PR restructures the internals of littlefs a bit to completely avoid recursion. In theory the design of littlefs uses strictly bounded RAM, though it was written using some recursive functions, and at least one code path relied on tail-call optimization for bounded RAM consumption, which isn't guaranteed in standard C.

This broke down into three pieces of logic that used recursion:

1. lfs_dir_commit/lfs_fs_relocate - 84da4c0b1af465014e80cb48fca26d407a4ab214

   The biggest culprit of recursion was the relocation logic responsible for moving metadata to new blocks when a block goes bad or has high wear. The reason is because this occurs deep within the commit logic and is inherently recursive (any commit can require relocation, which needs to commit to fix the relocation!)
   
   This required quite a bit of restructuring, inverting most of the commit logic so that relocations occur on an outer loop that invokes the core commit logic. The commit logic ends up with several new entry points that "fall back" to outer loops under different error conditions, with the top level loop being a full `lfs_fs_deorphan` pass.
   
2. lfs_file_read/write/flush - fedf646c79c44061ca62f5ae3019fe1198ec07d2

   A bit easier to solve, `lfs_file_read` and `lfs_file_write` sometimes need to flush state from other `lfs_file_read`/`lfs_file_write` calls. The only problem being that `lfs_file_flush` calls `lfs_file_read`/`lfs_file_write` to do the flush!
   
   This was easy to solve with an extra set of `lfs_file_flushedread`/`lfs_file_flushedwrite` entry points that avoid this. Note that
   this code path couldn't actually occur normally outside of filesystem bugs.

3. lfs_dir_traverse - 8109f282668d0396c1a8df05d38ed215b0b8feac

   `lfs_dir_traverse` was also a bit tricky. It's inherently recursive, but has a strict upper bound of 4 calls (a traverse with a filter with a move with a filter).
   
   I tried a few different things, but it was difficult to unroll things without an extra ~1KiB of code cost. The solution here
   is to add an explicit, bounded stack (just an array of state), which turned out to not increase the code cost too badly. 
   
   I suspect this hints that modeling traversals with callbacks is a poor choice for a non-recursive code, so this may be worth implementing differently at some point in the future.

---

This PR also adds a few more scripts and utilities to help analyze builds and compare different implementations:
- `make code` or `./scripts/code.py` - Reports the code size of functions
- `make data` or `./scripts/data.py` - Reports any data sections, this should always be zero for littlefs
[...]

PR #1137, opened 2025-09-03
fix: compact when dir count hits 0x3ff

dschendt:
When dealing with large block sizes (128KiB, for example) and lots of small files (64 bytes and smaller, in the case of 128KiB block sizes), it's possible to run into the constraint where it seems dir->count can't exceed 1023. For example, here https://github.com/littlefs-project/littlefs/blob/master/lfs.c#L1372. This can result in cases where you can fail to create new files even though you have plenty of disk space left.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -2264,7 +2264,7 @@ static int lfs_dir_relocatingcommit(lfs_t *lfs, lfs_mdir_t *dir,
         }
     }
 
-    if (dir->erased || dir->count >= 0xff) {
+    if (dir->erased && dir->count < 0xff) {
         // try to commit
         struct lfs_commit commit = {
             .block = dir->pair[0],
--- a/lfs.c
+++ b/lfs.c
@@ -2333,6 +2333,10 @@ static int lfs_dir_relocatingcommit(lfs_t *lfs, lfs_mdir_t *dir,
         lfs->gdisk = lfs->gstate;
         lfs->gdelta = (lfs_gstate_t){0};
 
+        if(dir->count == 0x3ff)
+        {
+            goto compact;
+        }
         goto fixmlist;
     }
 
--- a/lfs.c
+++ b/lfs.c
@@ -2264,7 +2264,7 @@ static int lfs_dir_relocatingcommit(lfs_t *lfs, lfs_mdir_t *dir,
         }
     }
 
-    if (dir->erased) {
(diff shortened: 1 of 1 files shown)


geky:
Hi @dschendt, thanks for creating a PR, sorry about the late response.

Digging through the history, it looks like this logic had a similar check at one point (https://github.com/littlefs-project/littlefs/commit/48bd2bff821ae70e85c82d7f6c6c6bcf2ee23dae), but was dropped during the big no-recursion refactor (https://github.com/littlefs-project/littlefs/pull/658).

Thoughts on moving this check into the `if (dir->erased) {` statement ([L2267](https://github.com/littlefs-project/littlefs/blob/8e251dd675da00342d45dac78b6f627f119aed03/lfs.c#L2267)) so it's clear these two conditions have a common code path?

Some other notes:
- The threshold we use in other parts of the codebase ([L2158](https://github.com/littlefs-project/littlefs/blob/8e251dd675da00342d45dac78b6f627f119aed03/lfs.c#L2158)) is `< 0xff`. We should probably do the same here.

  Maybe the threshold for splitting should be opened back up to all 10 bits, but I think that is out-of-scope for a patch release. At the very least we need to watch out for the special `id=0x3ff` value that is used to indicate deleted tags ([SPEC.md](https://github.com/littlefs-project/littlefs/blob/master/SPEC.md#metadata-tags)).
  
- This should probably check `>=` (or `<` in the if statement) instead of relying on `==` here. I don't think it's possible to increase id by more than 1 in the current implementation, but a comparison is safer.

dschendt:
@geky Sorry for the late reply, and thank you for taking the time to look at this and give your feedback. I haven't had much time to dig into your suggested changes but I'm not seeing any improvements on disk utilization with them (both before and after I get an error after less than 2,000 2-byte files). Even taking my original changes and changing the comparison to `>= 0xff` I'm able to see a huge improvement (no errors after 20,000+ 2-byte files). I think we should hold off on merging this and I'd like to take a little more time to dig into this some more.

geky:
@dschendt, oh thank you for double checking. That's probably because the logic I pushed up was wrong. Curiously, it looks like the logic was backwards in 48bd2bf. Progs to unerased flash tend to get picked up as corrupt during littlefs's prog checks, which may explain why this went unnoticed.

I pushed up what should be the correct logic, but can wait until the next patch release to bring this in.

PR #1140, opened 2025-09-16
Add littlefs-toy to the related projects section.

tjko:
For your consideration to possibly add in the README.

littlefs-toy is a small command-line tools with _tar_ like syntax for creating and modifying littlefs images. It supports working on littlefs filesystems embedded inside another file (firmware image, etc.) and can auto-detect blocksize when working on existing littlefs images.

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,7 +267,11 @@ License Identifiers that are here available: http://spdx.org/licenses/
   to create images of the filesystem on your PC. Check if littlefs will fit
   your needs, create images for a later download to the target memory or
   inspect the content of a binary image of the target memory.
-  
+
+- [littlefs-toy] - A command-line tool for creating and working with littlefs
+  images. Uses syntax similar to tar command for ease of use. Supports working
+  on littlefs images embedded inside another file (firmware image, etc).
+
 - [littlefs2-rust] - A Rust wrapper for littlefs. This project allows you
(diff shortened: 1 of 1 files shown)


tjko:
@geky, you're right current binary name (```lfs```) could be confusing. It is now renamed it to ```lfst```, so it doesn't conflict with _littlefs-fuse_ anymore.

Yes, I am capturing error/warning generated by _lfs_mount()_ and if it reports "Invalid block size", I'll try to parse the actual block size from the error message. 

Thanks, for pointing out the possible large block sizes (somehow, I had incorrectly assumed 64KiB was the maximum).
I removed the upper bound check on block size ([L591](https://github.com/tjko/littlefs-toy/blob/bc90483141bbea6415c86dfbdb368a35449db55c/src/lfst.c#L591)), since doesn't seem it was really necessary.

geky:
Just to be clear, the binary rename is _not_ a condition for merging. Feel free to keep the `lfs` name, though I do think it might be confusing for users with multiple tools installed. In the future I think littlefs-fuse's binary should probably be renamed to `lfs2-fuse` (or `lfs2f`?) to distinguish it from more general-purpose tools, but littlefs-fuse needs a revisit so it's been low-priority.

Will be bringing this in on the next patch release.

PR #1141, opened 2025-09-18
Change include style from "" to <>.

denravonska:
This allows using libraries to override headers in order to inject things like custom logging function definitions.
Without this the compiler is always going to pick the headers that are next to the sources.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -5,8 +5,8 @@
  * Copyright (c) 2017, Arm Limited. All rights reserved.
  * SPDX-License-Identifier: BSD-3-Clause
  */
-#include "lfs.h"
-#include "lfs_util.h"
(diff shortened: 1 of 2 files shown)


denravonska:
> IMO, this is hacky, and contrary to convention (`<>` for searching standard system dirs).
> 
> Maybe you can do what you want with e.g. GCC's `-iquote` option? https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html
> 
> Alternatively, maybe add some explicit override option.

MbedTLS uses `MBEDTLS_CONFIG_FILE` for a conditional include. Maybe that could be an option?

denravonska:
Wait, I'm a dofus. There's already support for `LFS_CONFIG`.

Issue #1145, opened 2025-09-28
inline files are 1/4 block or 1/8 block?


amgross:
In the code it seems like files stop being inline when they are more than 1/8 of block:
https://github.com/littlefs-project/littlefs/blob/8e251dd675da00342d45dac78b6f627f119aed03/lfs.c#L4331-L4345
But in the readme it written that the threshold is 14/ block:
https://github.com/littlefs-project/littlefs/blob/8e251dd675da00342d45dac78b6f627f119aed03/DESIGN.md?plain=1#L1470-L1472

geky:
Hi @amgross, good catch.

Looks like the 1/8 check was added in 916b30855878edb7d5af77d9cd2b661a3cd4f733, before DESIGN.md was rewritten. Clearly whoever wrote DESIGN.md (me) didn't double check the code.

1/4 is theoretically the optimal cutoff, but 1/8 was chosen as a safer cutoff. The concern is that the inline cutoff doesn't include file metadata (file names, types, revision counts, etc), and with the mdirs already splitting at 1/2 the block size, cutoff=1/4 risks only one inline file fitting in each mdir.

Note this was before the user-configurable `inline_max` was introduced.

Since then, two things:

1. Experiments have shown "wasted" mdir space is not really that big of a concern, since it contributes to logging and defers mdir compaction. (`split_thresh` should probably be use configurable in the future, but that is a high risk change.)

2. `inline_size` was added as a configuration option, so if 1/4 is problematic, users can override it now.

---

So I think it's safe to bump this back up to 1/4, but we should probably only make such a change on a minor release. littlefs2 is under a feature-freeze at the moment, so this may take some time.

amgross:
Thank you very much, I understand littlefs3 that already contains it is not stable enough for in-field product?

geky:
> Thank you very much, I understand littlefs3 that already contains it is not stable enough for in-field product?

Correct. The key issue is that littlefs3's disk format may change without notice, so if you deploy littlefs3, there may be no upgrade path in the future.

The irony is littlefs3 has significantly more testing than littlefs2 ever did. I think it's production ready from that perspective, but the above issue will make it difficult to adopt bug fixes, etc.

[...]

PR #1146, opened 2025-09-29
Bump default inline_max up to 1/4 block size

geky:
As noted by @amgross, the current `inline_max` default (when littlefs switches from inline files to CTZ skip-lists) does not match the theoretical value in DESIGN.md.

The reason for this is 1/8 was chosen as a safer default during development, due to concerns that 1/4 + mdirs splitting at 1/2 would lead to poor distribution of large inline files in mdirs.

However, two things have happened since then:

1. Experiments have show "wasted" mdir space is less of a concern than initially thought. Extra mdir space contributes to logging, delays mdir compaction, and can overall lead to better performance.

2. `inline_size` was added as a configuration option ([#914](https://github.com/littlefs-project/littlefs/pull/914)), so if 1/4 is problematic users can always override it.

So bumping this back up to 1/4 may make sense. Though open to feedback if this is worthwhile in the shadow of littlefs3.

Related: https://github.com/littlefs-project/littlefs/issues/1145

Note this is actually already bumped to 1/4 in littlefs3 (https://github.com/littlefs-project/littlefs/pull/1111, unstable).


amgross:
I think need also to change 
https://github.com/littlefs-project/littlefs/blob/adad0fbbcf5382c20978d07f94f9c13be9041c1b/lfs.h#L280

geky:
@amgross Good eye! Should be fixed now

I grepped lfs.c, but lfs.h is notoriously not in lfs.c.

PR #1161, opened 2025-10-27
Fixes for 'Implicit conversion loses integer precision' warnings.

amubiera:
Since `lfs_size_t` is defined as `uint32_t` and not `size_t`, there are a few warnings for conversion from `unsigned long` (64-bit) to `unsigned int` (32-bit) on 64-bit platforms.

```
lfs.c:288:12 Implicit conversion loses integer precision: 'unsigned long' to 'lfs_size_t' (aka 'unsigned int')
lfs.c:1760:29 Implicit conversion loses integer precision: 'unsigned long' to 'lfs_tag_t' (aka 'unsigned int')
```

Perhaps `lfs_size_t` should be defined as `size_t` instead?

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -286,7 +286,7 @@ static int lfs_bd_erase(lfs_t *lfs, lfs_block_t block) {
 
 // some operations on paths
 static inline lfs_size_t lfs_path_namelen(const char *path) {
-    return strcspn(path, "/");
+    return (lfs_size_t)strcspn(path, "/");
 }
 
 static inline bool lfs_path_islast(const char *path) {
@@ -1501,7 +1501,7 @@ nextname:
         if (lfs_tag_type3(tag) == LFS_TYPE_DIR) {
             name += strspn(name, "/");
         }
-        lfs_size_t namelen = strcspn(name, "/");
+        lfs_size_t namelen = (lfs_size_t)strcspn(name, "/");
 
         // skip '.'
         if (namelen == 1 && memcmp(name, ".", 1) == 0) {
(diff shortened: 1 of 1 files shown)


lfs.c
```diff
@@ -1761,7 +1761,7 @@ static int lfs_dir_commitcrc(lfs_t *lfs, struct lfs_commit *commit) {
 
         commit->off = noff;
         // perturb valid bit?
-        commit->ptag = ntag ^ ((0x80UL & ~eperturb) << 24);
+        commit->ptag = (lfs_tag_t)(ntag ^ ((0x80UL & ~eperturb) << 24));
```
geky: For this one, it looks like the issue is the `0x80UL`. Would it be better to do something like:

``` c
commit->ptag = ntag ^ ((lfs_tag_t)(0x80 & ~eperturb) << 24));
```

And try to minimize hiding other issues with the cast?

geky:
Hi @amubiera, thanks for creating a PR, and sorry for the late response.

These casts look reasonable to me, just one nit on the tag cast.

geky:
> Perhaps `lfs_size_t` should be defined as `size_t` instead?

That's an interesting question. Not without significant changes as least. We write `lfs_size_t`s to disk and that would break the on-disk format on 64-bit machines. This is probably fixable (I think it's fixed in littlefs3 (https://github.com/littlefs-project/littlefs/pull/1111, unstable)), but there's still an assumption `sizeof(lfs_size_t) < sizeof(lfs_off_t)` everywhere.

Even on 64-bit machines, keeping `lfs_size_t` uint32_t would be ideal to minimize the RAM consumption, but maybe we leave that up to users to decide.

Unfortunately changing the codebase to support 64-bit lfs_size_t, even just to error on >2GiB inputs, would require a significant amount of work.

geky (approved):
Looks good here, thanks for this. Will bring this in on the next patch release.

geky:
Sorry for the delay, this is finally coming in. Thanks again for the pr!

Issue #1076, opened 2025-02-24
lfs_setattr on open file


bailyzintl:
Are updated custom attributes written to disk immediately even on an open file? If I want to update a files contents and custom attribute atomically is the only way the tmp file and rename route?

geky:
Ah! So this is what [`lfs_file_config.attrs`](https://github.com/littlefs-project/littlefs/blob/0494ce7169f06a734a7bd7585f49a9fa91fa7318/lfs.h#L351-L362) is for.

The tricky part is that file-attached attrs need to be back by RAM, so you need to call `lfs_file_opencfg` to provide per-attribute buffers.

But any attributes in `lfs_file_config.attrs` are atomically written to the file's metadata when you call `lfs_file_sync` or `lfs_file_close`.

The RAM requirement is the reason there is no `lfs_file_setattr` or `lfs_file_getattr`. And, as it sounds like you have noticed, `lfs_setattr` writes to the disk immediately and may not be powerloss safe.

rcolombanWis:
The issue I found with `lfs_file_config.attrs` is that the file metadata is always written on sync/close, even if nothing has changed.

It is because when the file is opened, the `LFS_F_DIRTY` flag is added when reading the attributes.

rcolombanWis:
It's not super clean but I ended up doing like your second option by clearing the LFS_F_DIRTY flag on open and setting it when changing the attribute.

I find the `lfs_attr` struct more convenient than a dedicated API, having a specific flag set by a macro should be fine.

But other users might prefer an API, so having both and configurable is a good idea.

PR #1173, opened 2025-12-01
Make attribute writing be dependent on attr_count upon close/sync and not upon open

amgross:
Current user attributes behavior given upon open to write is bit complicated:
**If attribute count is not 0 upon open:**
At the first sync (or close) after the open things will be written to the disk (in cost of wear etc.) even if user decided not to write attributes and changed attribute count to 0 (maybe wanted to read attributes upon open and according that decided not to update them?).
at next syncs (or close) will write user attributes will be re-written (and changed if user changed them in the buffer) just if the file changed till then.
**If attribute count is 0 upon open:**
syncs (or close) user attributes will be re-written (and changed if user changed them in the buffer) just if the file changed till then.

This is complex behave, and in cases for example that user opened file with READ-WRITE and wanted to read attributes it will cause write to flash even if he decided not to change anything.

**Solution in this fix**
The write of the user attributes will be done if and just if file was opened to write and attribute count is greater than 0 upon the close/sync.

**tests that added (and will fail without this PR):**
1. if upon open attribute count was 1 and upon sync it was 0 and no change done to the file - there was no write to the file metadata (file position stay same)
2. changing attribute count to 1 before sync cause write of the user attributes to the file, even if the file wasn't change (before the fix it happened just in first sync after opening if the attribute count was bigger than 0 upon opening).
3. if multiple syncs happen when attribute count is bigger than 1, each time attributes will be written. (this is reasonable behavior, but user need to be aware of it if fore some reason it calls lots of syncs without any change and the attribute count is bigger than 0 each sync will cause write to the disk)

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -3180,8 +3180,6 @@ static int lfs_file_opencfg_(lfs_t *lfs, lfs_file_t *file,
                 err = LFS_ERR_NOSPC;
                 goto cleanup;
             }
-
-            file->flags |= LFS_F_DIRTY;
         }
 #endif
     }
@@ -3441,7 +3439,9 @@ static int lfs_file_sync_(lfs_t *lfs, lfs_file_t *file) {
     }
 
 
-    if ((file->flags & LFS_F_DIRTY) &&
+    if (((file->flags & LFS_F_DIRTY) ||
+            (/* User request attributes write */(file->cfg->attr_count > 0) && (file->flags & LFS_O_WRONLY)))
+            &&
             !lfs_pair_isnull(file->m.pair)) {
         // before we commit metadata, we need sync the disk to make sure
         // data writes don't complete after metadata writes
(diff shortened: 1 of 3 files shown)


geky:
Hi @amgross, thanks for creating a PR for this (and for writing tests!)

Sorry about the very late response. littlefs2 is currently under a "feature-freeze" to help focus on completing littlefs3 (https://github.com/littlefs-project/littlefs/pull/1111, unstable), so I appreciate this PR but unfortunately will not be able to review/merge it anytime soon.

The attr API is also being reworked in littlefs3, with some plans to adopt `lfs3_file_setattr/getattr` to simplify file-attached attrs (https://github.com/littlefs-project/littlefs/issues/1076#issuecomment-3325179260). So the attr API may look very different in the future.

I'll try to revisit this when actionable, but in case a forget, feel free to reopen or ping this PR after littlefs3 work lands.

PR #1180, opened 2026-01-22
 lfs: guard null callbacks in lfs_dir_fetchmatch

ChaserTlss:
lfs_dir_fetch relies on an impossible tag match to avoid calling a
  NULL callback. Add an explicit cb != NULL check in the match path
  so future refactors don’t risk a NULL function-pointer call.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -1290,7 +1290,7 @@ static lfs_stag_t lfs_dir_fetchmatch(lfs_t *lfs,
             }
 
             // found a match for our fetcher?
-            if ((fmask & tag) == (fmask & ftag)) {
+            if ((fmask & tag) == (fmask & ftag) && (cb != NULL)) {
                 int res = cb(data, tag, &(struct lfs_diskoff){
                         dir->pair[0], off+sizeof(tag)});
                 if (res < 0) {
--- a/lfs.c
+++ b/lfs.c
@@ -1290,7 +1290,8 @@ static lfs_stag_t lfs_dir_fetchmatch(lfs_t *lfs,
             }
 
             // found a match for our fetcher?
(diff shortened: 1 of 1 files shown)


geky:
Hi @ChaserTlss, thanks for creating a PR, sorry for the late response.

This was discussed in https://github.com/littlefs-project/littlefs/pull/1159. My opinion is a NULL cb should not be possible here, so an assert would be more appropriate than a runtime check (https://github.com/littlefs-project/littlefs/pull/1159#issuecomment-3513514454).

Could this be changed to an `LFS_ASSERT(cb != NULL)` after the if statement?

ChaserTlss:
Of course, I understand that the original logic should not produce a null pointer. The null check is only added as a safeguard for future changes and to prevent unpredictable behavior.

geky:
Thanks! Will bring this into the next patch release

---

If you want stronger guarantees than the assert, you can always define LFS_ASSERT as a relatively cheap `if (!(cond)) { while (1) {}}`, or `if (!(cond)) { return LFS_ERR_INVAL; }` (see https://github.com/littlefs-project/littlefs/pull/514 for caveats).

It loses some granularity, but also gives the user more options.

geky (approved):

lfs.c
```diff
@@ -1291,7 +1291,7 @@ static lfs_stag_t lfs_dir_fetchmatch(lfs_t *lfs,
 
             // found a match for our fetcher?
             if ((fmask & tag) == (fmask & ftag)) {
-                LFS3_ASSERT(cb != NULL);
```
geky: Hehe, a Freudian slip on my part.

Just a heads up, will squash this commit during merge.

PR #1181, opened 2026-01-23
increase hint size in lfs_dir_traverse

zain-noman:
Hi, I was noticing an issue where the number of read calls was too high. One of the functions that was causing this was lfs_dir_traverse. It was issuing read calls for each tag it read. From the comments in [pull request 621](https://github.com/littlefs-project/littlefs/pull/621), i concluded that the sequential reads could be reduced by increasing the hint size.

I compared the number of reads before and after changing the hint size and noticed a significant reduction (from 379K to 79K). The results are compiled in the attached excel file. These results were collected using the following script

```C
#include "lfs.h"
#include "lfs_rambd.h"
#include <stdio.h>

int readCount = 0;
int bytesRead = 0;

int read_with_count(const struct lfs_config *c, lfs_block_t block,
    lfs_off_t off, void *buffer, lfs_size_t size)
{
    readCount++;
    bytesRead += size;
    return lfs_rambd_read(c,block,off,buffer,size);
}

int main()
{
    struct lfs_rambd_config bdCfg = {
        .read_size = 1,
        .prog_size = 1,
        .erase_size = 256,
        .erase_count = 32768
    };
    lfs_rambd_t bd;
    
    lfs_t lfs;
    struct lfs_config lfsCfg = {
        .context =  &bd,
        .read = read_with_count,
        .prog = lfs_rambd_prog,
        .erase = lfs_rambd_erase,
        .sync = lfs_rambd_sync,
        .read_size = 1,
        .prog_size = 1,
        .block_size = 256,
        .block_count = 32768,
        .block_cycles = -1,
        .cache_size = 256,
        .lookahead_size = 4096
    };

    lfs_rambd_create(&lfsCfg,&bdCfg);
    lfs_format(&lfs, &lfsCfg);
    lfs_mount(&lfs, &lfsCfg);

    char fileData1[72];
    for (int i = 0; i < sizeof(fileData1); i++)
    {
        fileData1[i] = i;
    }

    lfs_mkdir(&lfs,"myDir");

    printf("iteration,bytesRead,readCount\n");
    int maxReadCount = 0;
    for (int i = 0; i < 120; i++)
    {
        lfs_file_t testFile;
        char fileName[100];
        sprintf(fileName,"myDir/test%d",i);
        lfs_file_open(&lfs,&testFile,fileName,LFS_O_RDWR|LFS_O_CREAT);
           
        for (int i = 0; i < 900; i++)
        {
            readCount = 0;
            bytesRead = 0;
            lfs_file_write(&lfs,&testFile,fileData1,sizeof(fileData1));
            
            if ((i%30)==0)
                lfs_file_sync(&lfs,&testFile);
            
            // if (readCount > maxReadCount){
            //     maxReadCount = readCount;
                printf("%d,%d,%d\n", i*900 + i, bytesRead, readCount);
            // }
        }
        
        lfs_file_close(&lfs,&testFile);
    }
    lfs_unmount(&lfs);    
}
```
[cache_dir_traverse.xlsx](https://github.com/user-attachments/files/24821767/cache_dir_traverse.xlsx)

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -930,7 +930,7 @@ static int lfs_dir_traverse(lfs_t *lfs,
             if (off+lfs_tag_dsize(ptag) < dir->off) {
                 off += lfs_tag_dsize(ptag);
                 int err = lfs_bd_read(lfs,
-                        NULL, &lfs->rcache, sizeof(tag),
+                        NULL, &lfs->rcache, lfs->cfg->block_size,
                         dir->pair[0], off, &tag, sizeof(tag));
                 if (err) {
                     return err;


geky:
Hi @zain-noman, thanks for creating a PR, sorry about the late response.

On one hand, I'm unlikely to merge this sort of change right now due to the current feature-freeze that allows focusing on littlefs3 (#1111, unstable). On the other hand, the main motivation for littlefs3 _is_ performance, so this would be a good thing to consider.

Let me see if I can run it in our current benchmarks (may take a while).

---

Copying the raw numbers here (not a big issue, but .csvs are a bit easier when you don't have excel):

```
              before   after
read calls:   379139   79918 (-78.9%)
read bytes: 10749289 9723289 (-9.5%)
```

Some aimless thoughts:

- The main concern is if this forces lfs_dir_traverse to read the full block when it does not need to. Mdir compaction is already a significant bottleneck, and involves nested lfs_dir_traverse "calls" (the emulated stack in this function). If `cache_size`/`block_size` is significantly higher (4KiB? 128KiB?), this risks worse performance.

- I wonder if the same result can be emulated by setting `read_size` to a larger value. This would force a larger byte/call ratio while avoiding the full block reads. Though it may be a cludge.

- NOR flash is surprisingly cheap for many small reads, there's really just the overhead of sending the address. But you're also seeing savings in total bytes read, which is interesting.

ashh87:
I've come up with the same solution to slow file renaming and was going to open a PR, but I'll just add to the comments here. Occasionally a file renaming operation will take a lot longer than usual, in the 10s of seconds, and increasing the hint in lfs_dir_traverse stops it from happening.

My setup:
read_size = 16,
cache_size = 256,
block_size = 4096

@geky I do get the same result by increasing read_size to 256 but also feel this isn't quite what I want, as I thought read_size should be as small as possible for performance(?) as well as reflecting the true minimum of what the flash chip supports. The chip in use also has a larger page size, so does support efficient reads of larger amounts of data.

I also set hint to `lfs->cfg->block_size` but I guess that gets clipped to the cache size, which I've already kept relatively small.

This PR seems like a good performance improvement. If the main concern is around mdir compaction, would it be worth adding a parameter for this hint size? Or perhaps calculating a hint which is somewhere between read_size and block_size? From what I've seen, setting the cache size higher and reading the full block isn't necessary to avoid the issue with rename times.

Issue #366, opened 2020-01-14
Infinite loop in lfs_fs_traverse after lfs_dir_read of sub directory


ArkEngr:
I created a user interface that let me make and read directories and make and read files on a PC.  However when reading a sub directory with one file, I got in an infinite loop in the lfs_fs_traverse function when it was called from lfs_fs_size after the lfs_dir_read of the sub directory path.  This was found to be caused by  lfs_t structure element f->next equal to the current f  which caused the for loop in lfs_fs_traverse:
    // iterate over any open files
    for (lfs_file_t *f = (lfs_file_t*)lfs->mlist; f; f = f->next) {
to never exit.
there seems to be a couple of f = f->next  in for loops that may have the same problem.

My source code:
              case 'l' :  /* fl [<path>] - Directory listing */
                    while(*ptr == ' ')
                    {
                      ptr++;
                    }

                    printf("ptr size %d\n", strlen(ptr));

                    res = lfs_dir_open(&LFSfs, &Dir, ptr);

                    if(res)
                    {
                      put_rc(res);
                      break;
                    }

                    p11 = s1 = s2 = 0;

                    for(;;)
                    {
                      res = lfs_dir_read(&LFSfs, &Dir, &Finfo);

                      if((res < LFS_ERR_OK) || !Finfo.name[0])
                      {
                        break;
                      }

                      if(Finfo.type == LFS_TYPE_DIR)
                      {
                        s2++;
                      }
                      else
                      {
                        s1++;
                        p11 += Finfo.size;
                      }

                      xprintf("%c %15,llu  %-12s  %s\n",
                                (Finfo.type & LFS_TYPE_DIR) ? 'D' : '-',
                                Finfo.size, Finfo.name,
                                "");
                    }

                    printf("%4u File(s),%15llu bytes total\n%4u Dir(s)", s1, p11, s2);

                    ss1 = lfs_fs_size(&LFSfs);  // infinite loop when path is subdirectory
                    if(ss1 >= 0)
                    {
                      printf(", %15d blocks used\n", ss1);
                    }
                    else
                    {
                      put_rc(ss1);
                    }
                    break;

geky:
Interesting, thanks for reporting this.

It shouldn't happen but may be a bug relating to closing files/dirs? Do you call lfs_file/dir_close for every  open call?

ArkEngr:
[...]

PR #1184, opened 2026-01-29
Separate LFS_ASSERT to LFS_USER_ASSERT and LFS_DISK_ASSERT

amgross:
I think it is very important to give user option to control differently user and disk asserts, especially now when can return error from asserts.

This already raised up by @geky in https://github.com/littlefs-project/littlefs/issues/366#issuecomment-2875052032 but claimed to be breaking change and low priority.

I think it is important, and my implementation is without breaking changes because if user does not define them they are changing to LFS_ASSERT.

### DIFF
--- a/bd/lfs_emubd.c
+++ b/bd/lfs_emubd.c
@@ -292,10 +292,10 @@ int lfs_emubd_read(const struct lfs_config *cfg, lfs_block_t block,
     lfs_emubd_t *bd = cfg->context;
 
     // check if read is valid
-    LFS_ASSERT(block < bd->cfg->erase_count);
-    LFS_ASSERT(off  % bd->cfg->read_size == 0);
-    LFS_ASSERT(size % bd->cfg->read_size == 0);
-    LFS_ASSERT(off+size <= bd->cfg->erase_size);
+    LFS_DISK_ASSERT(block < bd->cfg->erase_count);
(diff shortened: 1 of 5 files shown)


amgross:
This was mentioned also at https://github.com/littlefs-project/littlefs/pull/921#issuecomment-1896894113

geky:
Hi @amgross, thanks for creating a PR. Sorry if due to the ongoing feature-freeze I don't give the attention it deserves (for #1111, unstable).

Thanks for going through and figuring out which asserts are disk vs user! I will revisit this, it should be very useful in the future.

---

Reading through, I think this could be merged as-is if it wasn't for the feature-freeze.

I was also considering including an optional errno argument in these asserts to allow different error-codes per-assert, but maybe that's overengineering. I guess the question is can all LFS_USER_ASSERTs be mapped to the same error-code? (LFS_ERR_INVAL?)

One minor issue is prettyasserts.py will need to be updated with the new assert names. This may just involve adding `-p LFS_USER_ASSERT` and `-p LFS_DISK_ASSERT` to the Makefile ([here](https://github.com/littlefs-project/littlefs/blob/adad0fbbcf5382c20978d07f94f9c13be9041c1b/Makefile#L538-L542)), but it may be more involved if we need per-assert error-codes. prettyasserts.py operates at the syntax-tree, so unfortunately the fallback to LFS_ASSERT doesn't work here.

PR #1186, opened 2026-02-17
Render write buffers non mandatory for LFS_READONLY configuration.

Jerome-PS:
Render write buffers non mandatory for LFS_READONLY configuration.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -4278,7 +4278,7 @@ static int lfs_init(lfs_t *lfs, const struct lfs_config *cfg) {
             goto cleanup;
         }
     }
-
+#ifndef LFS_READONLY
     // setup program cache
     if (lfs->cfg->prog_buffer) {
         lfs->pcache.buffer = lfs->cfg->prog_buffer;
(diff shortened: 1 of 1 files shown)


amgross:
Nice, maybe also remove the free of `pcache.buffer` in `lfs_deinit`?

Jerome-PS:
> Nice, maybe also remove the free of `pcache.buffer` in `lfs_deinit`?

Well spotted!
MCUs always care about initialization, but not always about clean-up.

amgross (approved):

geky:
Hi @Jerome-PS, thanks for creating a PR, this would be a good change to have.

Unfortunately I think it tiptoes into new feature territory. Normally this would be added to the next minor release, but with the feature-freeze for littlefs3 work (#1111, unstable), there's no plan for a minor release any time soon. Sorry if this stays open for a while.

lfs.c
```diff
@@ -4289,11 +4289,12 @@ static int lfs_init(lfs_t *lfs, const struct lfs_config *cfg) {
             goto cleanup;
         }
     }
-
+#endif	//ndef LFS_READONLY
```
geky, line 4292: nit: These ndef comments are inconsistent with the rest of the codebase. Not a big deal, but the inconsistency hurts readability.

If we want ndef comments that should probably be a separate PR, but would likely be low-priority.

lfs.c
```diff
@@ -4278,7 +4278,7 @@ static int lfs_init(lfs_t *lfs, const struct lfs_config *cfg) {
             goto cleanup;
         }
     }
-
+#ifndef LFS_READONLY
     // setup program cache
     if (lfs->cfg->prog_buffer) {
         lfs->pcache.buffer = lfs->cfg->prog_buffer;
```
geky, line 4284: Also, feel free to not put more time into this considering the feature-freeze, but thoughts on ifdefing out the entire `pcache` fields in `lfs_t`? That would save another 16 bytes of RAM.

PR #1191, opened 2026-02-28
Add copyright and license info macros.

tjko:
Often libraries expose copyright and/or license information for programs that use the library.

This change adds ```LFS_COPYRIGHT``` and ```LFS_LICENSE_IDENTIFIER``` macros in _lfs.h_, to allow programs (that use littlefs) easily display littlefs copyright message.

### DIFF
--- a/lfs.h
+++ b/lfs.h
@@ -32,6 +32,14 @@ extern "C"
 #define LFS_DISK_VERSION_MAJOR (0xffff & (LFS_DISK_VERSION >> 16))
 #define LFS_DISK_VERSION_MINOR (0xffff & (LFS_DISK_VERSION >>  0))
 
+// Copyright info
+#define LFS_COPYRIGHT \
+	"Copyright (c) 2022, The littlefs authors.\n" \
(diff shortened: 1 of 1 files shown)


geky:
Hmmm, does any other library do this?

I appreciate the PR and interest in improving littlefs, but this seems atypical. There's many things we could define, but the more defines we have the more maintenance overhead there is.

tjko:
Some examples that come to mind:
- libjpeg (and variants) define [JCOPYRIGHT](https://github.com/libjpeg-turbo/libjpeg-turbo/blob/main/src/jversion.h.in) macro.
- libpng provides a function to retrieve the info:
  ```
  /*
   * A "png_get_copyright" function is available, for convenient use in "about"  
   * boxes and the like:
   *
   *    printf("%s", png_get_copyright(NULL)); 
   *
   * Also, the PNG logo (in PNG format, of course) is supplied in the  
   * files "pngbar.png" and "pngbar.jpg (88x31) and "pngnow.png" (98x31).
   */
  ```

PR #1192, opened 2026-02-28
Add basic CMake support.

tjko:
Make it easy to add littlefs in CMake projects using _add_subdirectory()_ command.

To add littlefs in an existing CMake project, user can simply add one line in their CMakeLists.txt:
```
add_subdirectory(somepath/littlefs)
```
(where "somepath/littlefs" is path to the directory where littlefs source is found)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,16 @@
+# littlefs 
+
+add_library(littlefs INTERFACE)
+
+target_include_directories(littlefs INTERFACE
+  ${CMAKE_CURRENT_LIST_DIR}
+)
+
+target_sources(littlefs INTERFACE
+  lfs.c
+  lfs_util.c
+)
+
+if (CMAKE_C_COMPILER_ID MATCHES "GNU|Clang")
+  target_compile_options(littlefs INTERFACE "-Wno-unused-function")
+endif()


tjko:
Just thought this might be useful for others.  As currently including littlefs in CMake based projects is somewhat involved.

Currently one must add something like this in their CMakeLists.txt:
```
add_library(littlefs INTERFACE)
target_include_directories(littlefs INTERFACE
  libs/littlefs/
)
target_sources(littlefs INTERFACE
  libs/littlefs/lfs.c
  libs/littlefs/lfs_util.c
)
```

After this change only one line would be needed in CMakeLits.txt:
```
add_subdirectory(libs/littlefs)
```

geky:
Hi @tjko, thanks for creating a PR.

There's some discussion on this in: https://github.com/littlefs-project/littlefs/pull/857

Given my limited experience with CMake, I don't think there's a single CMakeLists.txt that will make all downstream users happy, and a build system that requires changes to the target libraries seems like a broken build system.

I appreciate the PR, and feel free to maintain a littlefs-cmake wrapper repo, but this is not something worth messing with from my perspective.

tjko:
@geky, thanks I understand.  I'll close the PR.

PR #1195, opened 2026-03-18
refactor: remove possibly unnecessary call to lfs_ctz_fromle32

udovichenkoAlexander:
In `lfs_dir_getinfo` `lfs_ctz_fromle32` results not used if entry is `LFS_TYPE_INLINESTRUCT`.

In most cases, the compiler will optimize this anyway. But in debug builds, it makes sense.

### DIFF
--- a/lfs.c
+++ b/lfs.c
@@ -1433,10 +1433,9 @@ static int lfs_dir_getinfo(lfs_t *lfs, lfs_mdir_t *dir,
     if (tag < 0) {
         return (int)tag;
     }
-    lfs_ctz_fromle32(&ctz);
 
     if (lfs_tag_type3(tag) == LFS_TYPE_CTZSTRUCT) {
-        info->size = ctz.size;
+        info->size = lfs_fromle32(ctz.size);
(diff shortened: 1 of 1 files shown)


geky:
Hi @udovichenkoAlexander, thanks for creating a PR.

I think you're right the full `lfs_ctz_fromle32` is not necessary. Though @geky-bot confirms the compiler is smart enough to omit the unused fromle32.

To be honest, given the state of things with v3 on the horizon (https://github.com/littlefs-project/littlefs/pull/1111, unstable), the current feature freeze, and the compiler handling things, I'm not sure this is worth merging.

Some of the reason for `lfs_ctz_fromle32` immediately after the `lfs_dir_get` is to reduce the likelyhood that any fields are used without fromle32 in a future refactor.

udovichenkoAlexander:
> Some of the reason for lfs_ctz_fromle32 immediately after the lfs_dir_get is to reduce the likelyhood that any fields are used without fromle32 in a future refactor.

Ok. So just close it

